PR #388, opened 2014-12-19
use UNKNOWN as size type for table-like keywords

andlaus:
this allows the parser not to rely on the user-specified table
dimension which is demoted to a semantic check that, produces a
warning in the log if a mistake is detected.

The prime motivation for this is that it is a way to specify tables in decks which do not feature the TABDIMS keyword.

The reason why this is a separate PR is that github prevented Joakim from reopening #346. As an alternative to setting all keywords which determine their size using a different keyword to sizeType UNKNOWN, a new size type for tables could be introduced. since this solution requires more extensive changes to the low level code and probably also some additions to the JSON keyword descriptions of table keywords, it is a thing for after the holidays. (if it is decided to be the way to go.)


andlaus:
I figured that I don't have enough spare cycles to do this. I'll close it for now, but if someone wants to implement it properly, reopen it...

PR #395, opened 2015-01-13
Closed completions

iLoop2:
Added support for completions status on the WELOPEN keyword.


alfbr:
Thanks for the contribution! As a general note you should sign-off the commit (in the commit message preferably, so that it is contained in the git history) with full name and a contact e-mail. Anonymous contributions are problematic from a copyright perspective.

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -388,19 +389,111 @@ namespace Opm {
         }
     }
 
+    Opm::Value<int> Schedule::getValueItem(DeckItemPtr item){
```
joakim-hove: This *highly* specialized function should not be part of `Schedule` class, but rather be a free function with static linkage. The input argument can be of type `DeckIntItemPtr`.

opm/parser/eclipse/EclipseState/Schedule/Well.cpp
```diff
@@ -212,6 +212,10 @@ namespace Opm {
         m_completions->add( time_step , newCompletionSet);
     }
 
+    void Well::addCompletionsSet(size_t time_step, const CompletionSetConstPtr newCompletionSet){
```
joakim-hove: The `addCompletions()` method above should use the `addCompletionsSet()` (I suggest `addCompletionSet()`) method.

opm/parser/eclipse/EclipseState/Schedule/tests/ScheduleTests.cpp
```diff
@@ -270,4 +270,121 @@ BOOST_AUTO_TEST_CASE(ReturnMaxNumCompletionsForWellsInTimestep) {
...
+                    " 'OP_3' SHUT 0 0 0 / \n"
+                    "/\n"
+                    "DATES             -- 4\n"
+                    " 10  JUL 2008 / \n"
+                    "/\n"
+                    "WELOPEN\n"
+                    " 'OP_1' OPEN / \n"
+                    " 'OP_2' OPEN 0 0 0 4 6 / \n "
+                    " 'OP_3' OPEN 0 0 0 / \n"
+                    "/\n";
```
joakim-hove: I want a fifth `DATES` step with a `WELOPEN` line like:

```txt
WELOPEN
   'OP_1'   STOP  / 
/
```

Then the test should be that `well->getStatus() == STOP` and that all completions have status `SHUT` (i.e. `STOP`should be translated to `SHUT` for the completions. 

This should be handled automagically if the `WellCompletion::String2CompletionState()` function is used.

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -388,19 +389,111 @@ namespace Opm {
...
+                    haveCompletionData = true;
+                    break;
                 }
             }
+
+
+            const std::string& wellNamePattern = record->getItem("WELL")->getTrimmedString(0);
+            const std::vector<WellPtr>& wells = getWells(wellNamePattern);
+
             WellCommon::StatusEnum status = WellCommon::StatusFromString( record->getItem("STATUS")->getTrimmedString(0));
```
joakim-hove: This should be moved down to the `if (!haveCompletionData) { ... }` block.

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -388,19 +389,111 @@ namespace Opm {
...
+                            newCompletionSet->add(completion);
+                            continue;
+                        }
+
+                        if ((K.hasValue() && (!(K.getValue() == ck) || K.getValue()!=0))) {
+                            newCompletionSet->add(completion);
+                            continue;
+                        }
+
+                        Value<double> transmissibilityFactor("transmissibilityFactor", completion->getConnectionTransmissibilityFactor());
```
joakim-hove: Instead of the code block on lines 473 - 488 I would suggest adding a copy constructor: 
```C++
Completion::Completion(const Completion& oldCompletion, WellCompletion::StatusEnum newStatus) {
    ...
}
```

Also observe that when creating new shared pointers the preferred way is to use `std::make_shared<T>( ... )`

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -388,19 +389,111 @@ namespace Opm {
...
+                                newCompletionSet->add(completion);
+                                continue;
+                            }
+                        }
+
+                        int ci = completion->getI();
+                        int cj = completion->getJ();
+                        int ck = completion->getK();
+
+                        if ((I.hasValue() && (!(I.getValue() == ci) || I.getValue() !=0))) {
```
joakim-hove: With the `getValue<int>()`method we should never assign a zero to the `Value<int>` objects, so it should be sufficient to do:
```C++
 if ((I.hasValue() && (!(I.getValue() == ci)
    newCompletionsSet->add(completion);
...
```

joakim-hove:
OK - I'm merging this now, with this patch applied the dynamic opening & closing of completions should be supported.

totto82:
Great. I will start testing it asap.

atgeirr:
We will still have to implement proper completion data transfer, but you could test it with 0 rates for completions at this stage.

joakim-hove:
> We will still have to implement proper completion data transfer,

Ehhh?

Observe that completions not withstanding this: https://github.com/OPM/opm-core/pull/721 must be merged before Norne will run.

atgeirr:
I was thinking of transfer of well state from one timestep to the next in the presence of changing completions, sorry for being dense.

totto82:
I have done what you suggested. Testing in progress.

PR #398, opened 2015-01-16
Singleton Logger

joakim-hove:
This PR implements a singleton Logger pattern. It is currently implemented in opm-parser, but such a way that it should be simple to move it elsewhere. 

The system consists of three layers:

1. `OpmLog` is a static class which manages a `Logger` instance.
2. The `Logger`instance is a normal class which holds on to a number of logging backends.
3. All logging backends derive from `LogBackend`.

Using this system will be like this:

1. In the main application `LogBackend` instances are created and added to the logger with `OpmLog::addBackend()`.
2. In the library code requests for logging are issued as: `OpmLog::addMessage( )`

This is far from 100% finished, but I was quite ready to get some comments+++

The necessary updates to opm-core and opm-autodiff are in the works.


joakim-hove:
Merging this now in agreement with @atgeirr. There is certainly more to with the logging framework itself, in particular *verbosity*, but the changes in downstream modules `opm-core` and `opm-autodiff` are very small anyway.

PR #399, opened 2015-01-16
OPM 147: WELOPEN minor changes

chflo:
WELOPEN minor changes:
Shut well if all completions are shut; throw exception if trying to open a well with all completions shut.


totto82:
I think the current behavior if you try to open a well where all completions are shut is to keep it shut with a warning and not to throw. Also it does not handle cases where you try to open a well using WCONINJE/H or WCONPROD/HIST.

chflo:
Thanks for the feedback; we will remove the throw and also add handling of WCONINJE/H and  WCONPROD/HIST.

PR #401, opened 2015-01-22
Open well shut completions

iLoop2:
Change so that the status of a well do not change to open if there is no open completions.

Fredrik Gundersen
fgun@statoil.com


totto82:
Thanks. Tested and it works for Norne. 

I suggest to change the error message as well as it does not only apply to WELOPEN
for instacne. "ERROR when opening well ..."

joakim-hove:
OK good - I will merge now, error message disgregarded - that should hopefully go in a log file in the near future anyway.

PR #402, opened 2015-01-26
Well refdepth

joakim-hove:
This PR contains several small fixes to improve/correct the calculation of well reference depth:

 - The Schedule and Well objects have a grid reference.
 - Well reference depth is calculated as the depth of the first occuring completion.

As a followup to this PR opm-core/wellsmanager will just ask for the reference depth, without checking the default state of the input. Accompanying PR in opm-core is coming.


joakim-hove:
NB - this has been rebased ;-)

opm/parser/eclipse/EclipseState/Schedule/Well.cpp
```diff
@@ -190,14 +169,33 @@ namespace Opm {
...
-    double Well::getRefDepth() const {
-        return m_refDepth;
+
+    void Well::setRefDepthFromCompletions() const {
+        size_t timeStep = m_creationTimeStep;
+        while (true) {
+            auto completions = getCompletions( timeStep );
+            if (completions->size() > 0) {
+                auto firstCompletion = completions->get(0);
+                double depth = m_grid->getCellDepth( firstCompletion->getI() , firstCompletion->getJ() , firstCompletion->getK());
```
atgeirr: Are the (i,j,k) coordinates here 1-based or 0-based?

joakim-hove:
> Are the (i,j,k) coordinates here 1-based or 0-based?

All the ECLIPSE facing code I get my hands on follows this convention:

1. When loading ECLIPSE data immediately do a --; shift.
2. Only zero offset internally.
3. Make a +1 shifted copy if writing out again.

So - according to rule 2) the i,j,k should be zero based.

PR #403, opened 2015-01-27
Ordering for well completions

atgeirr:
With this, well completions will be ordered according to a simple yet useful algorithm:
 1. Start with the completion nearest to the well head in (I, J) coordinates. If not unique, choose the one with smallest depth.
 2. Choose as next completion the one nearest to the current in (I, J). If not unique, choose the one that is closest in depth. Repeat.

This should give something approximating the default COMPORD behaviour of Eclipse (TRACE), the two other possibilities (DEPTH, INPUT) are currently not properly supported:
 - For DEPTH, one should ignore the (I, J) coordinates and only look at depth. The current algorithm will skip to a deeper connection with the same (I, J) coordinates instead of going strictly in depth order.
 - We totally ignore input order, by unconditionally calling the ordering function.

An obvious next step is to properly support COMPORD/DEPTH and COMPORD/INPUT, but since the current approach works on SPE9 at least I consider it reasonable to merge this first.


joakim-hove:
OK - some minor comments added; address those you agree with and I'll merge.

joakim-hove:
Last comment: I think it is perfectly OK not to support the `COMPORD` keyword properly, but we could check it, i.e.:
```C++
if (deck->hasKeyword("COMPORD")) {
    auto compordKeyword = deck->getKeyword("COMPORD");
    for (auto record = compordKeyword->begin(); record != compordKeyword->end(); ++record) {
         auto methodItem = (*record)->getItem(1);
         if (methodItem->getString(0) != "TRACK")
            throw std::invalid_argument("Can only handle TRACK in well order");
    }
}
```

atgeirr:
I think your COMPORD check looks fine, I just would not know where to put it!

atgeirr:
I have now replied to all comments (I think, the review comments were not attached to the diff but to the commits, which makes it a little harder to track them), and find it interesting that we have so many small disagreements! I have given my reasons for preferring my approach, some of these are worth a little bit of discussion I think. Anyway, please indicate which ones you'd need me to change before merging and I will change them.

joakim-hove:
> and find it interesting that we have so many small disagreements!

That is because I have studied physics - never learned to read or write documentation! Seriously I think we have the same goals of readability, but have slightly different opinions of what constitutes readable code. 

> I think your COMPORD check looks fine, I just would not know where to put it!

My suggestion is adding a `Schedule::checkUnhandledKeywords()` method which can be called from the bottom of `Schedule::iterateScheduleSection` - I fear that method will get more checks as time goes by. With that I am happy to merge.

atgeirr:
Implemented checkUnhandledKeywords with your suggested code. That should be all I think.

joakim-hove:
Good - merging.

PR #404, opened 2015-01-29
Temperature dependent viscosity

andlaus:
these changes are required to support temperature dependent viscosity via the "*VISCT" keywords. the additions are pretty minor. (some new keywords plus the corresponding table classes with their glue code.)

PR #405, opened 2015-01-30
Revert "Temperature dependent viscosity"

joakim-hove:
Reverts OPM/opm-parser#404


joakim-hove:
REverting this due to uncertainty how to handle multi column input.

PR #408, opened 2015-01-30
Fixed gcc44 challenged calls to std::to_string()

joakim-hove:


andlaus:
ops. thanks. the joys of old compilers...

PR #409, opened 2015-02-02
make it possible to copy them

andlaus:
during OPM/opm-core#740 @atgeirr and me figured that it is the best solution to copy the tables from opm-parser in order to avoid object lifetime issues. This PR implements just that possibility.

to prevent accidential passing of table objects by value this is done
by makeing all "end user visible" table classes default-constructible
and adding an assignFrom() method...

### DIFF
--- a/opm/parser/eclipse/EclipseState/Tables/EnkrvdTable.hpp
+++ b/opm/parser/eclipse/EclipseState/Tables/EnkrvdTable.hpp
@@ -29,7 +29,6 @@ namespace Opm {
         typedef SingleRecordTable ParentType;
 
         friend class EclipseState;
-        EnkrvdTable() = default;
 
         /*!
          * \brief Read the ENKRVD keyword and provide some convenience
@@ -62,6 +61,8 @@ namespace Opm {
         }
 
     public:
+        EnkrvdTable() = default;
+
         using ParentType::numTables;
         using ParentType::numRows;
         using ParentType::numColumns;
@@ -70,6 +71,10 @@ namespace Opm {
         // using this method is strongly discouraged but the current endpoint scaling
         // code makes it hard to avoid
         using ParentType::getColumn;
+
+        void assignFrom(const EnkrvdTable& other) {
+            ParentType::assignFrom(other);
+        }
 
         /*!
          * \brief The datum depth for the remaining columns
--- a/opm/parser/eclipse/EclipseState/Tables/EnptvdTable.hpp
+++ b/opm/parser/eclipse/EclipseState/Tables/EnptvdTable.hpp
@@ -29,7 +29,6 @@ namespace Opm {
         typedef SingleRecordTable ParentType;
 
         friend class EclipseState;
-        EnptvdTable() = default;
 
         /*!
          * \brief Read the ENPTVD keyword and provide some convenience
@@ -64,6 +63,8 @@ namespace Opm {
         }
 
     public:
+        EnptvdTable() = default;
+
         using ParentType::numTables;
         using ParentType::numRows;
         using ParentType::numColumns;
@@ -72,6 +73,10 @@ namespace Opm {
         // using this method is strongly discouraged but the current endpoint scaling
         // code makes it hard to avoid
         using ParentType::getColumn;
+
+        void assignFrom(const EnptvdTable& other) {
+            ParentType::assignFrom(other);
+        }
 
         const std::vector<double> &getDepthColumn() const
         { return ParentType::getColumn(0); }
--- a/opm/parser/eclipse/EclipseState/Tables/FullTable.hpp
+++ b/opm/parser/eclipse/EclipseState/Tables/FullTable.hpp
@@ -63,7 +63,7 @@ namespace Opm {
                 InnerTable *curRow = new InnerTable;
                 curRow->init(keyword,
                              /*recordIdx=*/m_outerTable->firstRecordIndex() + rowIdx);
-                m_innerTables.push_back(std::shared_ptr<const InnerTable>(curRow));
+                m_innerTables.push_back(std::shared_ptr<InnerTable>(curRow));
             }
         }
 
@@ -85,9 +85,17 @@ namespace Opm {
             return m_innerTables[rowIdx];
         }
 
+        void assignFrom(const FullTable& other) {
+            m_outerTable->assignFrom(*other.m_outerTable);
+            m_innerTables.resize(other.m_innerTables.size());
+
+            for (size_t i = 0; i < other.m_innerTables.size(); ++i)
+                m_innerTables[i]->assignFrom(*other.m_innerTables[i]);
+        }
+
     protected:
-        std::shared_ptr<const OuterTable> m_outerTable;
-        std::vector<std::shared_ptr<const InnerTable> > m_innerTables;
+        std::shared_ptr<OuterTable> m_outerTable;
+        std::vector<std::shared_ptr<InnerTable> > m_innerTables;
 
     };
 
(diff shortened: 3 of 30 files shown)


joakim-hove:
The code here looks innocent enough - although a test has never harmed anyone? Unfortunately I did not follow the OPM/opm-core#740 discussion - but I must say I that I am quite skeptical to the lifetime reasoning behind this PR. As I see it the purpose of this PR is to make possible to refer to and use table object after the `EclipseState/Deck` have gone out of scope. 

Personally I have always assumed that the `EclipseState/Deck` would be available throughout the whole simulation (not the `Parser` mind you - that can obviously die); failing so is a quite significant design change which I would like to discuss on a sligthly more principle level than 'first we take the tables'.

andlaus:
> Unfortunately I did not follow the OPM/opm-core#740 discussion - but I must say I that I am quite skeptical to the lifetime reasoning behind this PR.  As I see it the purpose of this PR is to make possible to refer to and use table object after the EclipseState/Deck have gone out of scope.

Yes, but the logical consequence of not being able to copy tables in the current PVT code is that all data _must_ be copied out of the tables to a second set of objects which has quite a few consequences concerning code duplication, etc.

> Personally I have always assumed that the EclipseState/Deck would be available throughout the whole simulation (not the Parser mind you - that can obviously die); failing so is a quite significant design change which I would like to discuss on a sligthly more principle level than 'first we take the tables'.

I never assumed this (actually, this issue did not make it to the surface of my mind ;)): the design all classes except for the schedule code in opm-core and opm-autodiff follow is to grab all the data which they need from the Deck/EclipseState and to then live happily without it. @atgeirr probably has deeper insight into this, I suppose... (note that when it comes to this, I personally don't mind having the Deck and EclipseState objects around unconditionally, but there are rumors that some simulators within OPM do not even instantiate these.)

also note that, under the assumption that the Deck and EclipseState objects die last, the pervasive usage of `std::shared_ptr` by opm-parser is quite pointless...

joakim-hove:
> [...] but the logical consequence of not being able to copy tables in the current PVT code is that all data must be copied out of the tables ...

or just keep the `EclipseState` alive.

>the design all classes except for the schedule code in opm-core and opm-autodiff follow is to grab all the data which they need from the Deck/EclipseState and to then live happily without it.

Well - at the very least the `EclipseState::Schedule` class needs to be around; trying to copy/internalize that information wholesale into a new object is a freaking insane idea - not that anyone have suggested to do so.

>also note that, under the assumption that the Deck and EclipseState objects die last, the pervasive usage of std::shared_ptr by opm-parser is quite pointless...

Well - I really like `std::shared_ptr<>` - mainly due to the more flexible construction than what (I) can achieve with references; however I am the first to admit that I understood even less of these issues  when the opm-parser was born.

joakim-hove:
> [....] but there are rumors that some simulators within OPM do not even instantiate these.)

Well - they must get their tables from elsewhere then; so this discussion does not apply to those simulators anyway?

andlaus:
> or just keep the EclipseState alive.

that was my original implementation of OPM/opm-core#740...

> trying to copy/internalize that information wholesale into a new object is a freaking insane idea

welcome to the  OPM simulator modules!

> Well - they must get their tables from elsewhere then; so this discussion does not apply to those simulators anyway? 

the main reason I want this is that I don't want to copy-n-paste all the tables code  for temperature dependent viscosity because of maintainablility reasons and because this is a pretty exotic feature. 

In this case, I think it is admissible if a deck must be instantiated, and while @atgeirr wants the livetime of these objects to be independent of the deck, I don't really care so I will implement whatever you two end up with. 

> Well - I really like std::shared_ptr<> - mainly due to the more flexible construction than what (I) can achieve with references; however I am the first to admit that I understood even less of these issues when the opm-parser was born.

yeah, the design of opm-parser certainly does not consider the fact that while _dereferencing_ a `std::shared_ptr` is fast, _constructing_ one isn't because it needs to modify the reference counter if nothing else. That said, opm-parser is not performance critical and -- in my opinion -- fast enough as it is :)

atgeirr:
I guess we did not really have a discussion of expected lifetimes of opm-parser objects before? I agree that the EclipseState can be expected to live throughout the simulation, due to the need to keep asking for the next schedule timestep's state. I don't mind (much) using shared_ptr as a universal solution for object storage and referencing, for code that is not in the inner loop.

The reason I get nervous storing (plain, non-owning) pointers to these tables is precisely that we use shared_ptr everywhere else: the difference (using references in the interface of EclipseState instead of shared_ptr) seems to say that these objects will not live long, but of course that is me over-interpreting this.

Anyway, EclipseState does not document the lifetime of these things, so on principle I should expect the worst... I am not sure what the best solution is, but I think allowing copying is a good one, and one that has a very small cost, and is already implemented here (I have not looked at the code in this PR though, so this is not a review).

joakim-hove:
>I guess we did not really have a discussion of expected lifetimes of opm-parser objects before?

No - that is true; that is probably my fault - I should have raised the discussion earlier.
 
> I agree that the EclipseState can be expected to live throughout the simulation

Good.

>  [...] the difference (using references in the interface of EclipseState instead of shared_ptr) seems to say that these objects will not live long, but of course that is me over-interpreting this.

That is good point; to me the solution is "simple" - the table objects should follow the same pattern as the remaining elements in `EclipseState` - i.e. `std::shared_ptr`. To make deep copies of the table objects as implemented in this PR really breaches the pattern; this way doubt is spread about the lifetime of the remaining elements in `EclipseState` - I do not want that uncertainty to spread.

>Anyway, EclipseState does not document the lifetime of these things, so on principle I should expect the worst...

Well - as I see it the main question is the lifetime of the `EclipseState` itself, and that is managed by the simulator instantiating it.

joakim-hove:
I have thought a bit about this, and I will not merge this PR. The reason I will not merge is that I feel the   introduction of deep copying of tables indirectly spreads FUD about the lifetime of the other elements in the `EclipseState`, without any real benefit as long as the `EclipseState` must stay alive anyway.

As I understand it OPM/opm-core#740 can in principle work with an unmodified version of opm-parser? 
Ideally I would like the tables in `EclipseState` to follow the same `std::shared_ptr` pattern as the rest of the elements in `EclipseState` - but that is not a pressing issue. I would very much like to reorganize  the table code in `EclipseState` slightly[1], that would be a suitable time to change to using the `std::shared_ptr` also for the tables.


[1]: As a bare minimum we should have one or more `Table` classes which again can hold on to the actual table instances, right now there is a too big variation in abstraction level internally in the `EclipseState`.

andlaus:
> That is good point; to me the solution is "simple" - the table objects should follow the same pattern as the remaining elements in EclipseState - i.e. std::shared_ptr. To make deep copies of the table objects as implemented in this PR really breaches the pattern; this way doubt is spread about the lifetime of the remaining elements in EclipseState - I do not want that uncertainty to spread.

well, the way I interpret this statement is that you want the `get*Table()` methods can be changed to return a `std::shared_ptr<std::vector<std::shared::ptr<MyTableClass>>`  instead of a reference to `std::vector<MyTableClass>`. In my opionion this is quite a bit too "pointy", but I'd be willing to implement it. (this will encompass some "cross module PRs"). If you also oppose this way out, @atgeirr must either bite the apple and live with the lifetime issues of OPM/opm-core#740 or  there will be a lot of copy and paste of opm-parser code into opm-core.

andlaus:
> or there will be a lot of copy and paste of opm-parser code into opm-core.

(Note that in this case, there still will be deep copies of these objects, but the classes implementing this will be in opm-core...)

atgeirr:
Let's avoid the massive code copying. A more uniform shared_ptr-based interface would be nice, yes, but I guess I'm willing to accept the current API as good enough (and your original proposal for opm-core, @andlaus).

PR #413, opened 2015-02-06
Add RFT/RFTPLT key word parsing

iLoop2:


opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -587,6 +611,81 @@ namespace Opm {
         m_rootGroupTree->add(currentStep, newTree);
     }
 
+    void Schedule::handleWRFT(DeckKeywordConstPtr keyword, LoggerPtr /*logger*/, size_t currentStep) {
```
joakim-hove: As long as it is unused - just remove the `LoggerPtr` argument.

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -29,6 +29,7 @@
 #include <opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.hpp>
 #include <opm/parser/eclipse/EclipseState/Schedule/WellInjectionProperties.hpp>
 #include <opm/parser/eclipse/EclipseState/Schedule/WellPolymerProperties.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/WellRFTProperties.hpp>
```
joakim-hove: The class `WellRFTProperties` is not necessary, this collection only lives in the `iterateScheduleSection`method and the functionality can be handled with `std::pair`:
```C++
std::vector<std::pair<DeckKeywordConstPtr , size_t> > rftProperties;
...
rftPreperties.push_back( std::make_pair( keyword , currentStep ));
....
for (auto rftPair = rftProperties.begin(); rftPair != rftProperties.end(); ++rftPair) {
     DeckKeywordConstPtr keyword = rftPair->first;
     size_t timeStep = rftPair->second;
}
```

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -587,6 +611,81 @@ namespace Opm {
...
+            const std::string& wellNamePattern = record->getItem("WELL")->getTrimmedString(0);
+            const std::vector<WellPtr>& wells = getWells(wellNamePattern);
+
+            for (auto wellIter=wells.begin(); wellIter != wells.end(); ++wellIter) {
+                WellPtr well = *wellIter;
+                well->setRFT(currentStep, true);
+            }
+        }
+
+        const std::vector<WellPtr>& wells = getWells("*");
```
joakim-hove: `getWells( )` will return all wells, no need to pattern match with '*'.

opm/parser/eclipse/EclipseState/Schedule/Well.cpp
```diff
@@ -232,6 +237,47 @@ namespace Opm {
...
+    }
+
+    bool Well::getPLT(size_t time_step) const{
+     return m_plt->get(time_step);
+    }
+    void Well::setPLT(size_t time_step, bool value){
+        m_plt->add(time_step, value);
+    }
+
+    int Well::findWellFirstOpen(int startTimeStep) const{
```
joakim-hove: Not necessary to use `this` in `findWellFirstOpen()` and `setRFTForWellWhenFirstOpen()`

opm/parser/eclipse/EclipseState/Schedule/Well.cpp
```diff
@@ -107,14 +113,13 @@ namespace Opm {
         return m_status->get( timeStep );
     }
 
-    void Well::setStatus(size_t timeStep, WellCommon::StatusEnum status) {
+    void Well::setStatus(size_t timeStep, const WellCommon::StatusEnum status) {
```
joakim-hove: No harm - but in my opinion not necessary to set `const` on value types.

opm/parser/eclipse/EclipseState/Schedule/Well.hpp
```diff
@@ -86,6 +86,15 @@ namespace Opm {
         WellPolymerProperties          getPolymerPropertiesCopy(size_t timeStep) const;
         const WellPolymerProperties&   getPolymerProperties(size_t timeStep) const;
 
+        bool getRFT(size_t time_step) const;
```
joakim-hove: I guess these should be called something like `getRFTActive() ` - because what you get/set is a flag as to whether a simulated RFT should be stored - not the actual RFT.

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -587,6 +611,81 @@ namespace Opm {
...
+        for (auto wellIter=wells.begin(); wellIter != wells.end(); ++wellIter) {
```
joakim-hove: This looks a bit weird:

1. Set the RFT flag from now and the rest of the simulation for all wells matching the name.
2. Set all the wells to output an RFT when they first open

PR #414, opened 2015-02-06
Opm 166 Internalize thresholdpressure

chflo:
Extended EclipseState with threshold pressure values:

Added new classes SimulationConfig and ThresholdPressure.

EclipseState has a new member, - an instance of the new SimulationConfig class. 
A get method for fetching this object is added to EclipseState. 

From this SimulationConfig object a thresholdpressure vector can be fetched. 
The internalizing of the THPRES values are located in a separate class ThresholdPressure. 
SimulationConfig has got a member object of type ThresholdPressure.

Added tests for new class SimulationConfig and new class ThresholdPressure.
 
Testing on Norne in progress.


opm/parser/eclipse/EclipseState/EclipseState.cpp
```diff
@@ -305,6 +309,13 @@ namespace Opm {
         schedule = ScheduleConstPtr( new Schedule(getEclipseGrid() , deck, logger) );
     }
 
+    void EclipseState::initSimulationConfig(DeckConstPtr deck) {
+        auto eqlnum = getIntGridProperty("EQLNUM")->getData();
+        int maxEqlnum = *max_element(eqlnum.begin(), eqlnum.end());
+
+        m_simulationConfig = SimulationConfigConstPtr(new SimulationConfig(deck, maxEqlnum));
```
joakim-hove: Use `std::make_shared : m_simulationConfig = std::make_shared<const SimulationConfig>( deck , maxEqlnum));`

opm/parser/eclipse/EclipseState/SimulationConfig/ThresholdPressure.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+
+    private:
+
+        void initThresholdPressure(std::shared_ptr<const RUNSPECSection> runspecSection,
+                                   std::shared_ptr<const SOLUTIONSection> solutionSection,
+                                   int maxEqlnum);
+
+
+        std::vector<double> mThresholdPressureTable;
```
joakim-hove: Stick to the naming convention: `m_thresholdPressureTable`

opm/parser/eclipse/EclipseState/SimulationConfig/ThresholdPressure.cpp
```diff
@@ -0,0 +1,110 @@
...
+                    }
+                }
+            }
+        }
+
+
+        // Do we have the THPRES keyword?
+        // Check for consistency.
+        const bool thpresKeyword = solutionSection->hasKeyword("THPRES");
+        if (thpresKeyword != thpresOption) {
```
joakim-hove: This is too strict; no reason to raise an exception if an unused `THPRES` keyword is lying around in the deck.

joakim-hove:
The PR has good testing!

joakim-hove:
Should have a test creating a full `EclipseState` based on a deck with `EQLNUMS` input, instead of the hardcoded value of 3.

opm/parser/eclipse/EclipseState/EclipseState.cpp
```diff
@@ -305,6 +309,13 @@ namespace Opm {
         schedule = ScheduleConstPtr( new Schedule(getEclipseGrid() , deck, logger) );
     }
 
+    void EclipseState::initSimulationConfig(DeckConstPtr deck) {
+        auto eqlnum = getIntGridProperty("EQLNUM")->getData();
+        int maxEqlnum = *max_element(eqlnum.begin(), eqlnum.end());
```
joakim-hove: Write out the `std` in `std::max_element` explicitly - that is the way it is done everywhere else.

joakim-hove:
Mostly only trivial issues; however one slightly more significant issue: Instead of calculating the `maxEqlnum` in the `EclipseState`and then passing that on, I would prefer if the constructor got a reference/pointer to the `m_intGridProperties`. There are two reasons for that:

1. The call `EclipseState::getIntGridProperty("EQLNUM")` will autocreate a `ECLNUM` grid property if no one is in the deck - we do not need/want that side effect here.
2. The different arguments to the `SimulationConfig` and `ThresholdPressure` are at verye different levels of abstraction - looks kind of funny.

chflo:
After changes due to review: added new PR. Closing this PR.

PR #416, opened 2015-02-09
Keyword MINPVF

qilicun:
As we discussed in one Hangout, we should provide a new customs keyword for OPM behavior of "MINPV".
For the detail you can found from https://github.com/OPM/opm-core/pull/734 .

@alfbr was suggested we use MINPVFILL as the keyword, but @atgeirr said it exceeds the maximum length, so MINPVF is just the abbreviation. If you have any ideas about the keyword names, please fill free to make comments.


alfbr:
I still think MINPVFILL is a good choice, but if we are keeping the eight character limit, then I suggest FILLPV as a more intuitive choice, or even MINPVFIL. I am afraid MINPVF will be confusing and hard to remember.

qilicun:
I think MINPVFIL is a good choice, since it indicates that this keyword has relation to MINPV, but FILLPV  doesn't. If there is nobody against MINPVFIL, I will update the PR.

qilicun:
Or should we make a vote for all the candidates...?

joakim-hove:
Regarding name:

1. I would vote for `MINPVFIL` - as the funny eight character spelling is according to ECLIPSE tradition; but I have no strong feelings.
2. The eight character limit is quite hard; I do not see that we can relax that without breaking the normal ECLIPSE semantics.

joakim-hove:
I feel inclined to say that instead of having two boolean settings for the different MINPV behaviour we should have and enum and a value (attach no significance to my naming suggestions):

```C++
enum MinPVMode {
    NOT_ACTIVE = 0,
    ECLIPSE_STANDARD = 1,
    OPM_FILLING= 2
    }

    double minPVValue;
```

Would be good if the maintainers of the downstream modules voiced their opinions on this.

andlaus:
I don't have any opinion on this issure itself, but I could not find it in the RM, so you should probably move it to a new directory `002_Opm` (or similar) instead of using `000_Eclipse100`? (I don't personally care about this too much, though.)

atgeirr:
> I feel inclined to say that instead of having two boolean settings for the different MINPV behaviour we should have and enum and a value

I agree, the parser should ensure that either MINPV or MINPVFIL is accepted, not both.

I also think using a new directory for new keywords is a good idea, was not that the major reasons to create them in the first place?

joakim-hove:
>I agree, the parser should ensure that either MINPV or MINPVFIL is accepted, not both.

OK @qilicun - seems we agree on this, so good if you change the solution to use an enum and normal double value. Then I will merge.

qilicun:
@joakim-hove , Sounds good, I will update the PR.

qilicun:
@joakim-hove 
Do you mind that method isMinpvAcitve() and member m_minpv are removed and using getMinpvMode() instead?

joakim-hove:
That was what I suggested.
10. feb. 2015 09:49 skrev "Liu Ming" <notifications@github.com> følgende:

qilicun:
@joakim-hove I have update the PR.  could you please review it and make comments?

opm/parser/eclipse/EclipseState/Grid/MinpvMode.hpp
```diff
@@ -0,0 +1,34 @@
...
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef _MINPVMODE_
+#define _MINPVMODE_
+
+#include <string>
+
+namespace Opm {
+
```
joakim-hove: 1. Please put the enum in a separate namespace - see e.g. the `ScheduleEnums.hpp` file.
2. Unless I am completely mistaken these values should be used one-at-a-time; i.e. not as bitmasks. The use of numerical values 1,2,4,  hints at bitmask usage and can be misleading, please just use 1,2,3 or 0,1,2

joakim-hove:
Except the very minor comment about the enum this is just what the doctor ordered - and I'm ready to merge. 

@atgeirr: Please give a thumbs up here when you are ready to merge the downstream modules.

qilicun:
@joakim-hove 

Ok, will fix that. Since we agree on this , I will make the updates to downstream module soon.

joakim-hove:
Good - starting the merge fest.

PR #417, opened 2015-02-12
allow the schedule object to be immutable

andlaus:
quite a few important methods (e.g., getWells()) missed the variants
which are marked as 'const'. this leads to the problem that these methods
can't be called on the object returned by EclipseState::getSchedule()
because it returns a ScheduleConstPtr. (this issue can possibly also be
worked-around by creating a new mutable copy of the schedule object
from the one returned by EclipseState::getSchedule(), but to me, this
seems to be a strikingly inelegant and inefficient hack...)

this fixes the compilation of the ECL blackoil simulator of eWoms...

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
@@ -719,8 +719,8 @@ namespace Opm {
         return m_wells.size();
     }
 
-    size_t Schedule::numWells(size_t timestep) {
-      std::vector<WellPtr> wells = getWells(timestep);
+    size_t Schedule::numWells(size_t timestep) const {
+      std::vector<WellConstPtr> wells = getWells(timestep);
       return wells.size();
     }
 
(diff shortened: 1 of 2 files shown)


atgeirr:
I get error messages about this in opm-core, was there a recent API change?

joakim-hove:
Område post details, I think Andreas is totally innocent.
12. feb. 2015 13:18 skrev "Atgeirr Flø Rasmussen" <notifications@github.com>
følgende:

joakim-hove:
Thanks for patch. We have f...ed up a bit on this; I'll make sure I understand the issue before I mess it up even further.

atgeirr:
I can confirm that this patch fixes the problem for me. I have not looked at the code much, it would be good to make the new (non-const) overloads private since that seems to be where they were needed.

andlaus:
@akva2 
> seems this is cut and paste coding. is there really any non-const access in the code you have c&p'd here? if not, just keeping the const versions would be the saner approach.. also it yields a much smaller diff.

you're right about the copy-n-paste stuff. the problem is that the alternatives are:

a) call the non-`const` version of the methods from the const ones and then copy the pointers in the vectors. this requires a `const_cast` of the 'this' pointer.
b) the same as (a), but the other way around (this requires a `const_cast` of the pointers contained by the vector).

both of these are IMO not better than copy-n-paste, but if somebody voices a preference, I'm open to switch...

PR #421, opened 2015-02-12
Schedule: fix handling of the RESV item of the WCONINE keyword

andlaus:
it turns out that, in contrast to the RATE item, this item is always a
liquid rate (-> bbl/day in FIELD units), i.e., the dimension is always
"LiquidVolume/Time" irrespectibly of the fluid phase.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
@@ -293,7 +293,7 @@ namespace Opm {
 
 
                 if (!record->getItem("RESV")->defaultApplied(0)) {
-                    properties.reservoirInjectionRate = convertInjectionRateToSI(record->getItem("RESV")->getRawDouble(0) , injectorType, *deck->getActiveUnitSystem());
+                    properties.reservoirInjectionRate = record->getItem("RESV")->getSIDouble(0);
                     properties.addInjectionControl(WellInjector::RESV);
                 } else
                     properties.dropInjectionControl(WellInjector::RESV);
--- a/opm/parser/share/keywords/000_Eclipse100/W/WCONINJE
+++ b/opm/parser/share/keywords/000_Eclipse100/W/WCONINJE
@@ -4,7 +4,7 @@
     {"name" : "STATUS"       , "value_type" : "STRING" , "default" : "OPEN"},
     {"name" : "CMODE"        , "value_type" : "STRING" },
     {"name" : "RATE" , "value_type" : "DOUBLE" , "dimension" : "ContextDependent"},
-    {"name" : "RESV" , "value_type" : "DOUBLE" , "dimension" : "ContextDependent"},
+    {"name" : "RESV" , "value_type" : "DOUBLE" , "dimension" : "LiquidVolume/Time"},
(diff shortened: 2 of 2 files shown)


atgeirr:
Actually, I think this is not quite precise: it is a reservoir volume, and if we want to be consistent with ecl we must treat it as such. So instead of two unit-like quantities (liquid volume and gas volume) we have four (same two, but each can be at surface or reservoir). For METRIC they are all the same (if we accept reservoir and surface cubic meters to be the same) and in FIELD units they are all barrels except for surface gas volume. The old EclipseGridParser had this in its unit system, and I think we should have that here as well. Gas/oil ratios (rs) and similar quantities with units that are derived from these should be redone with this set of four units as well, to ensure consistency.

andlaus:
> I think this is not quite precise: it is a reservoir volume,

well, in retrospect the name of the dimension `LiquidVolume` is a bit unfortunate: it should better be `LiquidOrReservoirVolume`. alternatively, a new unit named `ReservoirVolume` could be introduced which would happen to exhibit the same value as `LiquidVolume`. (I'm not sure if this is consistent for all keywords that specify reservoir volumes, though.)

> So instead of two unit-like quantities (liquid volume and gas volume) we have four (same two, but each can be at surface or reservoir). For METRIC they are all the same (if we accept reservoir and surface cubic meters to be the same) and in FIELD units they are all barrels except for surface gas volume.

for some reason which beats me, the reservoir volumes are also the same for all fluid phases even if using the FIELD unit system, so `GasReservoirVolume` would be superfluous. (also, if your suggestion is implemented consequently, there should probably be 6 instead of 4 different volume units: one for each combination of phase for and volumes type.) But maybe that's dependent on the keyword, I don't know... 

> Gas/oil ratios (rs) and similar quantities with units that are derived from these should be redone with this set of four units as well, to ensure consistency.

be aware that these two currently have dedicated dimensions (`{Oil,Gas}DissolutionFactor`, just like e.g. transmissibility) and should be correct for this reason. Also, I think that this should stay this way as it is more flexible and also more expressive...

andlaus:
BTW: are you sure this bug here is the cause that well controls are switched back and in SPE-1? I'd say that this makes it _less_ likely to switch controls based on reservoir rates as the maximum reservoir rates are too large by a factor of 178 and the problem is basically surface rate limited for all the time...

atgeirr:
> BTW: are you sure this bug here is the cause of well controls being switched back and forth in SPE-1? I'd guess that this bug makes it less likely to switch controls based on reservoir rates as the maximum reservoir rates are too large by a factor of 178 and the problem is basically surface rate limited at all times...

No, I am not sure, but it certainly is a bug. I'll keep investigating this until I get to the bottom of it though...

joakim-hove:
OK - this is less than clear to me. Thank your for contributions and discussions - I will look carefully into it; today or next week.

atgeirr:
I have not tested this PR, but it looks like a decent workaround. I do not think it is a good long-term solution, but that does not mean it should not be merged in the meantime. No point in preventing something good waiting for the best. I will test this and report back.

andlaus:
I'm pretty sure that the patch is correct, but I agree with atgeirr that the name `LiquidVolume` is confusing and should be renamed (in this context).

joakim-hove:
@atgeirr confirms that this works for him; merging now - might come back to it.

PR #422, opened 2015-02-13
Correct reservoir volume handling

andlaus:
this is a follow-up of #421. Note that the second patch fixes a pretty bad bug in PORV for decks that use field units...


joakim-hove:
Thank you for the effort; you are probably right about `PORO` dimension - but it is not entirely clear to me how you determined that? But I guess the same fix should apply to `PORV` as well?

@atgeirr could you have a look at this.

andlaus:
> but it is not entirely clear to me how you determined that?

simple: I grepped the RM for all keywords that contained at least one item which required reservoir volume. PORV was one of them which appeared in the greping efforts but the dimension in the JSON was quite ackward...

joakim-hove:
>PORV was one of them which appeared in the greping efforts

Meaning that the Json configuration should be updated also for `PORV`?

atgeirr:
It looks reasonable, but I would not merge until tested first. I am also currently struggling with getting the correct SPE1 results, so I may have been to quick to approve the previous PR, I will post my findings. Meanwhile, I hope it's ok to be patient...

opm/parser/share/keywords/000_Eclipse100/C/COMPDAT
```diff
@@ -6,7 +6,7 @@
     {"name" : "K2" , "value_type" : "INT" },
     {"name" : "STATE" , "value_type" : "STRING" , "default" : "OPEN"},
     {"name" : "SAT_TABLE" , "value_type" : "INT" , "default" : 0},
-    {"name" : "CONNECTION_TRANSMISSIBILITY_FACTOR"     , "value_type" : "DOUBLE", "dimension" : "Viscosity*LiquidVolume/Time*Pressure"},
+    {"name" : "CONNECTION_TRANSMISSIBILITY_FACTOR"     , "value_type" : "DOUBLE", "dimension" : "Viscosity*ReservoirVolume/Time*Pressure"},
```
atgeirr: Just an off question here: how do you parse this expression? Is there a parser inside the opm-parser...? To me the expression looks indefinite, unless one assumes left associativity, in which case it is wrong. So I just do not understand how this can end up correct! (Apparently, it does, or have we not tested FIELD cases with given factors?)

andlaus:
> Meaning that the Json configuration should be updated also for PORV?

that's what 6385dd68c643b760dc4c58008cb067bfb3284659 does, right?

opm/parser/share/keywords/000_Eclipse100/C/COMPDAT
andlaus: that's a question best answered by joakim, but yes there is an interpreter for dimension strings. (it has a few constraints like no support for brackets and only a single division, IIRC.) basically everrything which comes before the `/` is evaluated indepenently from the stuff right to it and then the two results are divided. I could be wrong, though...

joakim-hove: @andlaus sums it up quite nicely; there is an extremely basic parser which understands '*' and *one* '/': https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/Units/UnitSystem.cpp#L99

joakim-hove:
> that's what 6385dd6 does, right?

Yes - you are absolutely right. What tripped me off was that the PR message mentioned `PORO` whereas the patch correctly updated `PORV`

atgeirr:
I have investigated the SPE1 issue, and will be posting an issue in opm-autodiff about it. I do not think the parser code is affected or involved.

andlaus:
> Yes - you are absolutely right. What tripped me off was that the PR message mentioned PORO whereas the patch correctly updated PORV

oops sorry. This is a case of "I read what I meant, not what I've written" ;)

joakim-hove:
What is the status of this; should it be merged?

atgeirr:
Yes I think this should be merged. It should fix a bug as @andlaus said, and the unit treatment makes more sense with this. I think it might be possible to go a little further in removing custom treatments of units (like those in Schedule) in favor of using the getSIDouble() method, but that need not delay this.

Note: I have not actually built and run this branch, but I reviewed the code.

joakim-hove:
>Yes I think this should be merged.

OK - thanks; merging.

PR #423, opened 2015-02-13
Remove some warnings when compiling the parser module.

dr-robertk:
The warnings are:

sizeType --> p_sizeType since it shadows a member method
trailing ;
comparison of signed and unsigned int


atgeirr:
I just did some of these things myself, as I did not realize there were already a PR. It would be good if this could be merged.

joakim-hove:
> It would be good if this could be merged.

Sorry - this fell through the cracks.

Merging now.

PR #424, opened 2015-02-17
Add missing throw check, also silencing warning.

atgeirr:
The function createMinpvDefaultCPDeck() was never used. I have just added the test that seemed to be forgotten.


joakim-hove:
Thanks - this was a merge fallout.

PR #427, opened 2015-03-02
Implement temperature dependent density

andlaus:
this adds the stuff which is required to implement a "pseudo-thermal" black oil simulator. The main thing to be noted is that some Eclipse keywords use thermodynamic temperatures (Rankine or Kelvin) instead of "ordinary" ones (Fahrenheit or Celsius). This makes it necessary to introduce an "AbsoluteTemperature" unit...

### DIFF
--- a/opm/parser/eclipse/EclipseState/Tables/OilvisctTable.hpp
+++ b/opm/parser/eclipse/EclipseState/Tables/OilvisctTable.hpp
@@ -49,7 +49,7 @@ namespace Opm {
             ParentType::checkMonotonic("Temperature", /*isAscending=*/true);
 
             ParentType::checkNonDefaultable("Viscosity");
-            ParentType::checkMonotonic("Viscosity", /*isAscending=*/true, /*strictlyMonotonic=*/false);
+            ParentType::checkMonotonic("Viscosity", /*isAscending=*/false, /*strictlyMonotonic=*/false);
         }
 
     public:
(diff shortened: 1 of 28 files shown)


joakim-hove:
This looks innocent enough - will build test++ tomorrow, and I guess we should double check the density inconistency? 

Is there a plan for including thermal properties++ in the (Eclipse)State object?

andlaus:
> This looks innocent enough - will build test++ tomorrow, and I guess we should double check the density inconistency?

thanks, more testing and checking is always a good idea. on the checking front, be aware that -- as far as I know -- the Eclipse inconsistency is for how temperature is specified, not density...

> Is there a plan for including thermal properties++ in the (Eclipse)State object?

besides the tables, I currently don't have such plans at the moment: Most of these keywords are pretty simple, and my focus so far was to get the sh** for pseudo-thermal support done. That said, some of the keywords have a variable-but-fixed number of items per record which depends on the value specified by COMPS. To my knowledge this is currently not supported by the low-level parser and if you have a good idea how to improve the current state of affairs, I won't be opposed to implement it.

joakim-hove:
> [...]  some of the keywords have a variable-but-fixed number of items per record which depends on the value specified by COMPS. To my knowledge this is currently not supported by the low-level parser

That is right.

>and if you have a good idea how to improve the current state of affairs, I won't be opposed to implement it.

Not really - I'll merge it now; if this turns out to be a problem we will have to fix it.

andlaus:
that was quick. thanks :)

PR #435, opened 2015-03-06
Remove surplus const for return-by-value.

atgeirr:
This triggers compiler warnings at high warning levels.


andlaus:
IMO such surplus `const`s should also be removed for parameters (if they exist in opm-parser, I know that they do exist in opm-core and opm-autodiff), because whatever reason the compiler has for issuing the warning on the return values also applies to parameters.

PR #436, opened 2015-03-09
Fixed bug with iterator modification

joakim-hove:
Thanks detailed review & testing by @bska.

Would be good if @bska could test this - because it "works for me" anyway.


bska:
> Would be good if @bska could test this

This fixes the `make test` failures I got following #398.

PR #437, opened 2015-03-16
added: build ERT in source tree if not found

akva2:
This is a POC of what was discussed at the opm seminar. This will build ERT in the opm-parser build tree if it is not found externally.

Note that I made ERT REQUIRED. i think it actually is and that the statement in CMakeLists.txt is wrong.

also 
```
if (condition)
else(condition)
endif(condition)
```
is evil, as highlighted by the unnecessary diff lines ;)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,7 +58,7 @@ endif()
 include("cmake/Modules/CheckCaseSensitiveFileSystem.cmake")
 add_definitions("-DHAVE_CASE_SENSITIVE_FILESYSTEM=${HAVE_CASE_SENSITIVE_FILESYSTEM}")
 
-find_package(ERT)
+find_package(ERT REQUIRED)
 include_directories( ${ERT_INCLUDE_DIRS} )   
 
 set(LIBRARY_OUTPUT_PATH ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR})
--- a/cmake/Modules/FindERT.cmake
+++ b/cmake/Modules/FindERT.cmake
@@ -108,6 +108,30 @@ list (APPEND ERT_LIBRARY
   ${ERT_LIBRARY_GEOMETRY}
   ${ERT_LIBRARY_UTIL}
   )
+
+if(ERT_FIND_REQUIRED AND NOT ERT_LIBRARY)
+    message(STATUS "ERT not found, will be built in source tree")
+    include(ExternalProject)
+    externalproject_add(ert
+                        GIT_REPOSITORY https://github.com/Ensembles/ert
+                        GIT_TAG origin/master
+                        CONFIGURE_COMMAND ${CMAKE_COMMAND} <SOURCE_DIR>/devel
+                                          -DBUILD_SHARED_LIBS=0
+                                          -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+                                          -DCMAKE_INSTALL_PREFIX=${CMAKE_BINARY_DIR}/ert
+                                          -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}
+                                          -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
+                                          -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
+                        PREFIX ${CMAKE_BINARY_DIR}/ert)
+    set(ERT_LIBRARY ${CMAKE_BINARY_DIR}/ert/${CMAKE_INSTALL_LIBDIR}/libecl.a
+                    ${CMAKE_BINARY_DIR}/ert/${CMAKE_INSTALL_LIBDIR}/libecl_well.a
+                    ${CMAKE_BINARY_DIR}/ert/${CMAKE_INSTALL_LIBDIR}/libert_geometry.a
+                    ${CMAKE_BINARY_DIR}/ert/${CMAKE_INSTALL_LIBDIR}/libert_util.a)
+    set(ERT_INCLUDE_DIR ${CMAKE_BINARY_DIR}/ert/include)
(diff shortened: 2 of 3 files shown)


atgeirr:
Cmake docs (from 2.8.0 on) state that repeating the condition is optional, so why not get rid of it? Or give a meaningful name to it and use that (unless I am mistaken in my belief that CMake is a relatively normal programming language)?

akva2:
yup, it can be dropped since 2.8. i just pointed it out, not my choice.

joakim-hove:
Thanks - looks fine. How will this work in the following situation:

1. I clone opm-parser; no ERT is found and the automatic download machinery downloads and builds ERT.
2. At a later stage both ERT & opm-parser are updated, and HEAD of opm-parser no longer works with the in-build installation of ERT?

> is evil, as highlighted by the unnecessary diff lines ;)

The evil-ness is related to the `(condition)` block - i.e. you would have preferred `()`? That would be fine by me.

The variable `${LIBRARY_DEPS}` - is that cmake magic trickery?

akva2:
currently it points to origin/master of ert. this cannot be like this - it needs to point to a specific revision, which needs to be updated whenever a new ERT is required. if you like this way of doing things, i'll likely put that in a separate (text)file for simplicity and it will be hosted in opm-cmake. whenever that revision is changed, cmake picks up on it, reconfigures and builds the new revision (with the appropriate DEPENDS statements if separated to a text file).

LIBRARY_DEPS is there to fix threaded builds - we need to declare the dependency on ert if built in the tree so ert is built before parser or it won't find the headers..

akva2:
oh, and so i don't confuse you; if we keep the origin/master, ert will be re-built whenever you do 'make' in opm-parser and something has been pushed to ert master so there is no problem with doing what you say. the text file isn't necessary for the technical side as such.

but it's very bad from a floss (and dev) point of view to not reproduce exactly the same build when i build a given revision of opm-parser.

joakim-hove:
OK - I have tested this and it works for me ;-) Some more questions:

1. As you say it will rebuild ert for every invocation of `make` - that is kind of wasteful?
2. Both `opm-parser` and `opm-core` have explicit dependencies on ert, so naively this would require duplication of the in-build build of ert?

Could it be an config option to supply an alternative root for the ERT build? Then the downstream modules could use the ert installation created by the opm-parser build process?

Just for my own understanding: Is this type of build meant as a developer convenience, or do envision this as an approach leading to packages as well?

akva2:
first of all it needs to be combined with opm-cmake to make sense, so we don't have to replicate even more stuff in all them tree's. we'll then pull in opm-cmake into the build tree on configure time if it's not found on the system. i'll handle the transitional stuff as i climb the dependency tree, i got a plan for it, so you will only be building a single ert. that part isn't visible yet for obvious reasons :)

it only builds ert on every invocation due to a combination of
1) pointing to origin/cmake
2) the ERT buildsystem is slightly broken - it directly depends on add_custom_target(OUTPUT blablabla) files. which is a big nono, you need to add a target depending on the output, then depend on that target since an OUTPUT target is always outdated (this bites even harder in opm-parser btw). it's a silly cmake-ism.

i consider these kind of build mostly a non-developer convenience, and maybe as a bone for jenkins setups. it's definitely not for packaging!

atgeirr:
Can this be the start of a universal treatment of such libs? So that this and the similar code for Eigen are just two cases using the same system?

akva2:
absolutely. it can be extended to every single dependency if we want to.

akva2:
a bit unrelated, but this shows the ideas in another system i have done that pulls off the same and more. https://github.com/notspiff/visualization.projectm/tree/master/depends/common

it will be a spin on this, probably slightly less general to suite the task at hand.

joakim-hove:
> first of all it needs to be combined with opm-cmake to make sense

OK - I see.

 > i'll handle the transitional stuff as i climb the dependency tree, i got a plan for it

OK - good.

>the ERT buildsystem is slightly broken - it directly depends on add_custom_target(OUTPUT blablabla) files. which is a big nono [...]

OK - I'd be more than happy to improve both the ERT and opm-parser build systems, but would need some advice to get it going - getting these `add_custom_xxx()` things to work in cmake is like chewing gravel.

> i consider these kind of build mostly a non-developer convenience, and maybe as a bone for jenkins setups. it's definitely not for packaging!

OK - good to know. I see something like this as the basis for an improved build system:

1. Each module opm-xxx maintains it's own `opm-xxx-prereqs.cmake` file, and utilizes cmake modules in opm-cmake - essentially the current build system without the duplication.

2. The opm-cmake module has a `global/CMakeLists.txt` which can handle managed builds of all modules. That way the the simplest way to build the `flow` simulator would be:
```txt
$ git clone opm-cmake
$ mkdir build ; cd build
$ cmake ../opm-cmake/global
```

Makes sense?

akva2:
get off my page! ;)

akva2:
i'll send patches for the buildsystems of ert and parser once i have some more dead time to kill... at least showing how to do it it, might not fully complete it.

joakim-hove:
> i'll send patches for the buildsystems of ert and parser once i have some more dead time to kill

Thanks a lot :-)

joakim-hove:
I guess the build system has evolved so much since march 2015 that there is no point in keeping this around any longer?

alfbr:
Not sure what the current state is, but a mechanism to automatically download and build ert if it is not found is very valuable. At least until we can get it stabilized and packaged in distributions.

atgeirr:
This PR does not make sense here anymore since it modifies FindERT.cmake which now lives in opm-common, and should be closed. But @akva2: do you intend to move the essential parts of this PR to opm-common? It would not go in before the release anyway, but it would be good to know your thoughts on this.

akva2:
as far as i know this approach has been deemed unwanted. in any case 
it's contained in the OpmStaticTargets.cmake which already sits in 
opm-common, so any further implementation will piggy-back on that code.

PR #439, opened 2015-03-18
Using external cmake modules

joakim-hove:
With this PR opm-parser can be built with an external buildsystem from opm-cmake


iLoop2:
Hi Joakim
Just tested this and it works fine on OS X.

PR #440, opened 2015-03-21
Sof3

joakim-hove:
This PR is a first step towards using the SOF3 keyword for relperm input.

This is here mainly as a placeholder; not ready for merge.


andlaus:
> This PR is a first step towards using the SOF3 keyword for relperm input.

would you please elaborate a bit more on that? in particular why do you think this PR is needed for SOF3 while it hasn't been for SOF2? (I can't remember any issues with what's currently available for SOF2?)

joakim-hove:
>would you please elaborate a bit more on that? in particular why do you think this PR is needed for SOF3 

Well - for first of all this is the first part in my five step program called "*Understanding the table code in five simple steps*" - so you are right, it might not be needed in the end.

The concrete situation which led me to introduce the little `Tabdims` class was the code which determines the number of saturation functions by asking the `swofTables` vector; I am currently looking at a deck which does not have the `SWOF` keyword at all - and the `assert()` here fails: https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Grid/GridPropertyInitializers.hpp#L203

andlaus:
>Well - for first of all this is the first part in my five step program called "Understanding the table code in five simple steps" - so you are right, it might not be needed in the end.

Okay. I was just wondering. For understanding (or at least using) the tables code, I'd guess the quickest way is to just copy another table class (e.g., the one for SWOF) and adapt it to SOF3...

> The concrete situation which led me to introduce the little Tabdims class was the code which determines the number of saturation functions by asking the swofTables vector; I am currently looking at a deck which does not have the SWOF keyword at all - and the assert() here fails

hm, IIRC, the actual tables code does not acces TABDIMS at all, it is only accessed by the low-level parser. (which is also the reason why I would prefer an "elastic" size type for tables, btw...)

for the concrete assertation in the initializer for the endpoints, it seems that this class needs to be considerably changed if SWOF cannot assumed to be not present because the number of tables is far from the only quantity retrieved from `SwofTable`. Probably this would need to go as far as internalizing the full relperm and capillary pressure relations, but I'm unsure if the parser is the correct place for this...

joakim-hove:
>for the concrete assertation in the initializer for the endpoints, it seems that this class needs to be considerably changed if SWOF cannot assumed to be not present because the number of tables is far from the only quantity retrieved from SwofTable. 

Yes - I can see that.

> Probably this would need to go as far as internalizing the full relperm and capillary pressure relations, but I'm unsure if the parser is the correct place for this...

Yes - probably; I think the parser is the right place for this - however the name *parser* is becoming more and more misplaced; I guess `opm-input` would be more suitable. My clear ambition is that the `EclipseState` class should normalize the input, so that consuming classes can get relperm, grid, rates and so on independently of the specific keyword combinations used in the input deck.

Anyway - I'm logging off for at least a week about now, so I will not be pushing this in the immediate future.

totto82:
@joakim-hove Norne simulates without problems. The time-stepping and convergence are slightly different and thus also results at some point in time, but the differences are not significant.

atgeirr:
I am curious: why does this PR change simulation results at all?

joakim-hove:
Ehhh - so am I; it should be by identity equal - so the differences (even though they are small) indicate a bug in my code :angry:

alfbr:
Keep in mind that we have seen small variations in results just by changing compiler, so this may be insignificant numerical noise.

totto82:
Let me first double check that I used a clean opm-core on my reference solution. There seems to be slight differences +-e20 in the well cntrl residual from the first iteration. I will report back.   .

totto82:
My current clean run agrees with the reference solution. I.e. the changes are triggered by this PR. The simulations agrees (within numerical noise) until 329 days when the version that includes this PR causes the CPRpreconditioner to fail. (That happens very rarely)

joakim-hove:
Thank you very much - I will update.

totto82:
But I agree with Alf that this just demonstrates that our current simulator is very sensitive to small changes. And it may be worthwhile to investigate what causes this instability.

GitPaean:
I pulled this PR with 
```
git checkout -b joakim-hove-SOF3 master
git pull https://github.com/joakim-hove/opm-parser.git SOF3
```
with the updated master and updated ERT. 

I could not compile successfully.  Did anyone see this problem?

```
In file included from /home/kaib/OPM/build/opm-parser/opm/parser/eclipse/EclipseState/EclipseState.cpp:33:0:
/home/kaib/OPM/build/opm-parser/opm/parser/eclipse/EclipseState/Grid/SatfuncPropertyInitializers.hpp: In instantiation of ‘void Opm::ISWCREndpointInitializer<EclipseState, Deck>::apply(std::vector<double>&) const [with EclipseState = Opm::EclipseState; Deck = Opm::Deck]’:
/home/kaib/OPM/build/opm-parser/opm/parser/eclipse/EclipseState/EclipseState.cpp:1404:1:   required from here
/home/kaib/OPM/build/opm-parser/opm/parser/eclipse/EclipseState/Grid/SatfuncPropertyInitializers.hpp:598:73: error: ‘ImbnumApply’ was not declared in this scope, and no declarations were found by argument-dependent lookup at the point of instantiation [-fpermissive]
         ImbnumApply(values , "SWCRIT" , this->m_criticalOilOWSat , false);

```

totto82:
I didn't. I have a few weeks old ERT (The release version) .

GitPaean:
I think it is not related to ERT. I saw this problem with updated master branch first then I updated ERT.

joakim-hove:
Ok thanks - more problems:-(
5. mai 2015 10:22 skrev "Kai Bao" <notifications@github.com> følgende:

joakim-hove:
OK  - I have fixed a bug and pushed an update. Looks correct to me - but would be very grateful for a new test run from @totto82.

joakim-hove:
> I could not compile successfully. Did anyone else also see this problem?

Works for me - ought to be enough for everyone :grin: Seriously I will try with another compiler; what compiler are you using?

GitPaean:
>> Works for me - ought to be enough for everyone :grin: Seriously I will try with another compiler; what compiler are you using?

Hi, the problem stays the same to me. I updated the master branch and ERT this morning then pulled your PR. 

I am using dune 2.2.1 and gcc 4.8.2

GitPaean:
I tried on a Vagrant VM with dune 2.3.1 and gcc 4.8.2. The same problem happened. gcc version can be a possible reason now.

joakim-hove:
>Hi, the problem stays the same to me. I updated the master branch and ERT this morning then pulled your PR.

OK thank you; this should not in any way be related to ERT. My suspicion is gcc version - I will investigate that myself.

totto82:
@joakim-hove Now the first couple of steps are exactly the same.

atgeirr:
I took a look at this, and found out what is going on.

You are calling a super-class method in a derived class that is a template, and the super-class also depends on the template parameters. In that context C++ requires that you qualify variables and methods from the superclass with Super::methodname(), where Super is the name of the superclass (or a typedef more likely). Alternatively, writing this->methodname() is also acceptable.

Old compilers (gcc 4.4?) accepted this, but later ones are more strict (and correct). I do not know the rationale for these rules, but we have to obey them just the same.

The simplest solution is to add `this->` in front of calls to ImbnumApply(), SatnumApply() and selectValue().

A few notes: we should be consistent about method naming. It should be satnumApply() for example. Also, I get lots of warnings:
```
/Users/atgeirr/opm-build-debug/opm-parser/opm/parser/eclipse/EclipseState/Tables/Tabdims.hpp:47:25: warning: private field 'm_ntpvt' is not used [-Wunused-private-field]
        size_t m_ntsfun,m_ntpvt,m_nssfun,m_nppvt,m_ntfip,m_nrpvt;
                        ^
/Users/atgeirr/opm-build-debug/opm-parser/opm/parser/eclipse/EclipseState/Tables/Tabdims.hpp:47:33: warning: private field 'm_nssfun' is not used [-Wunused-private-field]
        size_t m_ntsfun,m_ntpvt,m_nssfun,m_nppvt,m_ntfip,m_nrpvt;
                                ^
/Users/atgeirr/opm-build-debug/opm-parser/opm/parser/eclipse/EclipseState/Tables/Tabdims.hpp:47:42: warning: private field 'm_nppvt' is not used [-Wunused-private-field]
        size_t m_ntsfun,m_ntpvt,m_nssfun,m_nppvt,m_ntfip,m_nrpvt;
                                         ^
/Users/atgeirr/opm-build-debug/opm-parser/opm/parser/eclipse/EclipseState/Tables/Tabdims.hpp:47:50: warning: private field 'm_ntfip' is not used [-Wunused-private-field]
        size_t m_ntsfun,m_ntpvt,m_nssfun,m_nppvt,m_ntfip,m_nrpvt;
                                                 ^
/Users/atgeirr/opm-build-debug/opm-parser/opm/parser/eclipse/EclipseState/Tables/Tabdims.hpp:47:58: warning: private field 'm_nrpvt' is not used [-Wunused-private-field]
        size_t m_ntsfun,m_ntpvt,m_nssfun,m_nppvt,m_ntfip,m_nrpvt;
```

joakim-hove:
> I took a look at this, and found out what is going on.

Thank you very much  - I will update.

joakim-hove:
This warning I do not understand:

```txt
/Users/atgeirr/opm-build-debug/opm-parser/opm/parser/eclipse/EclipseState/Tables/Tabdims.hpp:47:25: warning: private field 'm_ntpvt' is not used [-Wunused-private-field]
        size_t m_ntsfun,m_ntpvt,m_nssfun,m_nppvt,m_ntfip,m_nrpvt;
```

the private fields in question are initialized by the constructor initializer list?!

akva2:
On 06/05/15 10:44, Joakim Hove wrote:
>
> This warning I do not understand:
>
> |/Users/atgeirr/opm-build-debug/opm-parser/opm/parser/eclipse/EclipseState/Tables/Tabdims.hpp:47:25: warning: private field 'm_ntpvt' is not used [-Wunused-private-field]
>          size_t m_ntsfun,m_ntpvt,m_nssfun,m_nppvt,m_ntfip,m_nrpvt;
> |
>
> the private fields in question are initialized by the constructor 
> initializer list?
>
yes, but not used later. applies to everything but m_ntsfun right now 
afaict.

joakim-hove:
OK - I have followed Atgeirrs advice; had a bit problems getting my newer compiler to run - so I would be grateful if @GitPaean or @atgeirr could give it a build test.

alfbr:
On Red Hat 6 nodes (rgs and workstations) we have now access to gcc 4.9.2 (4.9.1 if nobody ran yum update) using developer toolset. Users can test it by using the "scl enable devtoolset-3" command. To have your whole terminal session using the updated tool chain is most convenient, done with:
$ scl enable devtoolset-3 'bash'

GitPaean:
It compiles now and produces identical results with the current master branch (Without finishing running Norne). 

Today's master branch produces a little different results from yesterday's branch though, while it is not related to this PR and not necessarily related to opm-parser.

joakim-hove:
> On Red Hat 6 nodes (rgs and workstations) we have now access to gcc 4.9.2 

Well yes - but then came boost problems. We'll get there.

joakim-hove:
> It compiles now and produces identical results with the current master branch

Thanks a lot - getting this in has been a real community effort!

alfbr:
> but then came boost problems. We'll get there.

Indeed, looking forward to it :-)

PR #441, opened 2015-03-25
OPM-87, added code for handling keyword WELTARG.

edbru:
Handling keyword WELTARG. This keyword is used to modify WellProductionProperties for the spesified wells - at the specified timestep. 
Modified files:
WellProductionProperties.cpp,
WellProductionProperties.hpp,
Schedule.cpp,
Schedule.hpp,
ScheduleTests.cpp.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
@@ -29,9 +29,7 @@
 #include <opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.hpp>
 #include <opm/parser/eclipse/EclipseState/Schedule/WellInjectionProperties.hpp>
 #include <opm/parser/eclipse/EclipseState/Schedule/WellPolymerProperties.hpp>
-
-
-
+#include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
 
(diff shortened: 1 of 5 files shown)


atgeirr:
I am a little curious as to what the "87" refers to?

edbru:
It is just a task number in our internal Jira system. It has no significance for the task at hand.

joakim-hove:
>No description provided.

Please provide a brief description of the PR; most interesting is a brief description of the problem the PR is addressing.

joakim-hove:
According to the `WELTARG` documentation the well must have been fully specified using one of the `WCONxxxx` keywords prior to the `WELTARG` keyword; that behaviour is currently not enforced by this `WELTARG` implementation. Might be OK - but should at least be commented in the code.

joakim-hove:
The default behavior if the value item is not supplied is *really weird* - we do not have to support that in the first version, but we should check for it and raise an exception in the unlikely situation that it is encountered: 

```C++
auto item = record->getItem("NEW_VALUE");
if (!item->hasValue())
   throw std::invalid_argument("Sorry - the value item in WELTARG can not be defaulted");
```

joakim-hove:
When the minor comments have been addressed this PR should be ready for merge.

joakim-hove:
Thanks - merging.

PR #442, opened 2015-03-25
Fix warnings

atgeirr:
This suppresses two unused argument warnings. With this, all of opm-parser compiles with no warnings at a very high warning level.


joakim-hove:
Thanks.

PR #446, opened 2015-04-10
Implemented NOSIM keyword..

edbru:


joakim-hove:
OK - now when the release is out I am ready to merge this. Can you please rebase on top of current upstream/master. Also downstream modules https://github.com/OPM/opm-autodiff/pull/350 and https://github.com/OPM/opm-polymer/pull/100 should be rebased.

PR #447, opened 2015-04-13
quick fix: set soversion for installed objects

akva2:
installed objects needs to be versioned. this is a quick fix for the release branch, should be fixed by using the shared build system in later releases.


akva2:
it's not a fatal error so i have packed up rc1. no new tag needed.

joakim-hove:
OK - thanks.

akva2:
actually, i'm dumb. it is indeed fatal due to the packaging. i'll have to rebuild. and then there should be a tag.

joakim-hove:
There you go: https://github.com/OPM/opm-parser/releases

PR #454, opened 2015-04-17
Opm 191 TUNING

chflo:
Updated TUNING keyword, and added a Tuning class to hold tuning keyword item values.
Added Tuning class member and handling of Tuning keyword to Schedule class.
Updated DynamicState to support updating of default values in the Tuning class. 
Fixed bug in ParserDoubleItem.cpp; too low precision in doubles in generated code. 
Added testing.


joakim-hove:
Superceeded by:https://github.com/OPM/opm-parser/pull/457

PR #455, opened 2015-04-20
Adding keywords PLYSHLOG.

GitPaean:
I am trying to add the keyword PLYSHLOG.  

It is my first time trying to add a keyword, and this keyword is different. 

Please suggest if anything wrong and better way to do this keyword. 

Thanks.

### DIFF
--- a/opm/parser/share/keywords/000_Eclipse100/P/PLYSHLOG
+++ b/opm/parser/share/keywords/000_Eclipse100/P/PLYSHLOG
@@ -1,7 +1,6 @@
-{ "name" : "PLYSHLOG" , "sections" : ["SPECIAL", "PROPS", "SCHEDULE"] , "records" : [
+{ "name" : "PLYSHLOG" , "size" : 2 , "sections" : ["SPECIAL", "PROPS", "SCHEDULE"] , "records" : [
     [{"name" : "REF_POLYMER_CONCENTRATION" , "value_type" : "DOUBLE"  , "dimension" : "PolymerDensity"},
      {"name" : "REF_SALINITY" , "value_type" : "DOUBLE" , "dimension" : "Salinity"},
      {"name" : "REF_TEMPERATURE" , "value_type" : "DOUBLE" , "dimension" : "Temperature"}],
-    [{"name" : "WATER_VELOCITY" , "value_type" : "DOUBLE"  , "dimension" : "Length/Time"},
-     {"name" : "SHEAR_MULTIPLIER" , "value_type" : "DOUBLE", "dimension" : "1"}]
+    [{"name" : "DATA" , "value_type" : "DOUBLE"  , "size_type" : "ALL" , "dimension" : ["Length/Time" , "1"]}]
 ]}
--- a/opm/parser/share/keywords/000_Eclipse100/P/PLYSHLOG
+++ b/opm/parser/share/keywords/000_Eclipse100/P/PLYSHLOG
@@ -1,0 +1,7 @@
+{ "name" : "PLYSHLOG" , "sections" : ["SPECIAL", "PROPS", "SCHEDULE"], "size" : {"keyword" : "TABDIMS" , "item" : "NTPVT"} , "records" : [
+    [{"name" : "REF_POLYMER_CONCENTRATION" , "value_type" : "DOUBLE"  , "dimension" : "PolymerDensity"},
+     {"name" : "REF_SALINITY" , "value_type" : "DOUBLE" , "dimension" : "Salinity"},
+     {"name" : "REF_TEMPERATURE" , "value_type" : "DOUBLE" , "dimension" : "Temperature"}],
+    [{"name" : "WATER_VELOCITY" , "value_type" : "DOUBLE"  , "dimension" : "Length/Time"},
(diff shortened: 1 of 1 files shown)


joakim-hove:
> and this keyword is different.

Yes  - this keyword is different; maybe so different that the current keyword model breaks? First of all, this is a *multi record keyword* in the sense that not all records in the keyword have the same configuration; this is supported by using:
```json
... records: [ ["List of items in record1"] , ["List of items in record2"], ... ]
```

Observe that it is not necessary to add separate json records for each record in the keyword, the last json record will apply to all the consecutive records in the input deck; so in your case the configuration should be something like:

```json
... records : [
   [{"name" : "REF_CONC" , ...}, ... , {"name" : "REF_TEMP"}],
   [{"name" : "SHARE_RATE"} , {"name" : "SHARE_RATE_MULTIPLIER"}]
]
```

See e.g. https://github.com/OPM/opm-parser/blob/master/opm/parser/share/keywords/000_Eclipse100/W/WELSEGS for an example of a multirecord keyword.

The configuration I have written should work just OK; however the reason I am bit uncertain about this keyword is that it is not 100% clear to me how the multiple tables is supported. The suggestion above, which is the way I interpret the manual (*without* 100% confidence) is that the keyword is just a bunch of data which is slash terminated, and the separation into different tables is handled by calling scope (i.e. that would probably be in `EclipseState` in our case) - and the parser can ignore the `TABDIMS:NTPVT` information; however if the number of tables is fixed and the individual tables are '/' separated we are in trouble - because then the size of the keyword (i.e. the number of records) should be given as arithmetic operation - not supported:-(

The example in the documentation only has `NTPVT == 1` - it would be interesting if you have an example with `NTPVT > 1`.

GitPaean:
I updated a little bit.

I do not know what a example with multiple tables should look like. I will test it when I get a chance later. I do not think we will encounter that kind of example very soon.

I guess they should have the same number of group of the reference data and tables. 


And also, with another keyword SHRATE, the definition and unit of the first column of the table will also be different. I am not sure if will cause problem later.

joakim-hove:
OK; let us hope this is good enough. Two comments:

1. We have implicitly assumed that the whole keyword is '/' terminated, i.e. the `"size" : {...}` should just be removed; if we really need to access that information we are in trouble ...
2. The `"dimension" : 1` is unnecessary - but harmless of course.

GitPaean:
I removed the size information. I keep the ` "dimension" : 1 `. I think it can be more explicit to me that this is a dimensionless value.  Of course, I am very new to the keyword definition at the moment. 

Thanks for all the explanation.

joakim-hove:
OK - I am ready to merge this; but just a last check: You have tried this with a deck?

GitPaean:
Not Yet. I will try now and tell you what happens.

GitPaean:
Hi, I tried with flow-polymer in OPM-polymer with the deck file with PLYSHLOG. 

``` Exception caught: Tried to create a deck keyword from an incomplete raw keyword PLYSHLOG ```

I am not sure if there  is some other way to just test the parser?  

Thanks.

**Comment by joakim-hove** [comment]
> Exception caught: Tried to create a deck keyword from an incomplete raw keyword PLYSHLOG

OK - that was bad news. The crux of the matter is whether the `PLYSHLOG` keyword is '/' terminated. Can you paste it in here?

> I am not sure if there is some other way to just test the parser?

Not a good method, but you can run the `bin/opm-eclkwtest` application which tries to load a deck and create an `EclipseState` instance.

**Comment by GitPaean** [comment]
I will send the data file to you in an Email.

**Comment by joakim-hove** [comment]
OK - thanks; I have looked at it. I had not looked properly at the documentation - this breaks the model :-(

The problem is that this keyword exercises two features which are not supported simultaneously:

1. Multi record configuration.
2. Keyword size (i.e. the number of records) given as the value of *another* kewyord::item pair.

I guess the following should work for now - but this is not a proper solution:

```json
{ "name" : "PLYSHLOG" , "size" : 2 , "sections" : ["SPECIAL", "PROPS", "SCHEDULE"] , "records" : [
    [{"name" : "REF_POLYMER_CONCENTRATION" , "value_type" : "DOUBLE"  , "dimension" : "PolymerDensity"},
     {"name" : "REF_SALINITY" , "value_type" : "DOUBLE" , "dimension" : "Salinity"},
     {"name" : "REF_TEMPERATURE" , "value_type" : "DOUBLE" , "dimension" : "Temperature"}],
    [{"name" : "DATA" , "value_type" : "DOUBLE"  , "size_type" : "ALL" , "dimension" : ["Length/Time" , "1"]}]
]}
```

In this "solution" we fix the number of records to two with the `"size" : 2` in the header.

GitPaean:
Thanks. It works now. 

The requirement of support of TABDIMS for this keyword is not urgent while the support of SHRATE will come pretty soon. This also makes this keyword a little different. 

Thanks for all your help.

joakim-hove:
OK merging now. 

Do not really understand this:

> [...] while the support of SHRATE will come pretty soon. This also makes this keyword a little different.

GitPaean:
With SHRATE, the meaning and unit of the first column of the PLYSHLOG table are different. 

I am not sure if it will cause any problem then. We will discuss that when we are there.

PR #458, opened 2015-04-23
fixed: pkgconfig file

akva2:
multiarch had not been setup and the _TARGET variable is not set
in this build system.

please backport.


joakim-hove:
Thanks; backported to release/2015.04 an tag `relase/2015.04/rc4` created.

PR #459, opened 2015-04-23
Build classes for parser keywords

joakim-hove:
[NB: This breaks the sibling build and should not be merged yet. ]

This PR updates the code generator to create a small class with `static const` members for each keyword. The purpose of this is (currently) twofold:

1. To check if the deck has the `PORO` keyword we can now do: `deck.hasKeyword(ParserKeywords::PORO::keywordName)` instead of `deck.hasKeyword("PORO")` - that way typos++ can be caught at compile time.
2. Get easy access to the default values: `ParserKeywords::EQUIL::DATUM_DEPTH::defaultValue`.

The generated classes derive from `ParserKeyword` - at a later stage it might be interesting to create a default constructor as well.

A gist showing an excerpt of the generated header file is here: https://gist.github.com/joakim-hove/d51dc11b550031a15391 - I would be very grateful if some of the qualified C++ navigators would like to comment on the organisation (i.e. use of namespace / inner classes / ... )


atgeirr:
I looked at the suggested generated header, and I think it looks fine. I do wonder what we would use the itemName fields for, but I guess it means introspection is possible.

In your usage examples you write:
`deck.hasKeyword(ParserKeywords::PORO::keywordName)` instead of `deck.hasKeyword("PORO")`. I think the suggested change is a bit verbose, but it is safer. I propose to add a template variant of hasKeyword taking a ParserKeyword argument to reduce verbosity:
```
template <class Keyword>
bool hasKeyword()
{
    return hasKeyword(Keyword::keywordName)
}
```
The function can be used like this: `deck.hasKeyword<ParserKeywords::PORO>()`.

What do you think?

joakim-hove:
Thank you for looking into this.

>I do wonder what we would use the itemName fields

Well - most of the keywords have named items, e.g. the `EQLDIMS` keyword has items: `["NTEQUL","DEPTH_NODES_P",...,"TRACER_VD"]`. With help of the `itemName` fields we can get compile time checking of these names as well. https://github.com/OPM/opm-parser/blob/master/opm/parser/share/keywords/000_Eclipse100/E/EQLDIMS

> I think the suggested change is a bit verbose

Yes - it certainly is verbose!

>I propose to add a template variant of hasKeyword taking a ParserKeyword argument to reduce verbosity: [...]

That was a fantastic idea; I will do that.

PR #462, opened 2015-04-27
Adding the support for the keyword PLYSHLOG.

GitPaean:
Assuming only one group of PLYSHLOG data is available.

This PR is for suggestions and please do not merge it. 

I think at least I am not supposed to add friend class in `SingleRecordTable.hpp`, but please suggest some better ways. Thanks.


joakim-hove:
Hello Kai; thank you for the PR. I must admit I am not sufficiently familiar with the table code. Maybe @andlaus could take a look?

joakim-hove:
Hello; I have now build tested this code - and it is all green. I would certainly like to remove all these table classes, and instead have one (or a few) classes which can be instantiated by the different table keywords, however that is a significant job.

To me it seems your code is very much a copy & paste of the existing table classes, so based on that I would be happy to merge it in it's current state. Please indicate what you want me to do.

GitPaean:
The thing that bothered me is that a friend class is introduced for `SingleRecordTable.hpp`. If it is not a problem to you, I will be happy if you can merge it.

joakim-hove:
> The thing that bothered me is that a friend class is introduced for SingleRecordTable.hpp.

Well - I see that point; however all the tables are like that. Merging.

PR #464, opened 2015-05-05
Added C++ wrappers for ERT: ecl_kw and fortio

joakim-hove:
Basic wrappers for ert types ecl_kw_type and fortio_type - might be moved to the ert library in the near future?

### DIFF
--- a/opm/parser/eclipse/CMakeLists.txt
+++ b/opm/parser/eclipse/CMakeLists.txt
@@ -10,6 +10,7 @@ if (BUILD_TESTING)
    add_subdirectory(EclipseState/Tables/tests)
    add_subdirectory(EclipseState/Grid/tests)
    add_subdirectory(EclipseState/Util/tests)
+   add_subdirectory(ert/tests)
 endif()
 
 add_subdirectory( Applications )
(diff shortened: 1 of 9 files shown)


andlaus:
this is good stuff, but isn't ERT the natural place for this? (further, is there a technical reason not to  spin-off libecl from ERT?)

joakim-hove:
> [...], but isn't ERT the natural place for this? 

Yes obviously - I desperately needed the functionality when debugging: https://github.com/OPM/opm-parser/pull/440 - but yes - I will move it ERT asap.

> (further, is there a technical reason not to spin-off libecl from ERT?)

Well; it will of course increase the complexity of maintaining the full ERT distribution, but I see your point.

andlaus:
> Well; it will of course increase the complexity of maintaining the full ERT distribution, but I see your point.

I see. The thing is that if you ever want to swap Eclipse by FLOW for the forward simulations in ERT, you might get circular dependencies if libecl is shipped as part of the main ERT package.

joakim-hove:
>  The thing is that if you ever want to swap Eclipse by FLOW for the forward simulations in ERT, 

To use FLOW with the model update workflow in ERT is very high on the agenda; the ability to run an internal test model is one of the main reasons for the #440 effort. The coupling between ERT and the jobs in the the forward model, i.e. ECLIPSE and soon FLOW is so weak (very much process based ...), so circular dependencies is no problem.

andlaus:
> so circular dependencies is no problem.

really? I'd expect some 'find the FLOW binary' cmake code to be in ERT. but obviously you're the expert...

PR #469, opened 2015-05-22
change how tests are added in the build system

akva2:
- eliminate BUILD_TESTING cmake option
- do not build tests by default (i.e. EXCLUDE_FROM_ALL)
- add a 'test-suite' target which builds tests
- add a 'check' target which builds tests, then executes them

additionally it's much simplified through the use of macros. strictly speaking this does two changes in one pull, but splitting it would require lots of dupe work due to all the cut&paste.

sadly cmake does not allow us to extend built-in targets such as 'test' to have it build the tests, then execute them. thus the need for the 'check' target to achieve this. luckily this is how autotools systems is done, and thus this is compatible with packaging software and user expectations.

if this approach is acceptable i'll apply it across all modules. needs https://github.com/OPM/opm-cmake/pull/5


joakim-hove:
That was å simplification! I will merge as soon as the opm-cmake pr is
merged.

Naively I would assume that the updating of the test list should have used
APPEND?
22. mai 2015 16:31 skrev "Arne Morten Kvarving" <notifications@github.com>
følgende:

akva2:
changed to use updated opm-cmake macro.

PR #470, opened 2015-05-24
Opm 199 IOConfig

chflo:
Added IOConfig class. The class holds data about input and output configuration; currently the following is supported:
- If UNIFIN/UNIFOUT/FMTIN/FMTOUT is set
- Should restart file be written at a given timestep (RPTRST keyword is supported)
- Should EGRID file be written (GRIDFILE / NOGGF keyword)
- Should INIT file be written (INIT keyword)
A const pointer to the IOConfig object can be fetched from EclipseState

### DIFF
--- a/opm/parser/eclipse/Applications/Schedule.cpp
+++ b/opm/parser/eclipse/Applications/Schedule.cpp
@@ -16,7 +16,8 @@ int main(int /* argc */, char** argv) {
     std::string file = argv[1];
     Opm::DeckConstPtr deck = parser->parseFile(file);
     std::shared_ptr<const Opm::EclipseGrid> grid = std::make_shared<const Opm::EclipseGrid>( deck );
-    Opm::Schedule sched( grid , deck );
+    Opm::IOConfigPtr ioConfig;
+    Opm::Schedule sched( grid , deck, ioConfig );
 
     std::cout << "Wells: " << sched.numWells() << std::endl;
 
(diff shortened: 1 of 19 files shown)


joakim-hove:
This looks very good - just one last thing. My suggestion is that the  `IOConfig` class gets the ?`input_path` argument as input:

```
IOConfig::IOConfig( deck , std::string& input_path) { 
   ...
}
```

GitPaean:
With this PR, the Cartesian grid we have been using for some while failed. 

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  IOConfig: Reading GRIDFILE keyword from GRID section: Output of GRID file is not supported
Aborted
```

The grid file we are using is as follows for your testing. 
```
RUNSPEC

DIMENS
20 1 5 /

GRID

NEWTRAN

GRIDFILE
  2  /

GRIDUNIT
METRES  /

INIT

--------------------------------------------------------
--
--  	Grid
--
--------------------------------------------------------

DX
100*200 /
DY
100*200 /
DZ
100*25 /

TOPS
20*2500
20*2525
20*2550
20*2575
20*2600 /
```

joakim-hove:
Well - this is because of the `GRIDFILE` keyword. This:
```txt
GRIDFILE
  2  /
```
asks for output of a `.GRID` file in extended format - which has previously been ignored.The `.GRID` format is an old format which we do not support, instead we support the (not very) new `.EGRID` format. Output of an `.EGRID` file is default - so just removing the `GRIDFILE` keyword should solve this problem. If you insist on using the `GRIDFILE` keyword it should be:

```txt
GRIDFILE
  0  1 /
```

GitPaean:
Thanks for pointing out. I will try.

atgeirr:
I suggest that we do this:
 - Reduce the severity of the "output of GRID file not supported" to a warning.
 - Change the input file in question anyway to avoid getting an annoying warning every time.

chflo:
OK, I will change it to a warning

PR #475, opened 2015-05-26
Updates / fixes in tests

joakim-hove:


bska:
I'm seeing build failures in my `-j ${N}` builds of opm-parser in fresh build trees; `-j1` is fine.  `git bisect` suggests that commit d275063 may have inadvertently removed some essential dependency information with respect to the generated source code.  Is this PR in any way related?  Would you like me to test it?

joakim-hove:
Yes - this PR is my attempt to fix those build problems, must admit that I
did not really understand the problem - which always a bad sign.

Cool if you test it.
26. mai 2015 20:33 skrev "Bård Skaflestad" <notifications@github.com>
følgende:

akva2:
see https://github.com/OPM/opm-cmake/pull/11 for core of issue. i'd prefer you change back to use the macro.

bska:
> Cool if you test it.

For what it's worth, this patch survived my nightly "full" run.  In other words, this does at least treat the symptom.

joakim-hove:
OK - this has been updated to use the `opm_add_test()` macro along with: https://github.com/OPM/opm-cmake/pull/11 - looks good.

PR #479, opened 2015-05-28
fixed: copy dummy header in place

akva2:
we do not look for generated headers in the source tree. @joakim-hove for review.


joakim-hove:
OK - thanks; I thought I had covered but .... :-(

PR #484, opened 2015-05-29
Opm 205

chflo:
* Added handling of RPTSCHED RESTART to IOConfig and Schedule
* Reduced severity of "output of GRID file not supported" from throw to warning
* Handle RPTRST Basic = 0

### DIFF
--- a/opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
+++ b/opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
@@ -32,7 +32,8 @@ namespace Opm {
         m_UNIFOUT(false),
         m_FMTIN(false),
         m_FMTOUT(false),
-        m_eclipse_input_path(input_path) {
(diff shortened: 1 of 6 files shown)


opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
```diff
@@ -113,6 +114,19 @@ namespace Opm {
 
 
     void IOConfig::handleRPTRSTBasic(TimeMapConstPtr timemap, size_t timestep, size_t basic, size_t frequency, bool update_default) {
+
+        if (6 == basic )
+        {
+            throw std::runtime_error("OPM does not support the RPTRST BASIC=6 setting (write restart file every timestep)");
+        }
+
+        if (2 < basic) {
```
joakim-hove: `(6 == basic)` is barely ok, but `(2 < basic)` is just to painfull to read.

PR #488, opened 2015-06-03
Added WPIMULT keyword support

iLoop2:


opm/parser/eclipse/EclipseState/Schedule/Completion.hpp
```diff
@@ -63,11 +65,13 @@ namespace Opm {
         int m_i, m_j, m_k;
         Value<double> m_diameter;
         Value<double> m_connectionTransmissibilityFactor;
+        Value<double> m_wellPi;
```
joakim-hove: The `Value<T>` datatype is used for situations where an item has a default value, which should be used - e.g.: "If xxx equals -1 - use the alternative algorithm", or no default at all. For the WellPi the default value of 1.0 is perfectly ok to use  - and there is no reason to wrap it in a `Value<double>` - just use plain `double`.

opm/parser/eclipse/EclipseState/Schedule/Completion.cpp
```diff
@@ -45,7 +46,21 @@ namespace Opm {
...
+            :
+            m_i(oldCompletion->getI()),
+            m_j(oldCompletion->getJ()),
+            m_k(oldCompletion->getK()),
+            m_diameter(oldCompletion->getDiameterAsValueObject()),
+            m_connectionTransmissibilityFactor(oldCompletion->getConnectionTransmissibilityFactorAsValueObject()),
+            m_skinFactor(oldCompletion->getSkinFactorAsValueObject()),
+            m_state(oldCompletion->getState()),
+            m_direction(oldCompletion->getDirection()),
+            m_wellPi("WellPi",wellPi)
```
joakim-hove: The `wellPi` should be cumulative - i.e. we should multiply with the current value.

PR #489, opened 2015-06-05
Updated code generation

joakim-hove:
With this PR the system for generating code; i.e. ParserKeyword instances from Json, is completely refactored. The bulk of the PR is in: 76cced5. The short summary is that separate classes deriving from `ParserKeyword` are created for each keyword - with this we get:

 1. Keyword and item names are available as constants; i.e. we can use:
```
deck->getKeyword(ParserKeywords::ACTNUM::keywordName)
```
or the template form:
```
deck->getKeyword<ParserKeywords::ACTNUM>.
```

2. The item defaults are available as static constants.

3. Creating a keyword a fully configured keyword is quite simple:
```
ParserKeywords::ACTNUM actnumKeyword;
```

In addition the actual generator code has been completely revamped - now it should be readable?

Review comments very welcome.


andlaus:
While I think that this is a very good thing to have, I currently get cmake errors with this:

```
CMake Error at opm/parser/eclipse/CMakeLists.txt:5 (add_subdirectory):
  add_subdirectory given source "Generator/tests" which is not an existing
  directory.
```

Also, I think that it would be beneficial if the generated classes provided convenience methods similar to what the `FooWrapper` classes provide for some keywords. Maybe this PR already provides that, but I couldn't figure it out from a cursory look at it.

joakim-hove:
Thank you for testing - I have pushed an update.

> Also, I think that it would be beneficial if the generated classes provided convenience methods similar to what the FooWrapper classes provide for some keywords. 

That is not part of this PR - I have thought of it; a bit uncertain - it might essentially encourage increased use of `Deck` versus `EclipseState`?

andlaus:
> Thank you for testing - I have pushed an update.

it works now, thanks!

> it might essentially encourage increased use of Deck versus EclipseState?

yeah, in the end is a matter of where you want to draw the line in the sand between the deck and `EclipseState`. As far as I'm concerned, I see such convenience methods as the main benefit of generating a separate class for each keyword because IMO (1) there's no real advantages in integrating simple stuff like `DISGAS` into `EclipseState`, (2) `EclipseState` would become an even bigger monster than it already is (after all, you need to internalize every conceivable Keyword of every exotic feature somebody wants to play around with), (3) not having access to the deck makes it really hard to support custom keywords like the ones of opm-upscaling and (4) with such convenience methods, I consider most keywords to be fully internalized.

that said, I think it is mostly a matter of taste. (and willingness to endure pains. ;)

joakim-hove:
>in the end is a matter of where you want to draw the line in the sand between the deck and EclipseState

Yes - and you have a point. I see the `EclipseState` as essential for properties which are logically composed of several keywords, but it is getting large and unwieldy - and for simple properties autogenerated classes `DeckKeywords::T` might be a good alternative. A little into the future though.

opm/parser/eclipse/Generator/KeywordGenerator.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+    bool KeywordGenerator::updateTest(const KeywordLoader& loader , const std::string& testFile) const {
+        std::stringstream stream;
+
+        stream << testHeader();
+        for (auto iter = loader.keyword_begin(); iter != loader.keyword_end(); ++iter) {
+            const std::string& keywordName = (*iter).first;
+            std::shared_ptr<ParserKeyword> keyword = (*iter).second;
+            stream << startTest(keywordName);
+            stream << "    std::string jsonFile = \"" << loader.getJsonFile( keywordName) << "\";" << std::endl;
```
atgeirr: I think it would be better to use `\n` here rather than `std::endl`. This code is run for each keyword I assume, and actually flushing to file for every line should be unnecessary, and may take significant time (but I have not measured this).

atgeirr:
I have not read through this in detail, and think that I will not take the time to do that. I have tested the PR on my system and did not notice any negative changes (compiling, and a partial SPE9 run).

opm/parser/eclipse/Generator/KeywordGenerator.cpp
joakim-hove: Hmmm  - I have always thought `std::endl` was just the C++ish way to write `\n`? At least for the code in question the `<<` goes to a `std::stringstream` so no flushing should be required.

joakim-hove:
>I have tested the PR on my system and did not notice any negative changes (compiling, and a partial SPE9 run).

Thanks - unless problems arise now I will merge this afternoon.

opm/parser/eclipse/Generator/KeywordGenerator.cpp
atgeirr: No, std::endl is equivalent to '\n' and std::flush.

PR #490, opened 2015-06-05
Added new method getNumPhases to EclipseState

edbru:
This method will be used from opm-core wellstate


joakim-hove:
Try to use 'size_t' and not 'int' for integral size properties.

PR #491, opened 2015-06-08
Add NNC class

totto82:
This class provides the raw non-neighboring connections data as read from the deck and/or added using the `addNNC` method. The NNC data is currently not processed. I.e. multiple NNC connection
between the same cell can exist side by side.


totto82:
@joakim-hove I will soon start adding the NNC class to EclipseGrid. Merging should perhaps wait until then.

opm/parser/eclipse/EclipseState/Grid/NNC.cpp
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+
+    void NNC::addNNC(const int NNC1, const int NNC2, const double trans) {
+        nnc1_.push_back(NNC1);
+        nnc2_.push_back(NNC2);
+        trans_.push_back(trans);
+    }
+
+    int NNC::numNNC() {
```
joakim-hove: To be STL compliant I think this should be: `size_t size()`

opm/parser/eclipse/EclipseState/Grid/NNC.cpp
```diff
@@ -0,0 +1,74 @@
...
+        return(nnc1_.size());
+    }
+
+    bool NNC::hasNNC() {
```
joakim-hove: The `hasNNC_` private member is really not necessary - it is sufficent to check the size.

opm/parser/eclipse/EclipseState/Grid/NNC.cpp
```diff
@@ -0,0 +1,75 @@
...
+namespace Opm
+{
+    NNC::NNC(Opm::DeckConstPtr deck, EclipseGridConstPtr eclipseGrid)
+    {
+        if (deck->hasKeyword("NNC")) {
+            const std::vector<DeckKeywordConstPtr>& nncs = deck->getKeywordList("NNC");
+            for (size_t idx_nnc = 0; idx_nnc<nncs.size(); ++idx_nnc) {
+                Opm::DeckKeywordConstPtr nnc = nncs[idx_nnc];
+                size_t numNNC = nnc->size();
+                nnc1_.reserve(numNNC);
```
joakim-hove: This is *not* performance intensive at all; the `reserve( )` method always makes think twice - it is not worth it here.

opm/parser/eclipse/EclipseState/Grid/NNC.cpp
```diff
@@ -0,0 +1,75 @@
...
+                    xyz2[0] = nnc->getRecord(i)->getItem(3)->getInt(0)-1;
+                    xyz2[1] = nnc->getRecord(i)->getItem(4)->getInt(0)-1;
+                    xyz2[2] = nnc->getRecord(i)->getItem(5)->getInt(0)-1;
+                    size_t global_index2 = eclipseGrid->getGlobalIndex(xyz2[0],xyz2[1],xyz2[2]);
+                    nnc2_.push_back(global_index2);
+
+                    const double trans = nnc->getRecord(i)->getItem(6)->getSIDouble(0);
+
+                    trans_.push_back(trans);
+                    //std::cout << trans << std::endl;
```
joakim-hove: I guess this should be removed.

opm/parser/eclipse/EclipseState/Grid/NNC.cpp
```diff
@@ -0,0 +1,75 @@
...
+                    std::array<int, 3> xyz2;
+                    xyz2[0] = nnc->getRecord(i)->getItem(3)->getInt(0)-1;
+                    xyz2[1] = nnc->getRecord(i)->getItem(4)->getInt(0)-1;
+                    xyz2[2] = nnc->getRecord(i)->getItem(5)->getInt(0)-1;
+                    size_t global_index2 = eclipseGrid->getGlobalIndex(xyz2[0],xyz2[1],xyz2[2]);
+                    nnc2_.push_back(global_index2);
+
+                    const double trans = nnc->getRecord(i)->getItem(6)->getSIDouble(0);
+
+                    trans_.push_back(trans);
```
joakim-hove: As long as the method `addNNC()` is implemented I really think you should use it; i.e. 
```C++
addNNC( global_index1 , global_index2 , trans);
```
instead of individual calls to the `xxx_.push_back()` method.

opm/parser/eclipse/EclipseState/Grid/NNC.cpp
```diff
@@ -0,0 +1,75 @@
...
+            const std::vector<DeckKeywordConstPtr>& nncs = deck->getKeywordList("NNC");
+            for (size_t idx_nnc = 0; idx_nnc<nncs.size(); ++idx_nnc) {
+                Opm::DeckKeywordConstPtr nnc = nncs[idx_nnc];
+                size_t numNNC = nnc->size();
+                nnc1_.reserve(numNNC);
+                nnc2_.reserve(numNNC);
+                trans_.reserve(numNNC);
+
+                for (size_t i = 0; i<numNNC; ++i) {
+                    std::array<int, 3> xyz1;
```
joakim-hove: Maybe `ijk|` instead of `xyz1`?

opm/parser/eclipse/EclipseState/Grid/NNC.hpp
```diff
@@ -0,0 +1,51 @@
...
+    NNC(Opm::DeckConstPtr deck, EclipseGridConstPtr eclipseGrid);
+    void addNNC(const int NNC1, const int NNC2, const double trans);
+    const std::vector<int>& nnc1() const { return nnc1_; }
+    const std::vector<int>& nnc2() const { return nnc2_; }
+    const std::vector<double>& trans() const { return trans_; }
+    size_t numNNC() const;
+    bool hasNNC() const;
+
+private:
+    std::vector<int> nnc1_;
```
joakim-hove: Please - in this part of the woods we use `m_xxx` instead of `xxx_`:smile:

opm/parser/eclipse/EclipseState/Grid/NNC.cpp
bska: On the other hand, calling `.reserve()` makes it eminently clear to the reader that the author expected a general structure of
```C++
for (each element e) {
    stuff;
    x.push_back(e);
}
```
and that the `push_back()` call was not inserted by accident.

opm/parser/eclipse/EclipseState/Grid/NNC.cpp
```diff
@@ -0,0 +1,75 @@
...
+                    xyz1[0] = nnc->getRecord(i)->getItem(0)->getInt(0)-1;
```
joakim-hove: I think you should convert to size_t here: `static_cast<size_t>( nnc->getRecord() ... )` and then use `size_t` throughout. Right now it is a mix back and forth between `size_t` and `int`.

joakim-hove:
Sorry for the late review - I had not seen that you updated the PR. Anyway - apart from the very small details mentioned I am ready to merge this.

opm/parser/eclipse/EclipseState/Grid/NNC.cpp
atgeirr: I agree with @bska here: best practice when you know the size is either reserve()/push_back() or resize()/operator[]().

bska: That said, in this case, the `.reserve()` call is wrong.  The statement should be
```C++
Container.reserve(Container.size() + numNNC);
```
to account for `.reserve()` begin called in a loop.

atgeirr: So the case is actually that we do *not* know the size beforehand? Then I can agree with @joakim-hove that it is best to remove it: repeated reserve() in a loop is not a performance improvement, and it only confuses in that case (as I was).

joakim-hove: > and it only confuses

Hehe - I rest my case :smile:

bska: > So the case is actually that we do not know the size beforehand?

Well, yeah, kind of.  The total number of `NNC`s is the sum of the `->size()` of each `NNC` keyword.  It is entirely trivial to precompute that number in a separate loop, e.g.,
```C++
std::vector<double>::size_type numNNC = 0;
for (auto& b = nncs.begin(), e = nncs.end(); b != e; ++b) {
    numNNC += b->size();
}
```
and then use `numNNC` in the `.reserve()` call.

atgeirr: Yes I agree it could be done like that, but the benefit is not so large (so much less code to do that sum in other languages...). My main point is that there should only ever be a single reserve() or resize(), doing either in a loop is bound to confuse readers.

bska: > My main point is that there should only ever be a single `reserve()` or `resize()`

Sure.  I would have thought it obvious that precomputing `numNNC` would be accompanied by hoisting the `.reserve()` calls out of the loop.

totto82:
Thanks for the comments. They are adressed in the newest commit.

totto82:
Last int changed to size_t

PR #492, opened 2015-06-08
Added method globalReset() to DynamicState<T>

joakim-hove:
This small PR adds the ability to reset the content of DynamicState<T> property - this is support full global override of properties which can be specified both in the deck and on the commandline.


atgeirr:
Sounds good, I support this!

PR #493, opened 2015-06-09
Opm 211

edbru:
Added new class InitConfig to handle RESTART and SKIPREST keywords from the deck. 
EclipseState holds an instance of the InitConfig class. 

This PR is supporting the task of implementing initialzation from a RESTART file in Flow

### DIFF
--- a/opm/parser/eclipse/EclipseState/InitConfig/InitConfig.cpp
+++ b/opm/parser/eclipse/EclipseState/InitConfig/InitConfig.cpp
@@ -1,0 +1,69 @@
+/*
+  Copyright 2015 Statoil ASA.
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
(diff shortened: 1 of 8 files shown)


opm/parser/eclipse/EclipseState/InitConfig/InitConfig.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+        initRestartKW(deck);
+    }
+
+    void InitConfig::initRestartKW(DeckConstPtr deck) {
+        if (deck->hasKeyword("RESTART") && deck->hasKeyword("SKIPREST")){
+            DeckKeywordConstPtr restart_kw = deck->getKeyword("RESTART");
+            DeckRecordConstPtr restart_dataRecord = restart_kw->getRecord(0);
+            DeckItemPtr restart_rootname_item = restart_dataRecord->getItem(0);
+            const std::string restart_rootname_string = restart_rootname_item->getString(0);
```
joakim-hove: No need to take a copy: `const std::string&`

opm/parser/eclipse/EclipseState/InitConfig/InitConfig.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+            DeckItemPtr restart_report_step_item = restart_dataRecord->getItem(1);
+            int restart_report_step_int = restart_report_step_item->getInt(0);
+
+            DeckItemPtr save_item = restart_dataRecord->getItem(2);
+            if ((save_item->size() > 0) && (save_item->getString(0) == "SAVE")) {
```
joakim-hove: This test is related to the content of the `save_item`; better to just check if it has been assigned:
```
if (save_item->hasValue(0)) {
    throw ...
}
```

opm/parser/eclipse/EclipseState/InitConfig/InitConfig.cpp
joakim-hove: If *only* one of `RESTART` or `SKIPREST` is found we should throw. We *really* should get the `OpmLog` system in proper use - then it will be natural to issue a warning there.

joakim-hove:
Will merge when these small items have been adressed.

PR #496, opened 2015-06-12
Relaxed parse

joakim-hove:
Have introduced a flag bool flag for relaxed parsing. The main purpose of this is that one can create a custom parser with only some keywords, this parser will then ignore the unknown keywords gracefully.

The flag is defaulted to `false` - i.e. no change in behavior, see: bceb824 for more details and an example.


atgeirr:
This sounds like a very usable feature, and I support making it default to strict behaviour. (I have not reviewed the code.)

PR #499, opened 2015-06-18
Added information on Json for the uninformed.

alfbr:
Thanks for the improved documentation! Just a minor inclusion from me, I guess there are more than me who are familiar with XML but not with JSON, so a brief pointer might help.

### DIFF
--- a/docs/keywords.txt
+++ b/docs/keywords.txt
@@ -4,10 +4,13 @@ parser.
 1. The structure of a keyword
 -----------------------------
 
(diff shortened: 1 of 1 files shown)


alfbr:
Typo fixed, do you want me to expand on the JSON explanation?

joakim-hove:
> Feel free to elaborate if you want a more precise and detailed description of what JSON is.

My point was that I did not really want to use bits and bytes on explaining what Json is - it is well described on the http://www.json.org website - anyway I'll merge.

Issue #501, opened 2015-06-22
Problems in IOConfig and KeywordGenerator


atgeirr:
When I compile the latest master I get the following warnings:
```
/Users/atgeirr/opm-build-debug/opm-parser/opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp:268:5: warning: control reaches end of non-void
      function [-Wreturn-type]
    }
    ^
/Users/atgeirr/opm-build-debug/opm-parser/opm/parser/eclipse/Generator/KeywordGenerator.cpp:105:26: warning: declaration shadows a local variable
      [-Wshadow]
            std::fstream stream(filename.c_str() , std::fstream::out);
                         ^
/Users/atgeirr/opm-build-debug/opm-parser/opm/parser/eclipse/Generator/KeywordGenerator.cpp:93:23: note: previous declaration is here
        std::ifstream stream(filename.c_str());
```
The IOConfig warning is serious: failing to return will usually give problems, in this case the function returns a bool, so there is no way for downstream code to tell it is invalid. I *cannot* fix this function since I do not know what it is intended to return.

The KeywordGenerator warning is probably less serious, but it should be taken care of: the variable name `stream` is used both for an input and an output stream, which is quite obfuscating. Other minor issues with that function include using non-const reference arguments and manually close()-ing a stream instead of letting the destructor do its job. I'll submit a fix for this function.

chflo:
I have created a PR https://github.com/OPM/opm-parser/pull/503 for the IOConfig compiler warning, sorry for this mistake, the method overrideRestartWriteInterval should have return type void, not bool

atgeirr:
No more warnings, great! Closing.

PR #502, opened 2015-06-22
Improve updateFile and suppress warnings.

atgeirr:
This fixes one of the problems described in #501.

A few notes:
 - Now that we can assume quite new compiler versions we can finally use the std::string constructor of the file stream classes!
 - I used scoping to ensure the input stream is closed before the output stream opens. For all I know it was perfectly legal the way it was used, but it certainly opens the door to extremely thorny issues of how they would interact so it is better to keep them apart.
 - I could not see any tests for this, if there are any, where should I check?

### DIFF
--- a/opm/parser/eclipse/Generator/KeywordGenerator.cpp
+++ b/opm/parser/eclipse/Generator/KeywordGenerator.cpp
@@ -89,22 +89,24 @@ namespace Opm {
             boost::filesystem::create_directories( file.parent_path());
     }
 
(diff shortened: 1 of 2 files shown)


joakim-hove:
> I could not see any tests for this, if there are any, where should I check?

When the `KeywordGenerator` is run it will create the test `InlineKeywordTest` which verifies that a keyword generated from Json and the default constructor of the same keyword gives the same result. The command `ctest -R InlineKeywordTest` should run this generated test.

joakim-hove:
> so it is better to keep them apart.

I certainly agree - must have been quite low on caffeine when the original code was written.

PR #504, opened 2015-06-23
Bugfixes VFPPROD and test

babrodtk:
Minor changes to the test for VFP tables to test THP_VALUES and WFR_VALUES

Also changes the default of ALQ_DEF from THP (which I believe does not make sense for this item) to UNDEF.

### DIFF
--- a/opm/parser/eclipse/IntegrationTests/ParseVFPPROD.cpp
+++ b/opm/parser/eclipse/IntegrationTests/ParseVFPPROD.cpp
@@ -88,21 +88,21 @@ BOOST_AUTO_TEST_CASE( parse_VFPPROD_OK ) {
     }
 
     {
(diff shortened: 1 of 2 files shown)


joakim-hove:
OK; the default value of `ALQ_DEF` was obviously wrong - however instead of introducing a unusable value `UNDEF` I think you should just remove the default from this item altogether.

If at all possible the default value should be directly usable by the simulator, if you need to check if a deck item has a valid value: 
```C++
if (item->hasValue( 0 ))
   // A value has been provided - either from the deck or from the default in configuration
else
   // No value provided - deal with it!
```

babrodtk:
Removed the undef default as suggested

joakim-hove:
Merging.

PR #506, opened 2015-06-23
Schedule events

joakim-hove:
[ This is mainly put up as a request for comments - not yet merge ready. ]

With this PR we have created an initial system to record events from the Schedule file. The purpose of this is that you can query the events table for specific events, and then decide whether a potentially expensive calculation is required.

This is absolutely work in progress - what I would very much like feedback on are:

 1. Which events are required - the current list in Events.hpp is somewhat random; and I will *not* guarantee that they are set correctly.
 2. Querying for events is based on implementing `Events::operator()` - makes sense?

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Events.hpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Events.hpp
@@ -1,0 +1,95 @@
+/*
+  Copyright 2015 Statoil ASA.
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#ifndef SCHEDULE_EVENTS_HPP
+#define SCHEDULE_EVENTS_HPP
+
+#include <cstdint>
+#include <vector>
+
+namespace Opm
+{
+    namespace ScheduleEvents {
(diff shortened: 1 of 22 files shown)


opm/parser/eclipse/EclipseState/Schedule/Events.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include <cstdint>
+#include <vector>
+
+namespace Opm
+{
+    namespace ScheduleEvents {
+        // These values are used as bitmask - 2^n structure is essential.
+        enum Events {
+            NEW_WELL = 1,
```
atgeirr: Is there a reason for skipping 2 in the sequence of values?

opm/parser/eclipse/EclipseState/Schedule/Events.hpp
```diff
@@ -0,0 +1,65 @@
...
+           ...
+
+      Querying events is done with operator().
+    */
+
+    class Events {
+    public:
+        Events();
+        void addEvent(ScheduleEvents::Events event, size_t reportStep);
+        bool operator()(ScheduleEvents::Events event, size_t reportStep) const;
```
atgeirr: While somewhat cool, I think that operator() is more likely to confuse users than to inform them. I think a more (boring) traditional hasEvent() or stepHasEvent() method name would be better.

atgeirr:
This looks good, and quite easy to use from client code. I have not by any means checked that all events that should be added are actually added, but the parts I looked at seemed ok.

opm/parser/eclipse/EclipseState/Schedule/Events.hpp
joakim-hove: Thanks for looking; I also thought the `operator()` was kind of cool - but of course coolness is not the metric our code should be measured by. Will update.

joakim-hove:
>I have not by any means checked that all events that should be added are actually added

There are two aspects to this:

1. Enumerate the list of events we are interested in.
2. Actually record the events we are interested in.

There is a trade off here where fine grained events can lead to fine grained decisions in the downstream simulator code - which is good; and also potentially increase the complexity in the event recording code - which is bad. So we should not introduce fine grained events for the sake of coolness - but rather limit ourselves to the events which we assume will differentiate on in the downstream code. We can discuss this further.

joakim-hove:
Fixup using `hasEvent()` instead of `operator()` pushed.

PR #511, opened 2015-07-08
add grid properties required for the simple "vertical" endpoint scaling

andlaus:
i.e., the PC[GW] and KR[OGW] properties plus their imbibition and
directional variants. the properties required for three-point vertical
scaling (i.e., KRWR, et al.) are still to be done.

currently, vertical endpoint scaling is not supported by the satuation functions of opm-core, but it is wise for the parser to not stand in the way.

### DIFF
--- a/opm/parser/eclipse/EclipseState/EclipseState.cpp
+++ b/opm/parser/eclipse/EclipseState/EclipseState.cpp
@@ -786,10 +786,18 @@ namespace Opm {
         const auto IPCGLookup = std::make_shared<IPCGEndpointInitializer<>>(*deck, *this);
         const auto KRWLookup = std::make_shared<KRWEndpointInitializer<>>(*deck, *this);
         const auto IKRWLookup = std::make_shared<IKRWEndpointInitializer<>>(*deck, *this);
-        const auto KROLookup = std::make_shared<KRWEndpointInitializer<>>(*deck, *this);
-        const auto IKROLookup = std::make_shared<IKRWEndpointInitializer<>>(*deck, *this);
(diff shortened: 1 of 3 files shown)


andlaus:
ok, I now also added support for vertical three point relperm scaling and made the `EclipseState::has*GridProperty()` methods more strict. (because _having_ a keyword in a deck is a different concept than _supporting_ it in the parser. Basically it makes zero sense to query for unsupported grid properties, so that this kind of error should be brought to the developer's attention as quickly as possible.)

With this series, `flow` still works as before for SPE1, SPE9 and Norne and the ctests pass. I haven't tested more.

joakim-hove:
I get dizzy from all these tables ;-)

Will merge when you have added a test for the new GridProperty behavior.

andlaus:
> I get dizzy from all these tables ;-)

join the club ;) I'm not sure if there are no copy-and-pastos but at least everything which is exercised by Norne does not lead to a crash.

That whole mess would be a lot simpler if the "directional" properties were left unsupported. (they are not used in opm-core and IMO their use-case is pretty exotic.)

> Will merge when you have added a test for the new GridProperty behavior.

Ok, I'll do that tomorrow.

alfbr:
They are less exotic than you think ;-) please leave the support in there.

andlaus:
I haven't suggested removing it. (the parser is probably the easy part anyway.) I was merely lamenting that trying to support stuff which suffers of the combinatorial explosion problem tends to be messy and a pain in the a** in terms of developing and maintaining it ;)

andlaus:
ok, the unit test has been updated in the latest version. (cf  	ba96ad5)

PR #507, opened 2015-06-24
Individual support for parsing of VFPPROD tables

babrodtk:
This runs on my setup, and places the data in the correct variables inthe VFPProdTable class. It should do nothing if there are no VFPPROD keywords in a deck, and only adds functionality.

What is missing here, is the support for units, which will come in a subsequent PR.
Edited: Units should now be supported

PR #512, opened 2015-07-08
Fixed bug in test which checked for the wrong type of exception

babrodtk:
See pull request #507


joakim-hove:
Thanks for the fix - my bad.

joakim-hove:
Thanks for the fix - my bad.

PR #515, opened 2015-07-10
Add keywords for solvent model

totto82:
This PR adds the necessary keywords for the solvent model


joakim-hove:
Sorry - about the slow reply. I think you have used the keyword `SPECIAL` incorrect - as I understand the manual the `SPECIAL` flag in the keyword documentation indicates that a keyword is only available as part of separately licensed Eclipse extension, there is no `SPECIAL` section?!

totto82:
I noticed that the SPECIAL keyword was specified for the some of the polymer keywords, (See for instance PLYROCK) and did the same for the solvent keywords.

joakim-hove:
>I noticed that the SPECIAL keyword was specified for the some of the polymer keywords

Yes - I know; it should be removed from the polymer keywords as well.

totto82:
SPECIAL is removed from this PR

joakim-hove:
Merging

PR #517, opened 2015-07-13
Avoid using extremely long initialiser list.

atgeirr:
The long list caused extreme compile-times for clang in optimised mode. This cuts compilation time for EclipseState.cpp down to 11 seconds, down from almost 3 minutes. The recent addition of many more elements to the list made compile time explode. Probably one or more of the optimisation passes uses a bad (n^2) algorithm or uses so much memory that my 8 GB system started swapping.

In any case, this fixes my problem, and I'd say that the code is a little bit less mysterious when lines are terminated by semicolons not colons...

It is hard to review the changes here in the diff unfortunately, so this is a description:
1) Create an empty supportedDoubleKeywords instead of an initialised one.
2) Add all members with
```
        supportedDoubleKeywords->emplace_back( "SGL"    , SGLLookup, "1" );
```
instead of
```
        SupportedDoubleKeywordInfo( "SGL"    , SGLLookup, "1" ),
```
3) Reindent (avoiding this did not make the diff easier to read).

### DIFF
--- a/opm/parser/eclipse/EclipseState/EclipseState.cpp
+++ b/opm/parser/eclipse/EclipseState/EclipseState.cpp
@@ -875,330 +875,330 @@ namespace Opm {
         // Note that the variants of grid keywords for radial grids
         // are not supported. (and hopefully never will be)
         typedef GridProperties<double>::SupportedKeywordInfo SupportedDoubleKeywordInfo;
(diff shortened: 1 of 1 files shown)


andlaus:
just for the record (as the one who committed the felony): no objections/issues from my side, please merge.

PR #519, opened 2015-07-15
Fix warnings

atgeirr:
This silences a few warnings: signed/unsigned comparisons and member initialisation order (should be same as declaration order).


joakim-hove:
Thank you for persistent warning effort - merging.

PR #521, opened 2015-07-17
SatfuncPropertyInitializer: determine the critical saturations correctly

### LINKED ISSUE #520: Latent issue in EndpointInitializer::findVerticalPoints()
Lines 197--212 of [`SatfuncPropertyInitializers.hpp`](https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Grid/SatfuncPropertyInitializers.hpp#L197) (method `findVerticalPoints()` of class `EndpointInitializer`) contains the following code
```C++
for (size_t rowIdx = 0; rowIdx < krwCol.size(); ++rowIdx) {
    if (krwCol[rowIdx] == 0.0) {
        m_krorw[tableIdx] = krowCol[rowIdx - 1];
        break;
    }
}

// find the oil relperm which corresponds to the critical gas saturation
const auto &krgCol = sgofTables[tableIdx].getKrgColumn();
const auto &krogCol = sgofTables[tableIdx].getKrogColumn();
for (size_t rowIdx = 0; rowIdx < krgCol.size(); ++rowIdx) {
    if (krgCol[rowIdx] == 0.0) {
        m_krorg[tableIdx] = krogCol[rowIdx - 1];
        break;
    }
}
```
The expressions `rowIdx - 1` produce out-of-bounds access (`(size_t) -1`) when `rowIdx == 0` (i.e., on the first iteration).  I don't know what the correct fix is in this case.

andlaus:
At least I hope it's correct: the condition for breaking the loops
where inverted in some places.

thanks to [at]bska for discovering this.

this PR should fix #520, but it would be good if @bska could confirm this

### DIFF
--- a/opm/parser/eclipse/EclipseState/Grid/SatfuncPropertyInitializers.hpp
+++ b/opm/parser/eclipse/EclipseState/Grid/SatfuncPropertyInitializers.hpp
@@ -145,10 +145,10 @@ protected:
             // find the critical oil saturation of the oil-gas system
             numRows = sgofTables[tableIdx].numRows();
             const auto &kroOGCol = sgofTables[tableIdx].getKrogColumn();
-            for (int rowIdx = 0; rowIdx < numRows; ++rowIdx) {
-                if (kroOGCol[rowIdx] == 0.0) {
-                    double Sg = sgofTables[tableIdx].getSgColumn()[rowIdx];
-                    m_criticalOilOGSat[tableIdx] = 1 - Sg - m_minWaterSat[tableIdx];
+            for (int rowIdx = numRows - 1; rowIdx >= 0; --rowIdx) {
+                if (kroOGCol[rowIdx] > 0.0) {
+                    double Sg = sgofTables[tableIdx].getSgColumn()[rowIdx + 1];
+                    m_criticalOilOGSat[tableIdx] = 1 - Sg;
                     break;
                 }
             }
@@ -156,10 +156,10 @@ protected:
             // find the critical oil saturation of the water-oil system
             numRows = swofTables[tableIdx].numRows();
             const auto &kroOWCol = swofTables[tableIdx].getKrowColumn();
-            for (int rowIdx = 0; rowIdx < numRows; ++rowIdx) {
-                if (kroOWCol[rowIdx] == 0.0) {
-                    double Sw = swofTables[tableIdx].getSwColumn()[rowIdx];
-                    m_criticalOilOWSat[tableIdx] = 1 - Sw - m_minGasSat[tableIdx];
+            for (int rowIdx = numRows - 1; rowIdx >= 0; --rowIdx) {
+                if (kroOWCol[rowIdx] > 0.0) {
+                    double Sw = swofTables[tableIdx].getSwColumn()[rowIdx + 1];
+                    m_criticalOilOWSat[tableIdx] = 1 - Sw;
                     break;
                 }
             }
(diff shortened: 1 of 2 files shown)


opm/parser/eclipse/EclipseState/Grid/SatfuncPropertyInitializers.hpp
```diff
@@ -122,7 +122,7 @@ class EndpointInitializer
             for (int rowIdx = 0; rowIdx < numRows; ++rowIdx) {
                 if (krwCol[rowIdx] > 0.0) {
                     double Sw = 0.0;
-                    if (rowIdx > 0)
+                    if (rowIdx > 0.0)
```
bska: This looks strange.  Isn't `rowIdx` an integer?

andlaus: ehem, yes it is. fixed & thanks. (for some reason I didn't see a compiler warning, though.)

joakim-hove:
@bska: This is (unfortunately) quite outside my comfort zone - would be grateful if you could give this a thumbs up or down.

bska:
> would be grateful if you could give this a thumbs up or down.

I will take a closer look tonight.

bska:
> would be grateful if you could give this a thumbs up or down.

I'll give it a horizontal thumb.  The changes to `findVerticalPoints()`--i.e., looking for first non-zero water (gas) relative permeabilities--are good and should go in as is.  The changes to `findCriticalPoints()` are incomplete.  Reverse traversal of kr_o is a good idea, but there can be multiple zero kr_o values at the end of the table, so we need to account for that possibility.  In other words, the loop
```C++
// find the critical oil saturation of the oil-gas system
for (int rowIdx = numRows - 1; rowIdx >= 0; --rowIdx) {
    if (kroOGCol[rowIdx] == 0.0) {
        double Sg = sgofTables[tableIdx].getSgColumn()[rowIdx];
        m_criticalOilOGSat[tableIdx] = 1 - Sg;
        break;
    }
}
```
should rather be written as
```C++
// find the critical oil saturation of the oil-gas system
for (int rowIdx = numRows - 1; rowIdx >= 0; --rowIdx) {
    if (kroOGCol[rowIdx] > 0.0) {
        double Sg = sgofTables[tableIdx].getSgColumn()[rowIdx + 1];
        m_criticalOilOGSat[tableIdx] = 1 - Sg;
        break;
    }
}
```
I.e., the branch criterion should be `> 0` and the critical oil saturation should be extracted from the first row below the last non-zero kr_o value.

Similar considerations hold for the oil-water system (keyword `SWOF`).

By the way, this (#521) produces `make test` failure in [`runGridPropertyTests::GridPropertyInitialization`](https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Grid/tests/GridPropertyTests.cpp#L253)--at least when using libstdc++'s debug mode--but that is due to faulty input rather than erroneous code.  The `SWOF` and `SGOF` tables of that test **must** all define a critical water saturation and a critical gas saturation, meaning each table must have a record for which `kr_w(S_w)=0` (in `SWOF`) or `kr_g(S_g)=0` (in `SGOF`).  Moreover, each table **must** define a critical oil saturation--i.e. each table must have at least one record for which `kr_o(S_o)=0` (with `S_o = 1 - S_w` for `SWOF` or `S_o = 1 - S_g - Swco` for `SGOF`).

andlaus:
ok, both issues fixed. hopefully.

bska:
> both issues fixed.

Confirmed.  This PR is ready for merging.

joakim-hove:
Thank you for both patch and review - merging!

PR #524, opened 2015-07-23
implement support for fortran-style 'd' exponent indicators

### LINKED ISSUE #514: Using 'D' and 'E' when writing numbers on standard form
I don't know if this is very important, but I thought I should mention the following:
If you want to use scientific notation (i.e. writing numbers on standard form) in Eclipse, you can use the letter E (e.g. 0.000003 -> 3E-6). You can also use D (e.g. 3D-6), but this does not seem to be supported by Flow:
"Exception caught: Unable to parse string3.0D-6 to typeid: d"

andlaus:
also, extend the corresponding unit test.

this should fix #514

### DIFF
--- a/opm/parser/eclipse/RawDeck/StarToken.hpp
+++ b/opm/parser/eclipse/RawDeck/StarToken.hpp
@@ -36,15 +36,39 @@ namespace Opm {
             return boost::lexical_cast<T>(valueString);
         }
         catch (boost::bad_lexical_cast&) {
-            throw std::invalid_argument("Unable to parse string" + valueString + " to typeid: " + typeid(T).name());
+            throw std::invalid_argument("Unable to convert string '" + valueString + "' to typeid: " + typeid(T).name());
         }
+    }
+
+    template <>
+    inline double readValueToken<double>(const std::string& valueString) {
+        // Eclipse supports Fortran syntax for specifying exponents of floating point
+        // numbers ('D' and 'E', e.g., 1.234d5) while C++ only supports the 'e' (e.g.,
+        // 1.234e5). the quick fix is to replace 'D' by 'E' and 'd' by 'e' before trying
+        // to convert the string into a floating point value. This may not be the most
+        // performant thing to do, but it is probably fast enough.
+        std::string vs(valueString);
+        std::replace(vs.begin(), vs.end(), 'D', 'E');
+        std::replace(vs.begin(), vs.end(), 'd', 'e');
+
+        try {
+            return boost::lexical_cast<double>(vs);
+        }
+        catch (boost::bad_lexical_cast&) {
+            throw std::invalid_argument("Unable to convert string '" + valueString + "' to double");
+        }
+    }
+
+    template <>
+    inline float readValueToken<float>(const std::string& valueString) {
+        return readValueToken<double>(valueString);
     }
 
     template <>
     inline std::string readValueToken<std::string>(const std::string& valueString) {
         if (valueString.size() > 0 && valueString[0] == '\'') {
             if (valueString.size() < 2 || valueString[valueString.size() - 1] != '\'')
-                throw std::invalid_argument("Unable to parse string" + valueString + " as a string token");
+                throw std::invalid_argument("Unable to parse string '" + valueString + "' as a string token");
             return valueString.substr(1, valueString.size() - 2);
         }
         else
--- a/opm/parser/eclipse/RawDeck/StarToken.hpp
+++ b/opm/parser/eclipse/RawDeck/StarToken.hpp
@@ -42,20 +42,23 @@ namespace Opm {
 
     template <>
     inline double readValueToken<double>(const std::string& valueString) {
-        // Eclipse supports Fortran syntax for specifying exponents of floating point
-        // numbers ('D' and 'E', e.g., 1.234d5) while C++ only supports the 'e' (e.g.,
-        // 1.234e5). the quick fix is to replace 'D' by 'E' and 'd' by 'e' before trying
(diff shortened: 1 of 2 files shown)


joakim-hove:
Thank you - good testing.

opm/parser/eclipse/RawDeck/StarToken.hpp
```diff
@@ -36,15 +36,39 @@ namespace Opm {
...
     }
 
     template <>
+    inline double readValueToken<double>(const std::string& valueString) {
+        // Eclipse supports Fortran syntax for specifying exponents of floating point
+        // numbers ('D' and 'E', e.g., 1.234d5) while C++ only supports the 'e' (e.g.,
+        // 1.234e5). the quick fix is to replace 'D' by 'E' and 'd' by 'e' before trying
+        // to convert the string into a floating point value. This may not be the most
+        // performant thing to do, but it is probably fast enough.
+        std::string vs(valueString);
```
joakim-hove: I would prefer a nested approach; as I think this will only rarely be required:
```
try {
    return boost::lexical_cast<double>(vs);
} catch () {
     std::string vs(valueString);
     std::replace( ... )
     try {
         return boost::lexical_cast<double>(vs);
     }
     catch() { }
}

File: opm/parser/eclipse/RawDeck/StarToken.hpp [hunk 2]
```diff
@@ -36,15 +36,39 @@ namespace Opm {
...
+        std::replace(vs.begin(), vs.end(), 'D', 'E');
+        std::replace(vs.begin(), vs.end(), 'd', 'e');
+
+        try {
+            return boost::lexical_cast<double>(vs);
```
joakim-hove: Any idea of the performance of the current approach based on `try / catch` and `boost::lexical_cast<double>` would compare with an implementation based on `strtod()`?

joakim-hove:
The reason I am picking a bit about performance is that I am hoping Resinsight will consider using the parser, in that case Resinsight startup time will be affected by the parsing time; and in that case parsing of large floating point ASCII input fields like `COORD`, `ZCORN` and `PERM?/PORO` is highly relevant.

opm/parser/eclipse/RawDeck/StarToken.hpp
andlaus: I haven't measured, but parsing the Norne deck in debug mode takes about 5 seconds on my machine. I suppose that less than 5% of that time are spend in `boost::lexical_cast`, so I think `strtod` would only save a few milliseconds. other optimizations -- like reducing the use of `std::shared_ptr` -- would probably provide a drastically better return of investment. (This is only my gut-feeling, though...)

opm/parser/eclipse/RawDeck/StarToken.hpp
andlaus: I'll measure how large the difference is for the Norne deck tomorrow. if the signal/noise ratio is insignificant, I'd prefer to keep is as proposed in the patch. (because it is less complex and hence less can go wrong.)

andlaus:
> The reason I am picking a bit about performance is that I am hoping Resinsight will consider using the parser, in that case Resinsight startup time will be affected by the parsing time; and in that case parsing of large floating point ASCII input fields like COORD, ZCORN and PERM?/PORO is highly relevant.

I suppose the main performance issue for keywords with `sizeType==DATA` items  is the creation of very long strings in `RawRecord` before parsing.

opm/parser/eclipse/RawDeck/StarToken.hpp
joakim-hove: OK - fair enough; the truth of the matter is of course that I have shockingly limited knowledge of where the parser spends it's time.

andlaus: > OK - fair enough; the truth of the matter is of course that I have shockingly limited knowledge of where the parser spends it's time. 

for keywords with DATA items, I'm reasonably confident the following is the place to point the finger to: https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/RawDeck/RawKeyword.cpp#L78 . This is because for records with several hundreds of thousands of values (ZCORN?), `m_partialRecordString` can become several megabytes large. If you consider the fact that the data occasionally needs to be copied for `operator+=`, its actually surprisingly fast! 

I'm not sure how easily this can be speed up, though.

opm/parser/eclipse/RawDeck/StarToken.hpp
joakim-hove: > I'll measure how large the difference is for the Norne deck tomorrow.

Thank you.

>if the signal/noise ratio is insignificant, I'd prefer to keep is as proposed in the patch. 

Sure - the fact that your approach is cleaner and simpler is not debatable.

andlaus:
ok, I did the performance tests for Norne (using Clang in optimized mode) and there were actually some surprises (for me):

- the conversion from string to double takes slightly less than 20% on the current master
- with the unconditional 'd' -> 'e' replacement, the full parsing is about 7%-9% slower, i.e., the string->double conversion is 40 to 50% slower
- using `strtod()` instead of `boost::lexical_cast` does not change performance above the measurement noise

I've thus adapted the PR to use your "only replace 'd' by 'e' if conversion failed, then try again" approach. With this, performance for Norne is exactly where it was before on my machine.

joakim-hove:
> I've thus adapted the PR to use your "only replace 'd' by 'e' if conversion failed, then try again" approach. With this, performance for Norne is exactly where it was before on my machine.

Thanks a lot - merging.

PR #525, opened 2015-07-23
Debug challenge

joakim-hove:
The testing by @larsblatny has brought 'some' issue with the TOPS keyword; when all nx*ny*nz elements in the keyword have been specified in the input deck the simulation results are completely bogus - my guess is that this is memory corruption bug of some sort - but I have not been able to smoke it out with valgrind.

The current PR describes the problem in some detail, and contains a pure symptom fix in the handling of the TOPS keyword.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
+++ b/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
@@ -413,6 +413,39 @@ namespace Opm {
     }
 
 
+    /*
+      There is 'something' rotten which is triggered by through this
+      method. The TOPS keyword can be of any size in the range
+      [nx*ny,nx*ny*nz]. If the keyword has less than nx*ny*nz elements
+      the top of the cells is inferred by adding together the top and
+      thickness of the cell immediately above.
+
+      For some strange reason the code does not work when the TOPS
+      keyword has been fully specified in the input deck. The 'not
+      works' is quite mysterious:
+
+        o The TOPS vector itself is numerically correct.
+
+        o The ert Eclipse grid created with ecl_grid_alloc_dx_dy_dz_tops() in
+          EclipseGrid::initDTOPSGrid() seems to be perfectly right.
+
+        o The COORD, ZCORN and ACTNUM exported from the ert Grid and
+          used as input to the UnstructuredGrid constructor in opm-core
+          is correct.
+
+      The final UnstructuredGrid instance created in
+      GridManager::initFromEclipseGrid() is bogus - presumably some
+      memory is overwritten somewhere - for simple diff based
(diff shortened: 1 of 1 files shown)


alfbr:
Please do not throw on valid decks. Without having checked (curently on holiday), I am wondering whether  TOPS for everything but the top layer is redundant information, and hence simply can be ignored? In the meantime we should track down the bug of course.

joakim-hove:
> Please do not throw on valid decks.

Well - right now we can stop dead with a reasonably good error message, or continue simulating with *guaranteed wrong results* - I find it quite amusing that you seem to prefer the latter.

> I am wondering whether TOPS for everything but the top layer is redundant information, and hence simply can be ignored

Well - decks with TOPS values specified for all cells are valid, and in principle it is possible to create a non covering grid using a fully specified TOPS keyword. But of course - you are essentially right; for all practical purposes a fully specified TOPS keyword contains lots of redundant information.

The reason I am focusing so much on this issue is that I think/fear it is the signature of some completely unrelated bug which is hiding well. The main purpose of this PR has been to draw attention to the underlying problem - and communicate the current status as I see it.

bska:
> when all `nx*ny*nz` elements in the keyword have been specified in the input deck the simulation results are completely bogus - my guess is that this is memory corruption bug of some sort

I don't see anything that jumps out in opm-core's `GridManager::initFromEclipseGrid()`.  Since I don't know the code in question I'll just ask: Do we create reasonable initial pore volumes when specifying a complete `TOPS` vector?  Otherwise, that is at least a **possible** source of confusion.

atgeirr:
My current working assumption is that in the creation of the ecl_grid `z_current = z_previous + dz` and similar constructs end up giving different (such as 5.99999999 vs. 6.00000001 and similar) zcorn for cells in different layers, when they should have been equal. By default we use no tolerance when interpreting the data, which leads to the layers not being connected (and having 500 nodes with a top layer TOPS, and 700 nodes with a full TOPS). The use of zero tolerance is tied to the assumption that it all comes from input: if you input 5.9999999 and 6.00000001 you meant them to be different. However, for computed values that is not correct.

While this may not be the only issue, it certainly is an issue. The functions *ecl_grid_alloc_dx_dy_dz_tops* and *ecl_grid_init_zcorn_data__* of ERT are relevant here (especially the first).

joakim-hove:
>My current working assumption is that in the creation of the ecl_grid z_current = z_previous + dz and similar constructs end up giving different (such as 5.99999999 vs. 6.00000001 and similar) zcorn for cells in different layers, when they should have been equal.

Thanks a lot - that makes perfect sense. I have created a suggested fix: #526

joakim-hove:
Closing this - merging #526

Issue #522, opened 2015-07-20
Problem with TOPS keyword


larsblatny:
The below applies to the SPE1CASE2.DATA file at `opm-data/spe1_new/`:

I noticed that if one only specifies the depth of the top layer in TOPS, Flow will produce the 'correct' results (the same results as Eclipse). 
If the depths of all layers are specified, Flow does not produce the correct results. 
Could there be a bug in TOPS for Flow?

joakim-hove:
>Could there be a bug in TOPS for Flow?

Certainly it could - do you base this on the final simulation results, or on a EGRID comparison?

larsblatny:
I compared all the graphs (WBHP, WGIR, WOPR, and so on) which can be produced by Flow with the corresponding graphs from Eclipse. I guess that would be the "final simulation results" (?)

joakim-hove:
> I guess that would be the "final simulation results" (?)

Yes - that is what I meant with "final simulation results". Would be very convenient for me if you still have the .EGRID files from ECLIPSE and Flow with all layers specified in TOPS - could you mail them to `joaho at statoil.com`?

joakim-hove:
I agree that graphs of summary properties like WBHP, WGIR and WOPR are all wrong, and the only specifying the top layer in the TOPS keyword seems to fix the problem, however I do not think the bug(s) is with the basic `TOPS` treatment in opm-parser. My findings so far are: [In the summary below flow results should be interpreted as flow results *with all layers specified in the TOPS keyword* - with only the top layer specified in the TOPS keyword everything seems to work.]

1. Using METRIC units instead of field units the flow results agree with eclipse.
2. When using FIELDS units there is a significant difference in the initial pressure between eclipse and flow.

Will continue looking into this. Based on these temporary findings I think that: https://github.com/OPM/opm-data/pull/53 and https://github.com/OPM/opm-data/pull/54 actually hide a real problem (with the EQUIL initialisation??) - so I would prefer those PR's to be reverted.

larsblatny:
Thank you for your comments!

alfbr:
I fail to see how reverting the PRs will help. The files are cleaner with only the top layer specified, so I believe that is how they should be anyway. However, we can make an integration test with TOPS given for all layers.

joakim-hove:
> Will continue looking into this. [...] (with the EQUIL initialisation??) 

I really do not understand this - but the EQUIL initialisation seems to be off the hook.

GitPaean:
[...]

PR #526, opened 2015-07-24
When fully specified TOPS agrees within 1e-6 with TOPS(k=1) and DZ: Enforce bitwise equality

joakim-hove:
In #522 problems were reported when the TOPS keyword was fully specified. The code path from input deck to `struct UnstructeredGrid` is as follows:

1. Load input keywords from deck and create ert ecl_grid.
2. From the ert ecl_grid: export `ZCORN, COORD` and `ACTNUM` data.
3. In opm-core/GridManager(): create `struct UnstructeredGrid`

The problem was that when the TOPS keyword was fully specified in the input deck the bottom of cells in layer k and the top of cells in layer k + 1 were not bitwise equal, and the grid construction process would not set the cells in contact. With the current fix bitwise equality is enforced if the following condition is met:
```txt
   TOPS(k) + DZ(K) =~ TOPS(k+1) 
``` 
with a precision of 1e-6.

More details can be found in: #525.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
+++ b/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
@@ -20,6 +20,7 @@
 
 #include <iostream>
 #include <tuple>
+#include <cmath>
 
 #include <boost/lexical_cast.hpp>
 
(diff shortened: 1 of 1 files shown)


alfbr:
Have you tested this on the SPE1 deck with fully specified TOPS and confirmed that it fixes the discrepancies?

joakim-hove:
>Have you tested this on the SPE1 deck with fully specified TOPS and confirmed that it fixes the discrepancies?

Yes.

alfbr:
I also suggest a more precise title/description. This patch only enforces equality when the difference in zcorn values from TOPS and DZ is less than 1e-6.

PR #527, opened 2015-07-24
Fine grained control over behavior when errors/inconsistencies are encountered during parse and EclispeState construction. 

joakim-hove:
Have introduced a simple struct `ParseMode` which can be used to fine tune the behavior when the  parsing / EclipseState construction encounters errors and inconsistencies in the input. Each of the settings can take three values: `InputError::IGNORE, InputError::WARN` and `InputError::THROW_EXCEPTION`.

Currently four settings have been introduced, mostly as a proof of concept. The plan is to introduce more of these settings, as behavioral discussions arise. With this machinery the recurring discussion of strict versus sloppy can be on a per-problem basis.

**Unsupported features**: Have previously discussed with @atgeirr - and agreed, that opm-parser should not be aware of the capabilities of downstream modules; i.e. opm-parser should be completely oblivious to the fact that opm-core can not write `GRID` files. I fully agree that this a sound design principle; however with the basis of the new `ParseMode` class I would like to suggest a rematch on the topic - it would be extremely simple (and natural?) to add a field `unsupportedGRIDOutput` to the `ParseMode` struct?

**Name**: Observe that the new struct is currently called `ParseMode` - that is clearly not a good name, as it is equally important in the post-parse EclipseState construction phase. Suggestions for a better name appreciated.

**downstream modules**: The following PRs must be merged simultaneously with the current PR. For the downstream modules the changes consist in adding a `const ParseMode&` argument to the `Parser::parseFile()` and `EclipseState::EclipseState()` calls, in addition some extra '/' have been removed from testdata.

1. https://github.com/OPM/opm-porsol/pull/130
2. https://github.com/OPM/opm-core/pull/832
3. https://github.com/OPM/opm-autodiff/pull/421
4. https://github.com/OPM/opm-polymer/pull/136
5. https://github.com/OPM/opm-upscaling/pull/154 [Merged]
6. https://github.com/OPM/ewoms/pull/24
7. https://github.com/OPM/opm-data/pull/56 [Merged]


andlaus:
I suppose this means going back to the original `parserLog` approach but calling it 'ParseMode` this time? I'm in favor of this ;)

alfbr:
I love it! It provides us with an extensible mechanism for handling parser errors. The solution for random text / unknown keyword serves as a beautiful example.

atgeirr:
This is intended to become a more complex configuration class with time, as needed? I am in favour of that, as well.

Just a nitpick: why use Opm:: in front of all things, when we already are in the Opm namespace? (I suppose we should address this en masse at some point, no need to do it now.)

joakim-hove:
I'm glad you like it :smile:

> @atgeirr: This is intended to become a more complex configuration class with time, as needed

Well - the intention is to add more fields to control the behavior in different situations, however the goal is to keep the `struct ParseMode` very simple - i.e. currently it only has data members and no methods. If possible I would like to keep it that way.

> @atgeirr: why use Opm:: in front of all things,

I have tried to follow the convention in the surrounding code, I think what confuses you is that for no good/apparent reason most of the tests in opm-parser are *not* in the Opm namespace?

> @andlaus: I suppose this means going back to the original parserLog approach but calling it ParserMode 

Well - at least in the sense that a `const ParseMode&` argument is sprinkled through many many constructors and method calls - yes.

andlaus:
>> @atgeirr: why use Opm:: in front of all things,
>
> I have tried to follow the convention in the surrounding code, I think what confuses you is that for no good/apparent reason most of the tests in opm-parser are not in the Opm namespace?

I agree with joakim: IMO, even if a given piece of code is in the `Opm` namespace, explicitly mentioning `Opm::` makes it easier to read because it is clear where a given class comes from. (this applies to code which is not very closely related to these classes.)

atgeirr:
> Suggestions for a better name appreciated.

What about ParserConfiguration? Or ParserStrictness?

joakim-hove:
The (totally unrelated) PR: https://github.com/OPM/opm-parser/pull/533 - should fix the configuration of  which report steps to write restart files for; unfortunately the work with this PR has shaken out new bugs which affect the RPTRST / RPTSCHED / ... behavior - so a fixup/followup to https://github.com/OPM/opm-parser/pull/533 will come here.

joakim-hove:
>What about ParserConfiguration? Or ParserStrictness?

Well - I would like a verb prefix; this class is not about the parser itself - but how the parser should be used - to parse.

-- Linguistic police

PR #532, opened 2015-08-03
Opm 218 fix for write of restart files

chflo:
- Added proper handling for rptsched restart; not just in terms of rptrst basic for equal logic
- Added handling of rptrst basic as integer mnemonic (old eclipse files), and test
- Fixed timemap initTimestepMonths and initTimestepYears methods (to get equal startdate as eclipse) - and corresponding tests

To get correct writing of restart files, pull requests for OPM-218 for opm-core and opm-autodiff must also be merged.


joakim-hove:
Continued in: https://github.com/OPM/opm-parser/pull/533

PR #534, opened 2015-08-05
Adding support for changing WSOLVENT during simulations

totto82:
The WSOLVENT  implementation is more or less a copy of the WPOLYMER implementation. With this change WSOLVENT fraction can change during simulations.


opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -476,6 +480,24 @@ namespace Opm {
...
+            DeckRecordConstPtr record = keyword->getRecord(recordNr);
+            const std::string& wellNamePattern = record->getItem("WELL")->getTrimmedString(0);
+            std::vector<WellPtr> wells = getWells(wellNamePattern);
+
+            for (auto wellIter=wells.begin(); wellIter != wells.end(); ++wellIter) {
+                WellPtr well = *wellIter;
+
+                WellSolventProperties properties(well->getSolventPropertiesCopy(currentStep));
+
+                properties.m_solventConcentration = record->getItem("SOLVENT_FRACTION")->getSIDouble(0);
```
joakim-hove: This is a dimensionless property and I would prefer `getRawDouble(0)`

opm/parser/eclipse/EclipseState/Schedule/Well.hpp
```diff
@@ -112,6 +117,7 @@ namespace Opm {
         std::shared_ptr<DynamicState<WellProductionProperties> > m_productionProperties;
         std::shared_ptr<DynamicState<WellInjectionProperties> > m_injectionProperties;
         std::shared_ptr<DynamicState<WellPolymerProperties> > m_polymerProperties;
+        std::shared_ptr<DynamicState<WellSolventProperties> > m_solventProperties;
```
joakim-hove: As long as the solvent is only a single number I would completely remove the `WellSolventProperties` structure and just use:
```C++
std::shared_ptr<DynamicState<double> > m_solventProperties;
```

joakim-hove:
OK - all tests are green - good!

joakim-hove:
According to the documentation the `WSOLVENT` keyword should only be used for wells which have been declared as gas detectors - that is not checked in this PR. I would prefer that the `handlWSOLVENT` is modified along the lines:

```
{
   auto injectionProperties = well->getInjectionProperties( timeStep );
   if (well->isInjector( timeStep ) && injectionProperties->injectorType == GAS)
      ...
   else
      throw std::invalid_argument("WSOLVENT keyword can only be applied to Gas injectors");
}
```

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
andlaus: actually I think SI double is a bit better: who knows how Eclipse likes to specify fractions... (I would only slightly surprised if they sometimes use "parts of 42" or something.)

totto82:
Thanks for the comments. I thought that it might by usefull to have the possibility to add properties to WSOLVENT later, but I could n'tcome up with one example, so I have simplified it as you suggested. I have added a test in the simulator to test that WSOLVENT is used for an Gas injector, but I agree the test belongs here.

opm/parser/eclipse/EclipseState/Schedule/Well.hpp
```diff
@@ -86,6 +86,10 @@ namespace Opm {
         WellPolymerProperties          getPolymerPropertiesCopy(size_t timeStep) const;
         const WellPolymerProperties&   getPolymerProperties(size_t timeStep) const;
 
+        bool                           setSolventFraction(size_t timeStep , const double fraction);
+        double                         getSolventFractionCopy(size_t timeStep) const;
```
totto82: Do I need getSolventFractionCopy?

joakim-hove: No - as long as the object stored in the `DynamicState<T>` is only a scalar double that is not needed.

joakim-hove:
OK - good; I will merge this when there is test - in particular I want a `BOOST_CHECK_THROW()` of the situation of `WSOLVENT` applied to a well which is not a gas injector.

totto82:
Do you want me to make a new test file. ala WellSolventTests.cpp or add it to some other tests? If so where?

alfbr:
Are we sure throw is the best solution here? Maybe the parsermode can one to the rescue?

joakim-hove:
> Do you want me to make a new test file. ala WellSolventTests.cpp or add it to some other tests?

That is 100% up to you - but I guess the simplest would be to create a separate test.

joakim-hove:
> Are we sure throw is the best solution here? 

Well - the manual is quite clear on how it should be; and it is really not clear to me how to recover from an error here.

> Maybe the parsermode can one to the rescue?

Well maybe - but first we must merge it! The grand plan deep inside my head for situations like this is as follows:

1. We start out strict - and throw.
2. Iff it turns out that this is a common problem - **and** - there is a natural way to recover from incorrect input we add a `ParseMode::xxxx` field for it, and selectively allow this problem to be fussed over.

For the particular issue at hand I have no clue how/when/if point 2 will be relevant - only time will show.

totto82:
Test added.

joakim-hove:
Good testing - merging!

alfbr:
> Well - the manual is quite clear on how it should be; and it is really not clear to me how to recover from an error here.

The easiest recovery would be to simply ignore the WSOLVENT input for anything but gas injectors, and issue a warning. I actually expect all OPM modules to prefer that solution to throwing. It is more of an effort to implement, so it does come with added cost and complexity. However, I believe we have to put in the extra effort needed for non-strict parsing.

joakim-hove:
>It is more of an effort to implement, so it does come with added cost and complexity. 

Let me get this absolutely clear:

1. It is **absolutely** impossible to find safe and sound alternatives for every possible example of incorrect user input. We should therefor in my opinion prioritize elaborate error handling in situations which occur often and can be recovered from meaningfully. No one in the OPM community really knows how common the situation: "Ohh shit - I specified the `WSOLVENT` keyword for a well which is not a gas injector" really is - but frankly I think it is quite rare and not yet worthy of special treatment.

2. You can't really have your cake and eat it too. When we go to a more relaxed parsing & EclipseState construction we are essentially chipping away at the safety gear - something which should not be taken lightly in my opinion. The different error conditions we can handle will unfortunately never be 100% orthogonal - meaning that the moment we ignore one particular input error the chance of bugs/errors at a later stage in *will increase*.

atgeirr:
I agree that for the moment we can throw on obviously incorrect WSOLVENT usage and change that later if we encounter a deck that we need to run as is containing such usage. I'd rather do a little work to support a new deck's bad behaviour than do a lot of work now just in case: we'd probably do it wrong anyway, without a concrete example.

The Norne example is informative in this respect: it shows that we need policies to accept a variety of things (such as random slashes), and the solution is clean. If we need a policy for that later, that is fine.

However, some tools (ResInsight perhaps) may need to be able to attempt reading data in the face of clear errors: clearly that tool should use as permissive configuration as possible (with the not-yet-merged ParseMode feature), but it may also need the ability to lower any kind of error into a warning. That would be the domain of the Log system as we have discussed earlier.

atgeirr:
One more note: here the throw happens in the Schedule class, so parsing and creating the deck would succeed, it is only when trying to make an EclipseState it will fail. So it will not prevent you from reading the grid for example, although you would be unable to extract the wells. (Meaning there still will be a need for an errors-to-warnings feature, although less than if it failed during parsing.)

alfbr:
In general I do not think we can rely on users to do the testing. While such a strategy works for the linux kernel, it will probably not work for us. The reason is simply that developers on the project are not users, and users who are not developers on the project typically do not report issues. This means we cannot take the "when it happens" approach. We need to design for it.

As it is now, we do not have a mechanism in place to handle the situation, so I support the merging. However, once parsermode is in action I hope everybody contributing a new keyword will support strict parsing, but in addition thinks through a warning scenario. If the user does an error, e.g., sets WSOLVENT for a wag injector (or any other well that is not a gas injector at the time), then simply do not inject solvent in that well, and issue an error/warning to the effect. It is probably much more efficient to think through natural warning scenarios (if there are any) now, than later. I do see that it can be a slippery slope, particularly for Schedule objects that change over time. What I am asking for is that implementers of new keywords thinks through natural warning scenarios (where a strict parsing would throw) when implementing the keyword, and at least share thoughts on it.

Issue #535, opened 2015-08-05
std::replace in StarToken.hpp


totto82:
I had to add #include <algorithm> to opm/parser/eclipse/RawDeck/StarToken.hpp to make it compile.

andlaus:
quite possible. I only tested this using fairly recent compilers (GCC 4.8 and clang 3.5), so there might be such issues lurking around. Would you mind opening a PR?

PR #536, opened 2015-08-05
Add <algorithm> to make it compile for gcc 4-6

totto82:
see #535


joakim-hove:
Well; I can merge this - no problem. But we are targeting gcc version 4.8 now, so this might be a loosing battle.

PR #537, opened 2015-08-06
Add support for PVDS

totto82:
-the PVDS implementation is essentially a copy of the PVDG
implementation as the PVDS table analogous the PVDG table

### DIFF
--- a/opm/parser/eclipse/CMakeLists.txt
+++ b/opm/parser/eclipse/CMakeLists.txt
@@ -221,6 +221,7 @@ EclipseState/Tables/PvtgInnerTable.hpp
 EclipseState/Tables/PvdoTable.hpp
 EclipseState/Tables/MultiRecordTable.hpp
 EclipseState/Tables/PvdgTable.hpp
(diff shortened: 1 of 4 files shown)


atgeirr:
Why not use the PVDG table class then?

andlaus:
> Why not use the PVDG table class then?

that's basically because of the fundamental design of the table abstraction classes: the names of the columns are probably slightly different between PVTG and PVTS, the columns could have different constraints, etc. Besides this, the reason for having a separate class for each table is that in the higher layers you're forced to go on different code paths for different functionality.

If you look at the patch, you'll see that the class itself is not much copy-and-pasted code. (and the boiler-plate code in EclipseState would be required in any case...)

joakim-hove:
The new file `PvdsTable.hpp` is missing from the list of included files.

joakim-hove:
I really think the table classes should be refactored one shiny day - but I will merge this now (when the `CMakeLists.txt` file has been updated).

totto82:
The new file is now added.

PR #541, opened 2015-08-07
Changed VFPPRod string -> enum conversion routines.

joakim-hove:


opm/parser/eclipse/EclipseState/Tables/VFPProdTable.cpp
```diff
@@ -377,32 +380,39 @@ VFPProdTable::GFR_TYPE VFPProdTable::getGFRType(std::string gfr_string) {;
...
-    }
-    else if (alq_string == "COMP") {
-        return ALQ_COMP;
-    }
-    else if (alq_string == "BEAN") {
-        return ALQ_BEAN;
-    }
-    else if (alq_string == " ") {
+VFPProdTable::ALQ_TYPE VFPProdTable::getALQType(std::shared_ptr<const DeckItem> item) {
+    if (item->defaultApplied(0))
```
babrodtk: I like to have braces around if's (personal preference from previous experiences with bugs sneaking in). Do you object to adding
```
if (item->defaultApplied(0)) {
    return ALQ_UNDEF;
}
```
?

joakim-hove: Hehe - I realize I am in a quite desperate minority here, but quite frankly I really don't like those braces - I think they make the code uglier and less readable. But as long as the rest of the method is written with braces I'll succumb.

PR #546, opened 2015-08-08
Reimplmented ParseMode based on string keys.

joakim-hove:
With this PR the implementation of the ParseMode class has changed completely. Instead of hardwiring  the error scenarios the implementation is based on a map strings and expected action. The main advantage of this is that it becomes easier and more flexible to add update the policy.

When updating a policy you send a string which can contain '?' and '*' and several settings split by '|'. So to ignore all errors arising from unsupported features and random '/' found in the input:

````
parserMode.update("UNSUPPORTED_*|PARSER_RANDOM_SLASH" , InputError::IGNORE)
```

By default the ParseMode object will be instantiated with all settings set to `InputError::THROW_EXCEPTION` - but before returning the constructor will consult the environment variables `OPM_ERRORS_IGNORE, OPM_ERRORS_WARN` and `OPM_ERRORS_EXCEPTION` - if these variables have been set the value will be sent to the `update( )`function.

Followup PR's to the other modules will come; I suspect and hope only opm-autodiff will actually need an update.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
@@ -179,15 +179,8 @@ namespace Opm {
 
 
             if (unsupportedModifiers.find( keyword->name() ) != unsupportedModifiers.end()) {
-                auto action = parseMode.unsupportedScheduleGeoModifiers;
-
-                if (action != InputError::IGNORE) {
-                    std::string msg = "OPM does not support grid property modifier " + keyword->name() + " in the Schedule section. Error at report: " + std::to_string( currentStep );
-                    if (action == InputError::THROW_EXCEPTION)
-                        throw std::invalid_argument( msg );
-                    else if (action == InputError::WARN)
-                        OpmLog::addMessage(Log::MessageType::Warning , msg );
-                }
+                std::string msg = "OPM does not support grid property modifier " + keyword->name() + " in the Schedule section. Error at report: " + std::to_string( currentStep );
+                parseMode.handleError( ParseMode::UNSUPPORTED_SCHEDULE_GEO_MODIFIER , msg );
             }
         }
 
@@ -243,16 +236,8 @@ namespace Opm {
         for (const auto record : (*compordKeyword)) {
             auto methodItem = record->getItem<ParserKeywords::COMPORD::ORDER_TYPE>();
             if (methodItem->getString(0) != "TRACK") {
-                auto action = parseMode.unsupportedCOMPORDType;
-                if (action != InputError::IGNORE) {
-                    std::string msg = "The COMPORD keyword only handles 'TRACK' order. [ParseMode::unsupportedCOMPORDType]";
-
-                    if (action == InputError::THROW_EXCEPTION)
-                        throw std::invalid_argument(msg);
-                    else
-                        OpmLog::addMessage(Log::MessageType::Warning , msg );
-
-                }
+                std::string msg = "The COMPORD keyword only handles 'TRACK' order.";
+                parseMode.handleError( ParseMode::UNSUPPORTED_COMPORD_TYPE , msg );
             }
         }
     }
--- a/opm/parser/eclipse/EclipseState/SimulationConfig/ThresholdPressure.cpp
+++ b/opm/parser/eclipse/EclipseState/SimulationConfig/ThresholdPressure.cpp
@@ -104,14 +104,8 @@ namespace Opm {
                         m_thresholdPressureTable[r1 + maxEqlnum*r2] = p;
                         m_thresholdPressureTable[r2 + maxEqlnum*r1] = p;
                     } else {
-                        auto action = parseMode.unsupportedInitialTHPRES;
-                        if (action != InputError::IGNORE) {
-                            std::string msg = "Inferring threshold pressure from the initial state is not supported. [ParseMode::unsupportedInitialTHPRES]";
-                            if (action == InputError::THROW_EXCEPTION)
-                                throw std::invalid_argument( msg );
-                            else
-                                OpmLog::addMessage(Log::MessageType::Warning , msg );
-                        }
+                        std::string msg = "Inferring threshold pressure from the initial state is not supported.";
+                        parseMode.handleError( ParseMode::UNSUPPORTED_INITIAL_THPRES , msg );
                     }
                 } else
                     throw std::runtime_error("Missing region data for use of the THPRES keyword");
--- a/opm/parser/eclipse/EclipseState/SimulationConfig/tests/ThresholdPressureTest.cpp
+++ b/opm/parser/eclipse/EclipseState/SimulationConfig/tests/ThresholdPressureTest.cpp
@@ -222,7 +222,7 @@ BOOST_AUTO_TEST_CASE(ThresholdPressureThrowTest) {
         BOOST_CHECK_THROW(std::make_shared<ThresholdPressure>(parseMode , deck, gridPropertiesEQLNUMall0), std::runtime_error);
     }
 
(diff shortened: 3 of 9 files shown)


joakim-hove:
> hm. this function seems to implement an ERT specific glob-matching algorithm. since opm-parser uses full-blown regular expressions elsewhere (the std::regex stuff), wouldn't it be a good idea to apply it here as well?

Maybe? The reason I have used the `util_fnmatch()` function is twofold:

 1. I wanted the `fnmatch` semantics as I feel that is simpler and more intuitive when the full blown regex machinery is not called for.
 2. The `util_fnmatch()` function is portable - it is essentially a very thin wrapper around `fnmatch()`on Posix and some other function on Win32.

joakim-hove:
>The more philosophical question of this simplifies things more than it reduces the discoverability of these switches should be better taken to the mailing list IMO.

Well - where it is discussed is not so important to me; but more opinions is certainly valuable. From a pure developer point of view the previous approach is probably better, but I think the current approach is more suited for someone creating a runtime system. With the current approach e.g. Resinsight can do:

```c++
for (auto pair : parseMode) {
     // Add elements to a gui panel.
}
// Let the user update the gui and then call parseMode.update( )
```
Something like that will be much more painful to implement with the hardwired settings.

joakim-hove:
>The util_fnmatch() function is portable

And the reason I suddenly care about (windows) portability is that I hope Resinsight will start using the parser, and then Windows compatibility suddenly becomes an issue.

bska:
Can I just make a tangential comment?  The statement
```C++
for (auto i : coll) { /* ... */ }
```
should, in general, really be written
```C++
for (auto& i : coll) { /* ... */ }
```
to avoid copying items from the collection `coll`.  This holds even if `coll` is a `std::vector<int>`...

When/if we have C++17 this, in turn, should by default be shortened to
```C++
for (i : coll) { /* ... */ }
```
which uses the forwarding reference declaration
```C++
for (auto&& i : coll) { /* ... */ }
```
under the bonnet.

andlaus:
> Maybe? The reason I have used the util_fnmatch() function is twofold:
>
> 1)    I wanted the fnmatch semantics as I feel that is simpler and more intuitive when the full blown regex machinery is not called for.
> 2)  The util_fnmatch() function is portable - it is essentially a very thin wrapper around fnmatch()on Posix and some other function on Win32.

I won't deny that it is has a simpler syntax (extended glob() expressions?), my point is rather that it is *non-standard* (i.e., it needs to be learned just for this purpose, while regular expression usage is widespread in all kinds of utilities and knowledge can this more likely be re-used). Further, `std::regex` is mandated by the c++-2011 standard so it should also be quite portable...

andlaus:
> but I think the current approach is more suited for someone creating a runtime system

I agree. Probably the documentation needs to be pretty comprehensive, then.

joakim-hove:
> [...] (extended glob() expressions?), 

It is *the* glob syntax - glob is based on the `fnmatch()` function.

>(i.e., it needs to be learned just for this purpose, while regular expression usage is widespread in all kinds of utilities and knowledge can this more likely be re-used

Well the glob syntax is used in every conceivable shell I know - including the DOS prompt, so here I disagree.

joakim-hove:
>  [...] to avoid copying items from the collection col

Ahhh  this language - countless ways to do it wrong. I'll get the & habit for these range based loops as well. I actually thought 15 seconds about it and thought - "come on this automagical `auto`" ought to get this one right.

andlaus:
> It is the glob syntax - glob is based on the fnmatch() function.

really? I though the "or" operator in glob-style matches was "{foo,bar}" instead of "foo|bar"...

joakim-hove:
>really? I though the "or" operator in glob-style matches was "{foo,bar}" instead of "foo|bar"...

OK - no not the 'or' operator. The splitting on "|" or ":" is done *prior to* calling the `fnmatch()` function.

andlaus:
> OK - no not the 'or' operator. The splitting on "|" or ":" is done prior to calling the fnmatch() function. 

now I have two questions:

1) is `fnmatch()` really available on windows? if yes, does it adhere to the same syntax as on UNIX?
2) `{foo,bar}|baz` is equivalent to `foo|bar|baz` and `{foo,bar,baz}` for `util_fnmatch()`, right?

bska:
> Ahhh this language - countless ways to do it wrong

Scott Meyers gave a one-hour talk on type deduction at last year's CppCon and it's on [YouTube](https://www.youtube.com/watch?v=wQxj20X-tIU).  Be prepared for information overload...

joakim-hove:
This is to much fussing over details. But anyway:

> 1) is fnmatch() really available on windows? if yes, does it adhere to the same syntax as on UNIX?

Well - frankly I don't know 100%. The windows alternative implements '*' matching similarly to `fnmatch()` - and frankly that is all I have cared about previously - and also all I care about for this particular case; it should be documented though.

> 2) {foo,bar}|baz is equivalent to foo|bar|baz and {foo,bar,baz} for util_fnmatch(), right

1. The current implementation is a two step process: 1) Split the string on '|' or ':'. 2) if the components contain '*' - do a matching based on `util_fnmatch()`.
2. On posix `util_fnmatc() == fnmatch()` - on windows I don't know.

andlaus:
> Well - frankly I don't know 100%. The windows alternative implements '*' matching similarly to fnmatch() - and frankly that is all I have cared about previously - and also all I care about for this particular case; it should be documented though.

why don't you just use `std::regex`? You'll get rid of all these issues and it is, well, `std::`. But actually, `util_fnmatch()` is okay with me too if you keep insisting on it...

joakim-hove:
> [...] if you keep insisting on it...

What I am insisting is that I can write in the manual: 

" ... You can use '*' as a wildcard when specifying the error name, e.g.:
```txt
setenv OPM_ERRORS_IGNORE UNSUPPORTED_*
```
will make sure that all errors starting with 'UNSUPPORTED_' will be ignored' 

I.e. the pattern string comes from the user - and I think that for the user the `fnmatch()` based syntax is far simpler than the regex syntax. If we used std::regex the paragraph in the manual would be rewritten:

```txt
setenv OPM_ERRORS_IGNORE ^UNSUPPORTED_(.)*
```
(I think). To me that is quite a lot less user friendly.

bska:
Just go with `fnmatch()`.  Having to support arbitrary user-defined regular expressions is a nightmare in waiting.

alfbr:
Actually wikipedia has an overview of what is supported of globbing on different platforms:
https://en.wikipedia.org/wiki/Glob_(programming)
but I am afraid I have no opinion either way on regex or globbing, both should work fine as long as users know which it is. I am more eager to see how this new functionality plays out, thanks for the great work!

bska:
> Actually wikipedia has an overview of what is supported of globbing on different platforms:

Thanks for the link.  Essentially, globbing is a lot better understood by and easier to explain to the average user than regular expressions and forcing the user base to travel in the business of regular expressions just to be able to make a run-time decision for parsing mode is the wrong tradeoff.

joakim-hove:
>globbing is a lot better understood by and easier to explain to the average user than regular expressions

That is **exactly** my point.

alfbr:
Sure, but I hope this functionality is not intended for average users.

joakim-hove:
> but I hope this functionality is not intended for average users.

The main intention of the functionality which is the focus of this discussion is the following:

1. The application, e.g. `flow` is built with a certain error handling policy.
2. A user is running `flow` and the application encounters a problem which according to the compiled-in policy behavior leads to a hard crash.
3. The opm customer care center can then tell the user: "OK - that is because OPM does not properly support xxxx - however if you are adventurous you can fasten the seat belt, try setting the environment variable `OPM_IGNORE_ERRORS` to xxxx and try again'.

PR #547, opened 2015-08-10
Add warning-suppression utility headers.

atgeirr:
This should be merged simultaneously with the corresponding PR in opm-core that removes the header from there. Please review the (small) CMake changes, as I am not intimately familiar with how things should be done in opm-parser (since it does not use the exact same approach to adding files as the other modules do). The actual headers are the same as before.

This is a pragmatic move to make it possible to use this functionality in opm-parser. In the longer term the headers should end up in the topmost core module of opm (which no longer opm-core can be said to be). Keeping the paths identical means that no client code needs to change.


atgeirr:
Please hold a little. I think the UseWarnings module must be used as well. Currently testing.

atgeirr:
With the last commit, the feature works as intended on my system.

PR #549, opened 2015-08-10
Fix warnings

atgeirr:
With this, opm-parser compiles for me with no warnings.


atgeirr:
Hold again, please. One more commit coming in.

atgeirr:
I have added two more warnings to the disable list: unused-parameter is triggered by some boost includes in release mode only. This is a familiar pattern, when a parameter is only used in an assert().

The cast-align is triggered by some MPI headers (OpenMPI).

PR #550, opened 2015-08-10
Add support for sat func family 2 keywords

totto82:
Saturation functions input using SGFN and SOF3


joakim-hove:
Any idea how many more of these tables are coming; a refactor of this table code is way overdue.

totto82:
This is (hopefully) the last one from me in a while. Andreas is supposed to implement SLGOF soon. But the are numerous ways of implementing  saturation and pvt function so there may be need for more tables in the future.

totto82:
Tests to make sure saturation function families are not mixed are on its way.

andlaus:
> Any idea how many more of these tables are coming

that probably comes down to how many table keywords reservoir engineers use in the decks we want to support. I guess that except for SLGOF, the ones used by Model 2 are handled after this PR...

PR #552, opened 2015-08-12
Updates for VFPINJ

babrodtk:
Minor fixes that make vfpinj and vfpprod tables behave more similarly.


opm/parser/share/keywords/000_Eclipse100/W/WCONPROD
```diff
@@ -9,7 +9,7 @@
        {"name" : "RESV"         , "value_type" : "DOUBLE" , "default" : 0.0 , "dimension" : "LiquidSurfaceVolume/Time"},
        {"name" : "BHP"          , "value_type" : "DOUBLE" , "default" : 0.0 , "dimension" : "Pressure"},
        {"name" : "THP"          , "value_type" : "DOUBLE" , "default" : 0.0 , "dimension" : "Pressure"},
-       {"name" : "VFPTable"     , "value_type" : "INT"   , "default" : 0.0 , "comment":"The default is a state variable"},
+       {"name" : "VFP_TABLE"    , "value_type" : "INT"    , "default" : 0.0 , "comment":"The default is a state variable"},
```
joakim-hove: Two comments here:

1. The default value is obviously wrong - if it is `int` it should be `0`.
2. Now when we have a real implementation of the `VFPPROD` we can maybe implement the proper default behavior and remove the cryptic "The default is a state variable"?

babrodtk: I think both your comments are well made, and I'll push some fixes for it.

joakim-hove:
Ehh - yes; but the main point about the "The default is a state comment" is that (as far as i remember) is that the `VFPPROD`keyword had something semantic like:
```txt
if defaulted(...)
   then use_previous_value();
else
   ...
```
(as you can see I do not really remember this semantics), but when we have proper VFP opjects we can and should implement this default behavior properly. (NB: I have not checked the VFP object in detail - maybe what I am dragging on about actually is implemented?).

atgeirr:
I think the removal of the comments is correct, since the defaults are concrete (0 and 0.0), and not something context-dependent like "last value". One could argue that table number defaulting to 0 needs special treatment (since table 0 does not exist and this therefore is just like saying "no table").

PR #554, opened 2015-08-13
Add reservoirVolumeTarget property for group production.

atgeirr:
Also change defaults for production rate properties to -999e100, since 0 is an actual possible rate, and while it would be possible to ask if a default was applied (the default should be no rate or limit, not a zero rate), unlike for wells there are no methods doing this currently.

### DIFF
--- a/opm/parser/share/keywords/000_Eclipse100/G/GCONPROD
+++ b/opm/parser/share/keywords/000_Eclipse100/G/GCONPROD
@@ -1,10 +1,10 @@
 {"name" : "GCONPROD" , "sections" : ["SCHEDULE"], "items" : [
         {"name" : "GROUP" , "value_type" : "STRING" },
         {"name" : "CONTROL_MODE" , "value_type" : "STRING" , "default" : "NONE"},
-        {"name" : "OIL_TARGET" , "value_type" : "DOUBLE" , "dimension" : "LiquidSurfaceVolume/Time", "default" : 0},
-        {"name" : "WATER_TARGET" , "value_type" : "DOUBLE" , "dimension" : "LiquidSurfaceVolume/Time", "default" : 0},
-        {"name" : "GAS_TARGET" , "value_type" : "DOUBLE" , "dimension" : "GasSurfaceVolume/Time", "default" : 0},
-        {"name" : "LIQUID_TARGET" , "value_type" : "DOUBLE" , "dimension" : "LiquidSurfaceVolume/Time", "default" : 0},
+        {"name" : "OIL_TARGET" , "value_type" : "DOUBLE" , "dimension" : "LiquidSurfaceVolume/Time", "default" : -999e100},
+        {"name" : "WATER_TARGET" , "value_type" : "DOUBLE" , "dimension" : "LiquidSurfaceVolume/Time", "default" : -999e100},
+        {"name" : "GAS_TARGET" , "value_type" : "DOUBLE" , "dimension" : "GasSurfaceVolume/Time", "default" : -999e100},
+        {"name" : "LIQUID_TARGET" , "value_type" : "DOUBLE" , "dimension" : "LiquidSurfaceVolume/Time", "default" : -999e100},
         {"name" : "EXCEED_PROC" , "value_type" : "STRING" , "default" : "NONE"},
         {"name" : "RESPOND_TO_PARENT" , "value_type" : "STRING" , "default" : "YES"},
         {"name" : "GUIDE_RATE" , "value_type" : "DOUBLE"},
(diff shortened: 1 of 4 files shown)


atgeirr:
@joakim-hove Please consider this PR together with OPM/opm-core#850 and if you choose to merge, merge both.

joakim-hove:
OK; I guess this does not make matters any worse - but I really feel we should strive to get a decent test case *with* group control, because I fear there are more bugs in the pipeline.

Will merge when the minor issue in the json configuration has been adressed.

atgeirr:
> the minor issue in the json configuration

What issue is that?

opm/parser/share/keywords/000_Eclipse100/G/GCONPROD
```diff
@@ -1,10 +1,10 @@
 {"name" : "GCONPROD" , "sections" : ["SCHEDULE"], "items" : [
         {"name" : "GROUP" , "value_type" : "STRING" },
         {"name" : "CONTROL_MODE" , "value_type" : "STRING" , "default" : "NONE"},
-        {"name" : "OIL_TARGET" , "value_type" : "DOUBLE" , "dimension" : "LiquidSurfaceVolume/Time", "default" : 0},
-        {"name" : "WATER_TARGET" , "value_type" : "DOUBLE" , "dimension" : "LiquidSurfaceVolume/Time", "default" : 0},
-        {"name" : "GAS_TARGET" , "value_type" : "DOUBLE" , "dimension" : "GasSurfaceVolume/Time", "default" : 0},
-        {"name" : "LIQUID_TARGET" , "value_type" : "DOUBLE" , "dimension" : "LiquidSurfaceVolume/Time", "default" : 0},
+        {"name" : "OIL_TARGET" , "value_type" : "DOUBLE" , "dimension" : "LiquidSurfaceVolume/Time", "default" : -999e100},
```
joakim-hove: I have strived hard to keep such *raving-mad-values-as-indicators-of-default* out of the Json configuration; the goal is that any default value mentioned in the Json configuration should be usable as is. If the client is interested in knowing whether the default has been applied it should use the `item->defaultApplied()` method instead of comparing with a more or less funny value.

Even in in this particular case I do not think the value `-999e100` is required in any way, the code in `Group.cpp` initializes the rate vectors with the `INVALID_GROUP_RATE` value - and as long as there is no `GCONPROD` keyword in the deck the value given here will not even be consulted.

I will back off if your empirical evidence shows otherwise, but if that is not the case I would like these defaults to remain at 0.

atgeirr: The default in these cases should not be zero, but "no rate at all". I agree that this must be handled eventually with the defaultApplied(), but the default equal to zero is misleading, and an invalid rate is therefore better.

Since in JSON these things can be strings, could we use "default" : "no rate" and treat that somehow? I am not familiar with how the JSON is actually parsed and used, so it may not be feasible.

atgeirr: > if your empirical evidence shows otherwise

The empirical evidence is that the group-related tests in opm-core fail (but I am not 100% sure that they fail without this default, it could be that the rest of the patch is enough).

joakim-hove: Well - I hope and think (barring the empirical evidence) that no default at all should work - I would prefer that.

>Since in JSON these things can be strings, could we use "default" : "no rate" and treat that somehow? I am not familiar with how the JSON is actually parsed and used, so it may not be feasible.

From a Json point of view we could do that - but somewhere in opm-parser there would be type exception in that case.

joakim-hove: > no default at all should work

i.e.
```json
{"name" : "OIL_TARGET" , "value_type" : "DOUBLE" , "dimension" : "LiquidSurfaceVolume/Time"},
{"name" : "WATER_TARGET" , "value_type" : "DOUBLE" , "dimension" : "LiquidSurfaceVolume/Time"},
{"name" : "GAS_TARGET" , "value_type" : "DOUBLE" , "dimension" : "GasSurfaceVolume/Time"},
{"name" : "LIQUID_TARGET" , "value_type" : "DOUBLE" , "dimension" : "LiquidSurfaceVolume/Time"},
```

atgeirr: > no default at all should work

I do not think so, because then the parser would complain about a missing item, right?

joakim-hove:
> I do not think so, because then the parser would complain about a missing item, right?

Yes - that is right; however it would pass the deck-without-`GCONxxxx`-keywords test - which is the current situation. The Right &#8482; way to do it in the parser would be:

```C++
#include <opm/parser/eclipse/ParserKeywords.hpp>
...
... 
{
     auto oilTarget = record->getItem<ParserKeywords::GCONPROD::OIL_TARGET>();
     if (!oilTarget->defaultApplied(0))
         group->setOilTargetRate(oilTarget->getSIDouble(0));
}
...
```

but if you think that is just too much hassle I'll merge the current state.

atgeirr:
Maybe I was unclear, but we do have some tests that verify basic input of a deck with group controls in opm-core, and I can confirm that they fail if I do not set the defaults. I therefore ask you to merge this as is.

I have to admit I do not know where you would put the code you wrote above, in EclipseState?

joakim-hove:
OK - I'll merge.

>I have to admit I do not know where you would put the code you wrote above, in EclipseState?

That would be the `Schedule::handleGCONxxx()` method.

PR #555, opened 2015-08-13
Opm 217 IOConfig restart write

chflo:
- Extended IOConfig with support for RPTSOL RESTART (write of initial restart file)
- Fix for RPTRST BASIC=4/5 with frequency  
- Added IOConfig Integration test

### DIFF
--- a/opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
+++ b/opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
@@ -38,10 +38,6 @@ namespace Opm {
         m_ignore_RPTSCHED_RESTART(false){
     }
 
-    bool IOConfig::getWriteInitialRestartFile() const {
-      return m_write_initial_RST_file;
-    }
-
     bool IOConfig::getWriteEGRIDFile() const {
         return m_write_EGRID_file;
     }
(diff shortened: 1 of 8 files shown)


opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
```diff
@@ -109,8 +114,8 @@ namespace Opm {
                     write_restart_file = true;
                 } else {
                     std::vector<size_t>::const_iterator ci_start = timesteps.begin();
-                    int index = std::distance( ci_start, ci_timestep );
-                    if( (index % frequency) == 0) {
+                    int dist = std::distance( ci_start, ci_timestep ) + 1;
```
joakim-hove: Ahh that +1 looks funny; can you explain it?

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
```diff
@@ -37,6 +38,10 @@ namespace Opm {
         m_ignore_RPTSCHED_RESTART(false){
     }
 
+    bool IOConfig::getWriteInitialRestartFile() const {
```
joakim-hove: This method should be removed - the required special casing of step 0 should be internal to the `getWriteRestartFile()` method - i.e
```C++
bool IOConfig::getWriteRestartFile(size_t timeStep) {
    if (timeStep == 0)
       return m_write_initial_RST_file;
    else {
       ....
   }
}
```


nstead the `getWriteRestartFile()

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
chflo: For instance take the list of timesteps: " 3 4 5 6 7 8 9 12 16 19 21 27 ", frequency is 6, so every 6th of these should be written. 
The first one to be written is timestep 8, next one is 27. std::diff on iterators for timestep 8 and begin of timestelist is 5, so to get every 6th element I added the 1 to the dist integer. Please feel free to suggest a more elegant solution.

PR #556, opened 2015-08-14
Saturation function family II

totto82:
This PR adds
1. A function to get the saturation function family number.
2. Make use of the family 2 tables to find end points and critical points in the tables in the saturation function property initializer


opm/parser/eclipse/EclipseState/EclipseState.cpp
```diff
@@ -412,6 +412,29 @@ namespace Opm {
         initFullTables(deck, "PVTO", m_pvtoTables);
    }
 
+    const size_t EclipseState::getSaturationFunctionFamily() const{
+
+        bool family1 = !m_sgofTables.empty() && !m_swofTables.empty();
```
atgeirr: In two-phase cases one of them may be empty.

opm/parser/eclipse/EclipseState/EclipseState.cpp
```diff
@@ -412,6 +412,29 @@ namespace Opm {
...
+        bool family2 = !m_swfnTables.empty() && !m_sgfnTables.empty() && !m_swfnTables.empty();
```
atgeirr: You repeat one of the queries, furthermore as above, it fails for two-phase runs.

opm/parser/eclipse/EclipseState/EclipseState.hpp
```diff
@@ -153,6 +153,12 @@ namespace Opm {
         const std::map<int, VFPInjTable>& getVFPInjTables() const;
         size_t getNumPhases() const;
 
+        // The saturation function family.
+        // If SWOF and SGOF are specified in the deck it return 1
+        // If SWFN, SGFN and SOF3 are specified in the deck it return 2
+        // If keywords are missing or mixed, an error is given.
+        const size_t getSaturationFunctionFamily() const;
```
atgeirr: I can see that this method is useful internally when building tables, but should it not be private?

andlaus: it is needed in the opm-material code as well. I agree with the enum, but I can't really come up with good names.

joakim-hove: >it is needed in the opm-material code as well. I agree with the enum, but I can't really come up with good names.

As long as there are only two families I guess it could be a `bool`?

atgeirr: enum { FamilyI, FamilyII } ? Or is that too similar? Just curious: why is it needed in opm-material? Is this not abstracted away well enough by the EclipseState?

andlaus: > Just curious: why is it needed in opm-material? Is this not abstracted away well enough by the EclipseState?

that's because the columns of the tables must be initialized differently for every family. you can argue that this should be done into opm-parser, but if this is supposed to simplify matters in consumer modules, there must be a canonical representation for all families and I'm not sure if this is (easily) possible.

joakim-hove: > [...] you can argue that this should be done into opm-parser [ ...] must be a canonical representation for all the families. 

That has been my "wet dream" for the saturation tables in the parser - but I have no clue how large a job that would be.

I have no strong feelings if the switch should be an enum or a bool - but `size_t` is in my opinion even worse than int.

andlaus: > That has been my "wet dream" for the saturation tables in the parser - but I have no clue how large a job that would be.

I also think it would be nice, but I also don't really see how this can be done because there may be 2D tables, etc. the only thing which I can come up with is to merge the fluid-matrix interactions into the parser. (IMO, that would throw out the baby with the bathwater, though.)

atgeirr: The 2d tables settles it for me: we cannot hope to unify all saturation specification methods. So now I find it totally reasonable that the method is private (that started this discussion)...

totto82: Then I will change it to enum and make it private. (and add something similar to opm-material)

totto82:
The last changes includes 
- moving saturation function family check from EclipseState to
SatfuncPropertyInitializers
- using enum instead of size_t
- adding tests in SatfuncPropertyInitializersTests
- Fixing error in the SatfuncPropertyInitializers implementation for
familyII
- Fixing bug in ISWCR defaults  (I have not yet tested if this effects norne, I will report back)

totto82:
> Fixing bug in ISWCR defaults (I have not yet tested if this effects norne, I will report back)
The bug was in SWCR and not ISWCR and SWCR is specified for all cells i.e. no default values are used and the bug therefore had no effect on norne.

opm/parser/eclipse/EclipseState/Grid/SatfuncPropertyInitializers.hpp
```diff
@@ -54,6 +59,37 @@ class EndpointInitializer
         , m_eclipseState(eclipseState)
     { }
 
+    enum SaturationFunctionFamily { noFamily = 0, FamilyI = 1, FamilyII = 2};
+
+    void print() const
```
joakim-hove: This looks like a debugging aid?

totto82: It is. With the test in place, it can be removed.

joakim-hove:
OK - I'm ready to merge this when you say so.

totto82:
Is ready for merge.

joakim-hove:
I merged this prematurely - the tests: `runGridPropertyTests` and `runSatfuncPropertyInitializersTests` fail - my bad :rage:

joakim-hove:
The problem is this `assert( )`: https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Grid/SatfuncPropertyInitializers.hpp#L481 - initially it was:`assert( m_eclipseState.getNumPhases() == 3 )` which did not compile; I then applied the quick fix: #557 - but now the assert fails runtime when testing.

bska:
> `assert( )`

Which, of course, means that anyone building with `-DNDEBUG` (i.e., in release mode) **won't** see the `make test` failures.

PR #558, opened 2015-08-19
Add phase info to tests

totto82:
OIL, GAS and WATER is specified to avoid the 3-phase assert test in
SatfuncPropertyInitializers to fail. 

See #556


joakim-hove:
Good - thank you; merging this.

joakim-hove:
During the build test I got this install error in opm-autodiff, which is probably unrelated:

```txt
CMake Error at cmake_install.cmake:88 (file):
  file INSTALL cannot find ".../opm-autodiff-install-RH6/opm/autodiff/BlackoilModelSolvent.hpp".
```

but should be fixed anyway.

PR #560, opened 2015-08-21
Added support for new threshold pressure behavior

iLoop2:


opm/parser/eclipse/EclipseState/SimulationConfig/ThresholdPressure.hpp
```diff
@@ -37,17 +37,19 @@ namespace Opm {
...
 
-        void initThresholdPressure(const ParseMode& parseMode,
-                                   std::shared_ptr<const RUNSPECSection> runspecSection,
+        void initThresholdPressure(std::shared_ptr<const RUNSPECSection> runspecSection,
                                    std::shared_ptr<const SOLUTIONSection> solutionSection,
                                    std::shared_ptr<GridProperties<int>> gridProperties);
 
 
         std::vector<double> m_thresholdPressureTable;
+        bool m_hasThresholdPressure;
```
joakim-hove: This should be explicitly initialized to `true` - if a an empty THPRES keyword has been used to initialize we can handle an empty vector, without any error handling.

atgeirr: I may be misreading you, but if there is no THPRES keyword then hasThresholdPressure() should return false, and the above comment made me wonder if that would hold.

PR #563, opened 2015-08-24
Support slgof

andlaus:
as far as I can see, SLGOF is the same as SGOF but with the saturation column defined as So + Swco (instead of Sg), so everything is reversed. (wow! what a wonderful concept!) Also this is probably the last table (except for, maybe, for SOF2 and friends) which is required to support the family 1 saturation functions fully.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Tables/SlgofTable.hpp
+++ b/opm/parser/eclipse/EclipseState/Tables/SlgofTable.hpp
@@ -1,0 +1,78 @@
+/*
+  Copyright (C) 2015 by Andreas Lauser
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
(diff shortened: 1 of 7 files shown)


totto82:
I believe getSaturationFunctionFamily must be updated as well. Now familyI is SWOF && (SGOF || SLGOF)

opm/parser/eclipse/EclipseState/Tables/SlgofTable.hpp
```diff
@@ -0,0 +1,78 @@
...
+                  int recordIdx)
+        {
+            ParentType::init(keyword,
+                             std::vector<std::string>{"SL", "KRG", "KROG", "PCOG"},
+                             recordIdx,
+                             /*firstEntityOffset=*/0);
+
+            ParentType::checkNonDefaultable("SL");
+            ParentType::checkMonotonic("SL", /*isAscending=*/true);
+            ParentType::applyDefaultsLinear("KRG");
```
totto82: KRG should be decreasing or level
KROG should be increasing or level
PCOG should be decreasing or level
Shouldn't we add a test for this?

andlaus: right. fixed.

> Shouldn't we add a test for this?

Ok, I added one. Though I don't trust it too much, because I think that this keyword an example where you make your test match the implementation, not vice-versa...

andlaus:
> I believe getSaturationFunctionFamily must be updated as well. Now familyI is SWOF && (SGOF || SLGOF) 

true. fixed.

totto82:
I didn't mean to ask for a unit test, but just that the input data was checked. (As it is now). But I guess test are always welcome.

joakim-hove:
I have long since given up reviewing these table PR's - although my desire to refactor this grows stronger for every table :wink: 

All tests are in the green so I am ready to merge this when you say so.

andlaus:
please merge. the simulator and opm-data side seems to be ready, too: https://github.com/OPM/opm-data/pull/60

Issue #561, opened 2015-08-24
About the offending string


GitPaean:
Components of pressure drop in `WELSEGS` has three different values, `HFA`, `HF-` and `H--`. 

When I tried to test parsing some example with `H--`, I got the following parsing error. 

```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Input string is not a complete record string, offending string:      'PROD01' 2512.5 2512.5 1.0e-5 'ABS' 'H      2         2      1      1    2537.5 2537.5  0.3   0.00010 /
```
It looks like no problem happened when parsing `HFA` and `HF-`. 

The the data for `WELSEGS`  is as follows, 
```
WELSEGS
     'PROD01' 2512.5 2512.5 1.0e-5 'ABS' 'H--' 'HO' /
      2         2      1      1    2537.5 2537.5  0.3   0.00010 /
      3         3      1      2    2562.5 2562.5  0.2  0.00010 /
/
```
I am wondering if parser will have some problem with string like `H--`. 

Thanks.

joakim-hove:
> I am wondering if parser will have some problem with string like H--.

Ohh yes - it happened: https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/Parser/Parser.cpp#L449

Will look into fixing the parser to handle quoted `--`

joakim-hove:
OK Kai - cool if you could give this: https://github.com/OPM/opm-parser/pull/564 a spin.

GitPaean:
Problem solved with OPM/opm-parser#564, closing now.

PR #564, opened 2015-08-24
Handle quotes when stripping off comments.

joakim-hove:
With this we should handle case like this: https://github.com/OPM/opm-parser/issues/561


GitPaean:
I just tested with my DATA file. It works. Thanks for the fixing.

joakim-hove:
>I just tested with my DATA file. It works. 

OK - thanks for testing; I will merge now.

PR #565, opened 2015-08-25
Thpres

joakim-hove:
Using a map based approach to internalizing THPRES values.

### DIFF
--- a/opm/parser/eclipse/EclipseState/SimulationConfig/SimulationConfig.cpp
+++ b/opm/parser/eclipse/EclipseState/SimulationConfig/SimulationConfig.cpp
@@ -39,4 +39,9 @@ namespace Opm {
         return m_ThresholdPressure;
     }
 
+
+    bool SimulationConfig::hasThresholdPressure() const {
+        return m_ThresholdPressure->size() > 0;
+    }
+
 } //namespace Opm
--- a/opm/parser/eclipse/EclipseState/SimulationConfig/SimulationConfig.cpp
+++ b/opm/parser/eclipse/EclipseState/SimulationConfig/SimulationConfig.cpp
@@ -35,8 +35,8 @@ namespace Opm {
     }
 
 
-    const std::vector<std::pair<bool,double>>& SimulationConfig::getThresholdPressureTable() const {
-        return m_ThresholdPressure->getThresholdPressureTable();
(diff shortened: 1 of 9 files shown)


joakim-hove:
> Any reason not to implement this as

Not really - I considered that; the immediate reason I did not do it was that then the tests:
```C++
if (deck->hasKeyword("EQLOPTS") && deck->hasKeyword("THPRES")) 
....
```
had to be moved to the `SimulationConfig` object.

bska:
>> Any reason not to implement this as [...]
>
> Not really - I considered that; the immediate reason I did not do it was that then the tests:

Sorry, I don't understand why.  The member function `SimulationConfig::hasThresholdPressure()` already referenced `m_ThresholdPressure` so whatever constraints where in place do not change by switching to my suggested implementation.

joakim-hove:
>Sorry, I don't understand why. 

Well - I might be the one misunderstanding here? As I understand it the test 
```
m_thresholdPressure->empty()
```

check whether the smart pointer points to *something*? The current implementation will unconditionally create a `ThresholdPressure` instance - so the `m_thresholdPressure` will always point to *something*. However if the deck does not contains any `THPRES` configuration the `ThresholdPressure` instance will have zero elements - i.e. `size() == 0`.

That said - your approach could very well be a better implementation.

bska:
> As I understand it the test
```C++
m_thresholdPressure->empty()
```
check whether the smart pointer points to something?

Ah, right.  No, that's not the case.  The above is equivalent to
```C++
(*m_thresholdPressure).empty()
```
(i.e., call `empty()` on the object contained in the pointer).  You're probably thinking of the conceptual statement
```
m_thresholdPressure.empty()
```
(which doesn't compile because `std::shared_ptr<>` doesn't have an `empty()` method).  That would inspect the pointer itself rather than the contained object.

> the `m_thresholdPressure` will always point to *something*. However if the deck does not contains any THPRES configuration the `ThresholdPressure` instance will have zero elements - i.e., `size() == 0`.

Okay, if there is no `ThresholdPressure::empty()` then I'd just say
```C++
bool SimulationConfig::hasThresholdPressure() const {
    return m_ThresholdPressure->size() > 0;
}
```

joakim-hove:
> You're probably thinking of the conceptual statement

yes - that was what I was thinking of.

PR #572, opened 2015-09-02
Fix minor bug and warnings

atgeirr:
This fixes a minor bug and silences some warnings.

Many of the warnings were from -Wshadow, and some pointed to real problems (readability at least) if not bugs (see the change to Schedule.cpp), I suggest including that warning by default.


opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -263,14 +263,14 @@ namespace Opm {
 
                 DeckTimeStepConstPtr deckTimeStep = section->getDeckTimeStep(currentStep);
                 if (deckTimeStep->hasKeyword("COMPORD")) {
-                    DeckKeywordConstPtr keyword = deckTimeStep->getKeyword("COMPORD");
```
joakim-hove: This was just a renaming to better names - or am I actually missing something?

joakim-hove:
Thanks - will merge; just a minor question about the changes to the Schedule.cpp?

PR #573, opened 2015-09-02
Changed throw => catch to an if test for VFP tables

babrodtk:
This changes a couple of superfluous throws into if-tests instead. Should have no side effects hopefully.


atgeirr:
It is always good to avoid using exceptions for non-exceptional cases. Supported.

joakim-hove:
This was an obvious improvement - thank you!

PR #578, opened 2015-09-08
changed: sync with renaming to opm-common

akva2:
*do not merge* before renaming with friends has happened.


joakim-hove:
Sorry for being dense - but it is not really clear to me what is the merge signal?

akva2:
https://github.com/OPM/opm-common/pull/35

PR #584, opened 2015-09-17
Plymax bug

joakim-hove:
Fixing regressions with PLYMAX and PLYROCK tables.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Tables/PlymaxTable.hpp
+++ b/opm/parser/eclipse/EclipseState/Tables/PlymaxTable.hpp
@@ -34,11 +34,14 @@ namespace Opm {
          * \brief Read the PLYMAX keyword and provide some convenience
          *        methods for it.
          */
-        void init(Opm::DeckItemConstPtr item)
+        void init(Opm::DeckRecordConstPtr record)
         {
-            SimpleTable::init(item,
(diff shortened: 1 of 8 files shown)


GitPaean:
The results are verified with the polymer case in opm-data. Thanks for the fixing.

opm/parser/eclipse/EclipseState/Tables/PlymaxTable.hpp
```diff
@@ -34,11 +34,14 @@ namespace Opm {
          * \brief Read the PLYMAX keyword and provide some convenience
          *        methods for it.
          */
-        void init(Opm::DeckItemConstPtr item)
+        void init(Opm::DeckRecordConstPtr record)
         {
-            SimpleTable::init(item,
-                              std::vector<std::string>{"C_POLYMER", "C_POLYMER_MAX"});
-
+            createColumns(std::vector<std::string>{"C_POLYMER", "C_POLYMER_MAX"});
```
GitPaean: do you think the names are a little misleading? I think for PLYMAX is for maximum polymer concentration and maximum salinity.

joakim-hove: Thanks for testing - I'll merge as is now; but I am the process of reworking these things anyway. Will come back to your suggestion.

PR #585, opened 2015-09-21
Table container

joakim-hove:
With this PR a small container `Opm::TableContainer` is created to hold the different tables, instead of `std::vector<TableType>` - work in progress.

1. opm-porsol: https://github.com/OPM/opm-porsol/pull/134
2. opm-material: https://github.com/OPM/opm-material/pull/62
3. opm-core: https://github.com/OPM/opm-core/pull/886  
4. opm-polymer: https://github.com/OPM/opm-polymer/pull/147
5. opm-autodiff: https://github.com/OPM/opm-autodiff/pull/488

The failing test in opm-core should be investigated more - then this should be ready to merge.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Tables/TableContainer.hpp
+++ b/opm/parser/eclipse/EclipseState/Tables/TableContainer.hpp
@@ -74,7 +74,6 @@ namespace Opm {
           table N is not implemented use table N - 1 behavior.
         */
         size_t hasTable(size_t tableNumber) const;
-
         const SimpleTable& getTable(size_t tableNumber) const;
         const SimpleTable& operator[](size_t tableNumber) const;
 
--- a/opm/parser/eclipse/EclipseState/Tables/TableContainer.hpp
+++ b/opm/parser/eclipse/EclipseState/Tables/TableContainer.hpp
@@ -1,0 +1,96 @@
+/*
+  Copyright 2015 Statoil ASA.
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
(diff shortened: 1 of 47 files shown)


blattms:
I have just glanced at this commit and am by no means an expert concerning the parser, but one thing strikes me:

What is the rational behind moving from a vector of tables to a map<int,table>? Are there that many indices without an associated table that the memory savings outweight the added complexity when access a table (previously index operation, now a search)?

joakim-hove:
Thank you for looking.

>What is the rational behind moving from a vector of tables to a map? 

Well - as I see it the main change was to change from a public STL container (i.e. `std::vector<T>`) to a custom container - i.e. `TableContainer` which can hide it's internal implementation. It is currently based on `std::map<>` - but that could change.

>Are there that many indices without an associated table that the memory savings outweight the added complexity when access a table (previously index operation, now a search)?

No - this purely a developer convenience/maintainability thing. The table code should support the semantics:

```
if (!hasTable(N))
   getTable(N - 1)
```

In the `std::vector<>` code that was implemented with copying of tables, avoiding the table copying is a step towards a simpler and more maintainable (IMO) table code.

opm/parser/eclipse/EclipseState/Tables/TableContainer.hpp
```diff
@@ -0,0 +1,68 @@
...
+          TableContainer class is that the getTable method implements
+          the Eclipse behavior:
+
+             If table N is not implemented - use table N - 1.
+
+          The getTable() method will eventually throw an exception if
+          not even table 0 is there.
+        */
+    public:
+        TableContainer( size_t maxTables );
```
atgeirr: This constructor should be `explicit`.

opm/parser/eclipse/EclipseState/Tables/TableContainer.hpp
```diff
@@ -0,0 +1,68 @@
...
+        bool empty() const;
+        size_t size() const;
+        void addTable(size_t tableNumber , std::shared_ptr<const SimpleTable> table);
+        size_t hasTable(size_t tableNumber) const;
```
atgeirr: I would appreciate a comment about how hasTable() works: will it return false only if a call to getTable() with the same argument would fail? Or is it ok to call getTable(7) even if hasTable(7) is false (getting table 6 instead etc.).

Also, are typical table sets consecutive and indexed from 0 or 1? Are there table sets that typically have nonconsecutive numbers?

joakim-hove:
`TableContainer` constructor is explicit - and documentation added in rebased push.

atgeirr:
> documentation added

Much appreciated!

PR #587, opened 2015-09-22
Temporarily remove parser support for CPR keyword

joakim-hove:
The following has happened:

1. In https://github.com/OPM/opm-parser/pull/581 we added support for internalizing the CPR keyword.
2. In https://github.com/OPM/opm-autodiff/issues/489 @dr-robertk reported that flow/parser could no langer handle an old version of the SPE9 data.

It turns out that the old SPE9 input deck has a `CPR` keyword in the summary section: https://github.com/OPM/opm-data/blob/9af155c26bc0946d9af846ac6e503115c0267050/spe9/SPE9.DATA - this keyword has completely different schema than the keyword added in #581, and the parsing goes completely belly up. The current 'fix' is just to revert the functionality of #581 - but the longer term fix must determine one two:

1. I have found no references to `CPR` in the `SUMMARY` section in the documentation, maybe it is a "bug" in the old SPE9 deck which can be ignored (i.e. just comment out the `CPR`keyword from the summary section).
2. If it is indeed legitimate with `CPR` in the `SUMMARY` section - this suddenly implies that the parser must support section aware keyword schema - now that is a significant change.

### DIFF
--- a/opm/parser/eclipse/EclipseState/SimulationConfig/SimulationConfig.cpp
+++ b/opm/parser/eclipse/EclipseState/SimulationConfig/SimulationConfig.cpp
@@ -24,15 +24,31 @@
 
 
 
+/*
+  The internalization of the CPR keyword has been temporarily
+  disabled, suddenly decks with 'CPR' in the summary section turned
+  up. Keywords with section aware keyword semantics is currently not
+  handled by the parser.
+
+  When the CPR is added again the following keyword configuration must
+  be added:
+
+    {"name" : "CPR" , "sections" : ["RUNSPEC"], "size": 1 }
+
+*/
+
+
 namespace Opm {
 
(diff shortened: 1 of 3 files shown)


andlaus:
actually, I haven't found a CPR keyword in the reference manual. is it non-standard? In that case, I propose to move it out of the 000_Eclipse100 directory.

(actually, it is a bit more complicated: I have found the `CPR` keyword mentioned in the documentation for the summary section, but it is for outputting the connection pressure and has nothing to do with the linear solver.)

joakim-hove:
>it is a bit more complicated: I have found the CPR keyword mentioned in the documentation for the summary section, but it is for outputting the connection pressure and has nothing to do with the linear solver.)

Thank you - yes that was where I was expecting to find it, although I could not. Which version of the manual do you have - and which page number?

PR #590, opened 2015-09-24
Cpr support

joakim-hove:
The CPR keyword can appear both in the SUMMARY section and the RUNSPEC section, with completely different meaning. We just barely manage to get around this without introducing section aware keywords.

### DIFF
--- a/opm/parser/eclipse/Deck/Section.hpp
+++ b/opm/parser/eclipse/Deck/Section.hpp
@@ -40,13 +40,15 @@ class Section : public Deck
         const std::string& name() const;
         size_t count(const std::string& keyword) const;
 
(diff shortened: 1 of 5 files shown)


andlaus:
the approach seems a bit cumbersome for the user (in particular that CPR needs to be requested via a section object instead of bothering the deck directly), but I much prefer it to the alternatives. The patches themselves look good.

joakim-hove:
>this call seems to be unrelated to the rest of the patch? 

Well - it was only a white space change which tripped you up here.

> in this context, is the SimulationConfig class the correct place for the threshold pressure anyway?

Well - it has been there for a while; it enters into the simulator - that is the rationale for having it in `SimulationConfig`.

joakim-hove:
> the approach seems a bit cumbersome for the user 

Well - on the one hand this is the price we pay to avoid a large rewrite; however I also think using section objects is quite OK on a general basis.

PR #591, opened 2015-09-28
Updated the TABDIMS keyword.

joakim-hove:


bska:
> Updated the `TABDIMS` keyword.
```C++
/*
  The TABDIMS keyword has a total of 25 items; most of them
  are [...] quite exotic.
*/

So what you're saying is that it should really be called the [`TARDIS`](https://en.wikipedia.org/wiki/TARDIS) keyword :smirk:

PR #594, opened 2015-10-02
Add keywords DRSDT DRVDT VAPPARS

iLoop2:


opm/parser/eclipse/EclipseState/Schedule/OilVaporizationProperties.cpp
```diff
@@ -0,0 +1,61 @@
+#include "OilVaporizationProperties.hpp"
```
joakim-hove: Use full path to the header: `#include <opm/parser/eclipse/ ...>`

opm/parser/eclipse/EclipseState/Schedule/OilVaporizationProperties.hpp
```diff
@@ -0,0 +1,39 @@
...
+        OilVaporizationProperties();
+
+        static std::shared_ptr<OilVaporizationProperties> createOilVaporizationPropertiesDRSDT(double maximum, std::string option);
+        static std::shared_ptr<OilVaporizationProperties> createOilVaporizationPropertiesDRVDT(double maximum);
+        static std::shared_ptr<OilVaporizationProperties> createOilVaporizationPropertiesVAPPARS(double vap1, double vap2);
+        Opm::OilVaporizationEnum getType() const;
+        double getVap1() const;
+        double getVap2() const;
+        double getMaximum() const;
+        std::string getOption() const;
```
joakim-hove: Use: `const std::string& getOption() const`

joakim-hove:
The double values should be have dimensions: 
```
DRSDT -> "dimension" : "GasDissolutionFactor/Time"` 

DRVDT -> "dimension" : "OilDissolutionFactor/Time"
```

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -295,6 +309,38 @@ namespace Opm {
...
+            OilVaporizationPropertiesPtr vappars = OilVaporizationProperties::createOilVaporizationPropertiesVAPPARS(vap, density);
+            setOilVaporizationProperties(vappars, currentStep);
+
+        }
+    }
+
+    void Schedule::handleDRVDT(DeckKeywordConstPtr keyword, size_t currentStep){
+        for (size_t recordNr = 0; recordNr < keyword->size(); recordNr++) {
+            DeckRecordConstPtr record = keyword->getRecord(recordNr);
+            double max = record->getItem("DRVDT_MAX")->getRawDouble(0);
```
joakim-hove: Should use `getSIDouble()` - same with `handleDRSDT`

opm/parser/eclipse/EclipseState/Schedule/OilVaporizationProperties.hpp
```diff
@@ -0,0 +1,39 @@
...
+#include <string>
+#include <memory>
+#include <opm/parser/eclipse/EclipseState/Schedule/ScheduleEnums.hpp>
+
+namespace Opm
+{
+    class OilVaporizationProperties
+    {
+    public:
+        OilVaporizationProperties();
```
joakim-hove: Does this have to be public?

opm/parser/eclipse/EclipseState/Schedule/OilVaporizationProperties.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Opm {
+
+
+    OilVaporizationProperties::OilVaporizationProperties(){
+
+    }
+
+    double OilVaporizationProperties::getMaximum() const{
```
joakim-hove: Better name: `getMaxDRSDT()`

opm/parser/eclipse/EclipseState/Schedule/OilVaporizationProperties.hpp
```diff
@@ -0,0 +1,39 @@
...
+        double getVap2() const;
+        double getMaximum() const;
+        std::string getOption() const;
+
+    private:
+        Opm::OilVaporizationEnum m_type;
+        double m_maximum;
+        double m_vap1;
+        double m_vap2;
+        std::string m_option;
```
joakim-hove: This string can only take the values `ALL` and `FREE` and should be internalized as:
```C++
bool m_maxDRSDT_allCells;
```

opm/parser/eclipse/EclipseState/Schedule/OilVaporizationProperties.hpp
```diff
@@ -0,0 +1,39 @@
...
+        static std::shared_ptr<OilVaporizationProperties> createOilVaporizationPropertiesVAPPARS(double vap1, double vap2);
+        Opm::OilVaporizationEnum getType() const;
+        double getVap1() const;
+        double getVap2() const;
+        double getMaximum() const;
+        std::string getOption() const;
+
+    private:
+        Opm::OilVaporizationEnum m_type;
+        double m_maximum;
```
joakim-hove: Should have two variables: `m_maxDRSDT` and `m_maxDRVDT`.

opm/parser/eclipse/EclipseState/Schedule/OilVaporizationProperties.hpp
```diff
@@ -0,0 +1,39 @@
...
+        double m_vap1;
+        double m_vap2;
+        std::string m_option;
+
+        OilVaporizationProperties(Opm::OilVaporizationEnum type, double maximum, std::string option);
```
joakim-hove: Instead of having these private constructors you could just do member assignment in the static factory methods?

joakim-hove:
I know the static factory methods were my suggestion - but it looks kind of funny - at least forcing the implementation choice of `std::shared_ptr<>`. An alternative could be to have init methods?

opm/parser/eclipse/EclipseState/Schedule/OilVaporizationProperties.hpp
```diff
@@ -0,0 +1,39 @@
+#ifndef DRSDT_HPP
+#define DRSDT_HPP
+#include <iostream>
+#include <string>
+#include <memory>
+#include <opm/parser/eclipse/EclipseState/Schedule/ScheduleEnums.hpp>
+
+namespace Opm
+{
```
joakim-hove: This class implements a "tristate" between three different keywords; brief documentation of this slightly funny behavior would be good.

opm/parser/eclipse/EclipseState/Schedule/OilVaporizationProperties.cpp
```diff
@@ -0,0 +1,61 @@
+#include "OilVaporizationProperties.hpp"
+
+namespace Opm {
+
+
+    OilVaporizationProperties::OilVaporizationProperties(){
+
+    }
+
```
joakim-hove: The different `getXXX()` methods should be protected with a check of the type variable - i.e. asking for `m_maxDRSDT` when `m_type == VAPPARS` is an internal error and should give an exception.

opm/parser/eclipse/EclipseState/Schedule/OilVaporizationProperties.hpp
atgeirr: I agree -- I do not really understand the relationship between the VAPPARS, DRVDT and DRSDT keywords enough to understand why this logically belongs in a single class, in particular why (or if?) behaviour is exclusive in some sense (since you can only construct one of them for a particular instance).

joakim-hove: > in particular why (or if?) behaviour is exclusive in some sense

Yes - the behavior is exclusive; any one of the three keywords `{VAPPARS, DRSDT, DRVDT}`will cancel previous settings of the other keywords.

PR #599, opened 2015-10-06
Move warning headers

joakim-hove:
1. Added a warning in the disable_warnings header that the include path should be updated to use opm-common.

2. Updated all occurences to use the headers in opm-common

Downstream PRs:

1. https://github.com/OPM/opm-core/pull/896
2. https://github.com/OPM/opm-autodiff/pull/493
3. https://github.com/OPM/dune-cornerpoint/pull/179

In addition there is a PR to opm-common: which makes opm-common a **REQUIRED** dependency for `opm-core, dune-cornerpoint` and `opm-autodiff`: https://github.com/OPM/opm-common/pull/52

You will only affected by this if you build with `-DSILENCE_EXTERNAL_WARNINGS` - and *if* I have missed something the worst consequence is that you will get a preprocessor warning during the build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,6 +119,7 @@ include( cmake/Modules/CheckCaseSensitiveFileSystem.cmake )
 add_definitions("-DHAVE_CASE_SENSITIVE_FILESYSTEM=${HAVE_CASE_SENSITIVE_FILESYSTEM}")
 
 find_package(opm-common)
+include_directories( ${opm-common_INCLUDE_DIR} )   
 include( Findopm-data )
 
 find_package(ERT)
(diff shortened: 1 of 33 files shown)


CMakeLists.txt
```diff
@@ -119,6 +119,7 @@ include( cmake/Modules/CheckCaseSensitiveFileSystem.cmake )
 add_definitions("-DHAVE_CASE_SENSITIVE_FILESYSTEM=${HAVE_CASE_SENSITIVE_FILESYSTEM}")
 
 find_package(opm-common)
+include_directories( ${opm-common_INCLUDE_DIR} )
```
atgeirr: Is something like this necessary in other modules, too? Or is this due to opm-parser having more of a special build structure?

atgeirr:
A good way to approach this, I approve. Feel free to merge in all modules where I am maintainer, after build testing (critical point: that the headers are found).

CMakeLists.txt
joakim-hove: No - that works automagically in the other modules.

joakim-hove:
> A good way to approach this

Thank you :-)

PR #603, opened 2015-10-08
Shut wells with zero rates and crossflow not allowed

totto82:
Note: Producing wells in prediction mode is kept open even if the
control rate is set to zero.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
@@ -362,6 +362,16 @@ namespace Opm {
                 updateWellStatus( well , currentStep , status );
                 if (well->setProductionProperties(currentStep, properties))
                     m_events.addEvent( ScheduleEvents::PRODUCTION_UPDATE , currentStep);
+
+                if ( !well->getAllowCrossFlow() && !isPredictionMode && (properties.OilRate + properties.WaterRate + properties.GasRate) == 0 ) {
+
+                    std::string msg =
+                            "Well " + well->name() + " is a history matched well with zero rate where crossflow is banned. " +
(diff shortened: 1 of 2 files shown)


opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -610,6 +629,15 @@ namespace Opm {
 
             if (well->setInjectionProperties(currentStep, properties))
                 m_events.addEvent( ScheduleEvents::INJECTION_UPDATE , currentStep );
+
+            if ( ! well->getAllowCrossFlow() && injectionRate == 0 ) {
```
joakim-hove: Hello; can I have a paranthesis as: `(injectionsRate == 0)`?

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -525,8 +535,17 @@ namespace Opm {
                         throw std::invalid_argument("Tried to set invalid control: " + cmodeString + " for well: " + wellNamePattern);
                     }
                 }
+
                 if (well->setInjectionProperties(currentStep, properties))
                     m_events.addEvent( ScheduleEvents::INJECTION_UPDATE , currentStep );
+
+                if ( ! well->getAllowCrossFlow() && properties.surfaceInjectionRate == 0 ) {
```
joakim-hove: `(properties.surfaceInjectionRate == 0)`

joakim-hove:
The PR has three nearly identical codeblocks shutting the well, that could be in method on the `Well`object.

joakim-hove:
Will merge then `( ... )` comment has been addressed.

PR #604, opened 2015-10-09
Add support for PINCH item 4 and 5.

qilicun:
Add support for pinchout option and multz option.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
+++ b/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
@@ -172,12 +172,12 @@ namespace Opm {
             auto record = deck->getKeyword<ParserKeywords::PINCH>( )->getRecord(0);
             auto item = record->getItem<ParserKeywords::PINCH::THRESHOLD_THICKNESS>( );
             m_pinch.setValue( item->getSIDouble(0) );
-            if (record->getItem<ParserKeywords::PINCH::PINCHOUT_OPTION>()->getString(0) != "TOPBOT") {
-                m_pinchoutMode = PinchMode::ModeEnum::ALL;
-            }
-            if (record->getItem<ParserKeywords::PINCH::MULTZ_OPTION>()->getString(0) != "TOP") {
-                m_multzMode = PinchMode::ModeEnum::ALL;
-            }
+
+            auto pinchoutString = record->getItem<ParserKeywords::PINCH::PINCHOUT_OPTION>()->getString(0);
+            m_pinchoutMode = PinchMode::PinchModeFromString(pinchoutString);
+            
+            auto multzString = record->getItem<ParserKeywords::PINCH::MULTZ_OPTION>()->getString(0);
+            m_multzMode = PinchMode::PinchModeFromString(multzString);
         }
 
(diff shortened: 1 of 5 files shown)


qilicun:
This PR is needed by https://github.com/OPM/opm-core/issues/892

joakim-hove:
These items are essentially boolean (or even better enumaration) values; in such situations the `EclipseState` objects should take the responsability of converting from a string. My suggestion would be to create an enum:
```
enum PinchoutMode {
    ALL = 1,
    TOPBOT = 2,
    TOP = 3
} 
```

Then both the variables can be of type `enum PincoutMode`.

qilicun:
Ok, that sounds the same treatment with MINPV mode, I will update the PR.

qilicun:
@joakim-hove 

In the PINCH case, you still need two variables to identify both item 4 and item 5, this is different with MINPV case, I think enum can't reduce the number of private members.  Could you elaborate it a bit more?

joakim-hove:
> you still need two variables to identify both item 4 and item 5

Yes - that I agree. My point was that one could introduce only one enum type, and then have two variables of that type in the `EclipseGrid` class:

```C++
enum PinchoutMode {
    ALL = 1,
    TOPBOT = 2,
    TOP = 3
} 

...

class EclipseGrid {
....
private:
   PinchoutMode m_pincoutMode;
   PinchoutMode m_multzMode;
}
```

My argument is based on the general principle: 

    The parser/EclipseState object should as much as possible try to raise the semantic quality
    of the data; that typically includes converting strings which can only have values in a limited 
    set to explicitly enumerable types.

However - I have not very detailed looked into your case; I might be barking up the wrong tree. What I suggest might be impossible - or ill advised?

>  this is different with MINPV case

I have not considered `MINPV` at all in this context.

qilicun:
@joakim-hove  This PR is updated as your suggestion

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
```diff
@@ -164,6 +172,12 @@ namespace Opm {
             auto record = deck->getKeyword<ParserKeywords::PINCH>( )->getRecord(0);
             auto item = record->getItem<ParserKeywords::PINCH::THRESHOLD_THICKNESS>( );
             m_pinch.setValue( item->getSIDouble(0) );
+            if (record->getItem<ParserKeywords::PINCH::PINCHOUT_OPTION>()->getString(0) != "TOPBOT") {
```
joakim-hove: No - please implement a `PinchMode.cpp` file with:
```C++
PinchMode::ModeEnum PinchModeFromString( const std::string& stringValue ) {
      PinchMode::ModeEnum mode;
      if (stringValue == "TOPBOT")
          mode =PinchMode::ModeEnum::TOPBOT;
      else if (stringValue == "TOP")
          ...
      else
          throw std::invalid_argument("Can not convert ....");

     return mode;
}
```

joakim-hove:
> This PR is updated as your suggestion

Good - I will merge when you have implemented the string2enum conversion function.

qilicun:
The conversion is added.

joakim-hove:
Good - merging!

PR #607, opened 2015-10-12
Do not use const for value return.

atgeirr:
What matters is the type (const or not) used at the call site. It does not matter if the method returns a const. Note that this applies to value types, not references.

Update: the reason I fix this is that it generates a warning for me.


joakim-hove:
Thank you.

PR #608, opened 2015-10-13
fix build on GCC 4.4

andlaus:
the changes are mostly trivial, but the patch required to disable a
few unit tests on that compiler.

On my machine all tests continue to pass on a more modern compiler (clang 3.5)

### DIFF
--- a/opm/parser/eclipse/Deck/Deck.cpp
+++ b/opm/parser/eclipse/Deck/Deck.cpp
@@ -65,7 +65,7 @@ namespace Opm {
             if (index < keywordList.size())
                 return keywordList[index];
             else
-                throw std::out_of_range("Keyword " + keyword + ":" + std::to_string( index ) + " not in deck.");
+                throw std::out_of_range("Keyword " + keyword + ":" + std::to_string( static_cast<long long>(index) ) + " not in deck.");
         } else
             throw std::invalid_argument("Keyword " + keyword + " not in deck.");
     }
@@ -82,7 +82,7 @@ namespace Opm {
         if (index < m_keywordList.size())
             return m_keywordList[index];
         else
-            throw std::out_of_range("Keyword index " + std::to_string( index ) + " is out of range.");
+            throw std::out_of_range("Keyword index " + std::to_string( static_cast<long long>(index) ) + " is out of range.");
     }
 
     size_t Deck::numKeywords(const std::string& keyword) const {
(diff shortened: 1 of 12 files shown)


joakim-hove:
No - sorry; we are targeting gcc 4.8 now.

andlaus:
> No - sorry; we are targeting gcc 4.8 now.

I can't really get this argument: I thought that "we are targeting gcc 4.8 now" means that even though the master occasionally breaks GCC 4.4, as long as there are patches coming in for older versions, they are merged as long as they do not cause the code to degrade significantly. has this policy changed?

joakim-hove:
> I can't really get this argument: I thought that "we are targeting gcc 4.8 now"

Well - to me *targeting gcc 4.8* means that we are free to (personally I would go further and say encouraged to) use all C++ features found in version 4.8 of gcc.  As you have demonstrated it is currently quite simple to submit patches to ensure that the code works also on gcc 4.4 - but that is just because the change from 4.4 to 4.8 is still quite recent, and people have not yet gotten around to use much of the newly available features. 

> has this policy changed?

Well - as I see it you are the one suggesting a policy change here - essentially going back to gcc 4.4?

atgeirr:
> as long as there are patches coming in for older versions, they are merged as long as they do not cause the code to degrade significantly

I might do so, but what one considers significant is very subjective. For an example relevant to this PR, range-for is a nice little improvement in readability, so I would not like to make that into a less readable traditional for loop. Basically if you want a 4.4-ok module you would need to forbid or undo all use of range-for.

I therefore agree with @joakim-hove in closing this, however you could maintain a 4.4-compatible branch yourself if you require it. I do think that will be a lot of work going forward, though.

andlaus:
> Well - as I see it you are the one suggesting a policy change here - essentially going back to gcc 4.4?

I can't see this: before each patch needed to pass the GCC 4.4 toolchain before it could be merged. for me "targeting GCC 4.8" just means that it got demoted to a "second class citizen".

Anyway, since I'd like to keep eWoms GCC 4.4 compatile for now, I'll limit that compatibility to the non-ECL parts...

joakim-hove:
> for me "targeting GCC 4.8" just means that it [ gcc 4.4? ] got demoted to a "second class citizen".

No - gcc 4.4 is not a second class citizen - it fell the off the cliff before summer.

PR #609, opened 2015-10-16
Added missing curly brace (found using sonar with cppcheck and vera++)

flikka:


atgeirr:
I ran the clang static analyzer a few months back, and found only one problem (now fixed), different tools probably give different results -- have you ran the tools you mention on all the modules?

joakim-hove:
Hmmm - I don't understand how this has even compiled?!

flikka:
Hi @atgeirr - no I have just started experimenting with it a bit, using opm-parser as the guinea pig :-) If it makes sense, we can consider adding it to the ci.opm-project.org site (using SonarQube).

akva2:
@joakim-hove because this file is not built, ParseMULTREGT.cpp is, but 
not MULTREGT.cpp

flikka:
Hehe. The file seems to be something we have forgotten to remove :-) So, lesson learnt - don't stop thinking just because you get new tools...

PR #610, opened 2015-10-16
Deleted unused file MULTREGT.cpp

flikka:
Rewrite history, or let past mistakes be a part of it? I prefer the last...


andlaus:
> Rewrite history, or let past mistakes be a part of it? I prefer the last...

+1: once something is merged into master, rewriting the branch's history to fix a technical bug is a _very_ bad idea. IMO, the only occasion where history rewriting is the best alternative (but not a good one) is dealing with copyright infringements.

PR #612, opened 2015-10-19
added/fixed: set soversion for libraries

akva2:
This was quick-fixed for last release. Now uses the information from the dune.module file to set versions.


joakim-hove:
Thanks - but where do the `${opm-parser_VERSION_MAJOR}` and `${opm-parser_VERSION_MINOR}` variables come from?

akva2:
they come from the dune.module file - and are parsed in OpmInit.cmake. thus the extra include of it. this is normally taken care of by the OpmProject call but since this buildsystem is different, it has to be manually included for a system-level/installed opm-common.

PR #615, opened 2015-10-21
Support for keywords WELSEGS and COMPSEGS

GitPaean:
This PR is created for discussion and improvement, please do not merge until we think it is ready. 

This PR is proposed to provide the current requirement for the multi-segment wells, mostly the segment structures and the relation between the segments and completions.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
@@ -47,7 +47,7 @@ namespace Opm {
         // The first record here only contains the well name
         std::vector<std::shared_ptr<Compsegs>> compsegs;
 
-        for (size_t recordIndex = 1; recordIndex < compsegsKeyword->size(); ++recordIndex) {
+        for (int recordIndex = 1; recordIndex < int(compsegsKeyword->size()); ++recordIndex) {
             DeckRecordConstPtr record = compsegsKeyword->getRecord(recordIndex);
             // following the coordinate rule for completions
             const int I = record->getItem("I")->getInt(0) - 1;
--- a/opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
@@ -40,16 +40,16 @@ namespace Opm {
     }
 
     std::vector<std::shared_ptr<Compsegs>> Compsegs::compsegsFromCOMPSEGSKeyword(DeckKeywordConstPtr compsegsKeyword,
-                                                                                       EclipseGridConstPtr grid) {
+                                                                                 EclipseGridConstPtr grid) {
+        static_cast<void>(grid); // Silence warning about unused argument
+
         // only handle the second record here
         // The first record here only contains the well name
         std::vector<std::shared_ptr<Compsegs>> compsegs;
 
-        double previous_distance_end = 0.;
-
         for (size_t recordIndex = 1; recordIndex < compsegsKeyword->size(); ++recordIndex) {
             DeckRecordConstPtr record = compsegsKeyword->getRecord(recordIndex);
-            // following the corrdinate rule for completions
+            // following the coordinate rule for completions
             int I = record->getItem("I")->getInt(0) - 1;
             int J = record->getItem("J")->getInt(0) - 1;
             int K = record->getItem("K")->getInt(0) - 1;
--- a/opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
@@ -50,10 +50,10 @@ namespace Opm {
         for (size_t recordIndex = 1; recordIndex < compsegsKeyword->size(); ++recordIndex) {
             DeckRecordConstPtr record = compsegsKeyword->getRecord(recordIndex);
             // following the coordinate rule for completions
-            int I = record->getItem("I")->getInt(0) - 1;
-            int J = record->getItem("J")->getInt(0) - 1;
-            int K = record->getItem("K")->getInt(0) - 1;
-            int branch = record->getItem("BRANCH")->getInt(0);
+            const int I = record->getItem("I")->getInt(0) - 1;
+            const int J = record->getItem("J")->getInt(0) - 1;
+            const int K = record->getItem("K")->getInt(0) - 1;
+            const int branch = record->getItem("BRANCH")->getInt(0);
 
             double distance_start;
             double distance_end;
@@ -100,12 +100,6 @@ namespace Opm {
                 center_depth = 0.;
             }
 
-            /* if (center_depth == 0.){ // dangerous?
-                //TODO: get the depth from the segement data
-                // throw std::runtime_error("this way to obtain CENTER_DISTANCE not implemented yet!");
-                center_depth = 1.e100;
-            } else */
-
             if (center_depth < 0.) {
                 //TODO: get the depth from COMPDAT data.
                 throw std::runtime_error("this way to obtain CENTER_DISTANCE not implemented yet either!");
@@ -116,8 +110,7 @@ namespace Opm {
                 thermal_length = record->getItem("THERMAL_LENGTH")->getRawDouble(0);
             } else {
                 //TODO: get the thickness of the grid block in the direction of penetration
-                // throw std::runtime_error("this way to obtain THERMAL_LENGTH not implemented yet!");
-                thermal_length = -1.e100;
+                throw std::runtime_error("this way to obtain THERMAL_LENGTH not implemented yet!");
             }
 
             int segment_number;
@@ -125,7 +118,7 @@ namespace Opm {
                 segment_number = record->getItem("SEGMENT_NUMBER")->getInt(0);
             } else {
                 segment_number = 0;
-                //TODO: decide the segment number based on the distance in a process later.
+                // will decide the segment number based on the distance in a process later.
             }
 
             if (end_IJK < 0) { // only one compsegs
@@ -133,7 +126,7 @@ namespace Opm {
                                                                       direction, center_depth, thermal_length, segment_number);
                 compsegs.push_back(new_compsegs);
             } else { // a range is defined. genrate a range of Compsegs
-                // throw std::runtime_error("this way to obtain THERMAL_LENGTH not implemented yet!");
+                throw std::runtime_error("entering COMPSEGS entries with a range is not supported yet!");
             }
         }
 
--- a/opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
@@ -1,0 +1,143 @@
+/*
+  Copyright 2015 SINTEF ICT, Applied Mathematics.
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <opm/parser/eclipse/EclipseState/Schedule/Compsegs.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/ScheduleEnums.hpp>
+#include <opm/parser/eclipse/EclipseState/Grid/EclipseGrid.hpp>
+
+namespace Opm {
+
+
+    Compsegs::Compsegs(int i_in, int j_in, int k_in, int branch_number_in, double distance_start_in, double distance_end_in,
+                       WellCompletion::DirectionEnum dir_in, double center_depth_in, double thermal_length_in, int segment_number_in)
+    : m_i(i_in),
+      m_j(j_in),
+      m_k(k_in),
+      m_branch_number(branch_number_in),
+      m_distance_start(distance_start_in),
+      m_distance_end(distance_end_in),
+      m_dir(dir_in),
+      m_center_depth(center_depth_in),
+      m_thermal_length(thermal_length_in),
+      m_segment_number(segment_number_in)
+    {
+    }
+
+    std::vector<std::shared_ptr<Compsegs>> Compsegs::compsegsFromCOMPSEGSKeyword(DeckKeywordConstPtr compsegsKeyword,
+                                                                                       EclipseGridConstPtr grid) {
+        // only handle the second record here
+        // The first record here only contains the well name
+        std::vector<std::shared_ptr<Compsegs>> compsegs;
+
+        double previous_distance_end = 0.;
+
+        for (size_t recordIndex = 1; recordIndex < compsegsKeyword->size(); ++recordIndex) {
+            DeckRecordConstPtr record = compsegsKeyword->getRecord(recordIndex);
+            // following the corrdinate rule for completions
+            int I = record->getItem("I")->getInt(0) - 1;
+            int J = record->getItem("J")->getInt(0) - 1;
+            int K = record->getItem("K")->getInt(0) - 1;
+            int branch = record->getItem("BRANCH")->getInt(0);
+
+            double distance_start;
+            double distance_end;
+            if (record->getItem("DISTANCE_START")->hasValue(0)) {
+                distance_start = record->getItem("DISTANCE_START")->getRawDouble(0);
+            } else if (recordIndex == 1) {
+                distance_start = 0.;
+            } else {
+                // TODO: the end of the previous connection or range
+                // 'previous' should be in term of the input order
+                // since basically no specific order for the completions
+                throw std::runtime_error("this way to obtain DISTANCE_START not implemented yet!");
+            }
+            if (record->getItem("DISTANCE_END")->hasValue(0)) {
+                distance_end = record->getItem("DISTANCE_END")->getRawDouble(0);
+            } else {
+                // TODO: the distance_start plus the thickness of the grid block
+                throw std::runtime_error("this way to obtain DISTANCE_END not implemented yet!");
+            }
+
+            WellCompletion::DirectionEnum direction;
+            if (record->getItem("DIRECTION")->hasValue(0)) {
+                direction = WellCompletion::DirectionEnumFromString(record->getItem("DIRECTION")->getString(0));
+            } else if (!record->getItem("DISTANCE_END")->hasValue(0)) {
+                throw std::runtime_error("the direction has to be specified when DISTANCE_END in the record is not specified");
+            }
+
+            int end_IJK;
+            if (record->getItem("END_IJK")->hasValue(0)) {
+                // following the corrdinate rule for completions
+                end_IJK = record->getItem("END_IJK")->getInt(0) - 1;
+                if (!record->getItem("DIRECTION")->hasValue(0)) {
+                    throw std::runtime_error("the direction has to be specified when END_IJK in the record is specified");
+                }
+            } else {
+                // only one completion is specified here
+                end_IJK = -1;
+            }
+
+            double center_depth;
+            if (record->getItem("CENTER_DEPTH")->hasValue(0)) {
+                center_depth = record->getItem("CENTER_DEPTH")->getRawDouble(0);
+            } else {
+                center_depth = 0.;
+            }
+
+            /* if (center_depth == 0.){ // dangerous?
+                //TODO: get the depth from the segement data
+                // throw std::runtime_error("this way to obtain CENTER_DISTANCE not implemented yet!");
+                center_depth = 1.e100;
+            } else */
+
+            if (center_depth < 0.) {
+                //TODO: get the depth from COMPDAT data.
+                throw std::runtime_error("this way to obtain CENTER_DISTANCE not implemented yet either!");
+            }
+
+            double thermal_length;
+            if (record->getItem("THERMAL_LENGTH")->hasValue(0)) {
+                thermal_length = record->getItem("THERMAL_LENGTH")->getRawDouble(0);
+            } else {
+                //TODO: get the thickness of the grid block in the direction of penetration
+                // throw std::runtime_error("this way to obtain THERMAL_LENGTH not implemented yet!");
+                thermal_length = -1.e100;
+            }
+
+            int segment_number;
+            if (record->getItem("SEGMENT_NUMBER")->hasValue(0)) {
+                segment_number = record->getItem("SEGMENT_NUMBER")->getInt(0);
+            } else {
+                segment_number = 0;
+                //TODO: decide the segment number based on the distance in a process later.
+            }
+
+            if (end_IJK < 0) { // only one compsegs
+                CompsegsPtr new_compsegs = std::make_shared<Compsegs>(I, J, K, branch, distance_start, distance_end,
+                                                                      direction, center_depth, thermal_length, segment_number);
+                compsegs.push_back(new_compsegs);
+            } else { // a range is defined. genrate a range of Compsegs
+                // throw std::runtime_error("this way to obtain THERMAL_LENGTH not implemented yet!");
+            }
+        }
+
+        return compsegs;
+    }
+
+}
--- a/opm/parser/eclipse/EclipseState/Schedule/Segment.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Segment.cpp
@@ -9,7 +9,7 @@ namespace Opm {
       m_outlet_segment(-1),
       m_length(-1.e100),
       m_depth(-1.e100),
-      m_internal_dimaeter(-1.e100),
+      m_internal_diameter(-1.e100),
       m_roughness(-1.e100),
       m_cross_area(-1.e100),
       m_volume(-1.e100),
@@ -28,7 +28,7 @@ namespace Opm {
       m_outlet_segment(outlet_segment),
       m_length(length),
       m_depth(depth),
-      m_internal_dimaeter(internal_diameter),
+      m_internal_diameter(internal_diameter),
       m_roughness(roughness),
       m_cross_area(cross_area),
       m_volume(volume),
@@ -65,7 +65,7 @@ namespace Opm {
 
 
     double& Segment::internalDiameter() {
-        return m_internal_dimaeter;
+        return m_internal_diameter;
     }
 
 
--- a/opm/parser/eclipse/EclipseState/Schedule/Segment.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Segment.cpp
@@ -4,7 +4,8 @@
 namespace Opm {
 
     Segment::Segment()
-    : m_branch(-1),
+    : m_segment_number(-1),
+      m_branch(-1),
       m_outlet_segment(-1),
       m_length(-1.e100),
       m_depth(-1.e100),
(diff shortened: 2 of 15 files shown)


opm/parser/eclipse/EclipseState/Schedule/Segment.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+    Segment::Segment(int segment_number, int branch, int outlet_segment, double length, double depth,
+                     double internal_diameter, double roughness, double cross_area,
+                     double volume, double length_x, double length_y, bool data_ready)
+    : m_segment_number(segment_number),
+      m_branch(branch),
+      m_outlet_segment(outlet_segment),
+      m_length(length),
+      m_depth(depth),
+      m_internal_dimaeter(internal_diameter),
```
atgeirr: Typo in the variable name.

GitPaean: fixed. Thanks.

joakim-hove:
Ohhh my - this was large; review will take some time!

opm/parser/eclipse/EclipseState/Schedule/Segment.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+    class Segment {
+    public:
+        Segment();
+
+        Segment(int segment_number, int branch, int outlet_segment, double length, double depth,
+                double internal_diameter, double roughness, double cross_area,
+                double volume, double length_x, double length_y, bool data_ready);
+
+        int& segmentNumber();
```
joakim-hove: Hmmm - looks really weird with reference here; as the code stands now you will be able to modify the internals of segment object with:
```C++

double length = segment.getLength();
length *= 2;

```
I assume that this is a mistake; i.e. that the methods should have returned scalar values `double / int / bool`. If the use of references is actually intentional I will say that the API must be changed.

opm/parser/eclipse/EclipseState/Schedule/Well.cpp
```diff
@@ -240,6 +248,209 @@ namespace Opm {
...
+            throw std::runtime_error(" unknow length_depth_type in the new_segmentset");
+        }
+    }
+
+    void Well::addSegmentSetINC(size_t time_step, const SegmentSetPtr new_segmentset, const bool first_time) {
+        // The following code only applied when no loop exist.
+        // where loops exist, some modification should be made.
+        if (first_time) {
+            // update the information inside new_segmentset to be in ABS way
+            (*new_segmentset)[0]->length() = new_segmentset->lengthTopSegment();
```
joakim-hove: Hmmm - looks like the use of references was intentional? I have not looked detailed into this yet - but be prepared; you are going to have a really hard time convincing me that this way of manipulating the internal state of the segment objects is kosher.

opm/parser/eclipse/EclipseState/Schedule/Well.cpp
```diff
@@ -240,6 +248,209 @@ namespace Opm {
...
+               int outlet_segment = (*new_segmentset)[i]->outletSegment();
+               int outlet_location = new_segmentset->numberToLocation(outlet_segment);
+               double segment_length = (*new_segmentset)[i]->length() - (*new_segmentset)[outlet_location]->length();
+               if ((*new_segmentset)[i]->volume() < 0.5 * meaningless_value) {
+                   (*new_segmentset)[i]->volume() = (*new_segmentset)[i]->crossArea() * segment_length;
+               }
+            }
+
+            // for debugging output
+            bool output_debugging = false;
```
joakim-hove: We have proper tests - not `std::cout` in the main code - sorry :-(

opm/parser/eclipse/EclipseState/Schedule/Segment.hpp
GitPaean: At beginning it was a struct. I forgot why I changed it to be a trivial class. Now I think it is good to just be a simple struct. How do you think?

opm/parser/eclipse/EclipseState/Schedule/Well.cpp
GitPaean: Okay. I will add a test.

opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
```diff
@@ -0,0 +1,136 @@
...
+        static_cast<void>(grid); // Silence warning about unused argument
+
+        // only handle the second record here
+        // The first record here only contains the well name
+        std::vector<std::shared_ptr<Compsegs>> compsegs;
+
+        for (size_t recordIndex = 1; recordIndex < compsegsKeyword->size(); ++recordIndex) {
+            DeckRecordConstPtr record = compsegsKeyword->getRecord(recordIndex);
+            // following the coordinate rule for completions
+            const int I = record->getItem("I")->getInt(0) - 1;
```
joakim-hove: I am trying to push the use of of the compile time constants in all new code; i.e.
```C++
#include <opm/parser/eclipse/Parser/ParserKeywords.hpp>
....
const int I = record->getItem<ParserKeywords::COMPSEGS::I>()->getInt(0) - 1;
```

opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
```diff
@@ -0,0 +1,136 @@
...
+      m_distance_start(distance_start_in),
+      m_distance_end(distance_end_in),
+      m_dir(dir_in),
+      m_center_depth(center_depth_in),
+      m_thermal_length(thermal_length_in),
+      m_segment_number(segment_number_in)
+    {
+    }
+
+    std::vector<std::shared_ptr<Compsegs>> Compsegs::compsegsFromCOMPSEGSKeyword(DeckKeywordConstPtr compsegsKeyword,
```
joakim-hove: Maybe the `Compsegs` constructor could take a `DeckKeyword` - so that all of this decoding was in that class, and not in the general well class?

joakim-hove:
OK - this was just a very very early start; I think it would be better for all of us  - if this PR was split, with one feature at a time. The first should probably just contain the `Compsegs` implementation. This that sound feasible?

opm/parser/eclipse/EclipseState/Schedule/Segment.hpp
joakim-hove: > How do you think?

Don't know yet; but as you say it is currently a mix. Got to go now - continue tomorrow.

GitPaean:
It is doable, since the code is already there. It can be split to be three PRs. One is for WELSEGS, the second one will be the COMPSEGS. The third one will be how to link COMPSEGS and WELSEGS. 
It is also more doable to make them two. I will handle it tomorrow.

opm/parser/eclipse/EclipseState/Schedule/Segment.hpp
GitPaean: fixed. Thanks for pointing out.

opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
GitPaean: I guess you mean `ParserKeyword.hpp`. 

Could you please tell a little more? I did not figure out how it works.

GitPaean:
The best way to split this PR is to do it manually by copying code and rewriting the whole commit history. At beginning I had two branches for WELSEGS and COMPSEGS separately, while the implementation of COMPSEGS depends on WELSEGS and I gave up maintaining two branches. I will try to create two separate PRs later to cover this PR.

opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
GitPaean: A little more specific information is also requested here. 
I believe some parts can be moved to the COMPSEGS, while not all of them. Some parts may not be easy when information of existing completions and segments are required. These keywords can be re-entered partly later to change some of the segment structures and configurations.

joakim-hove:
> I will try to create two separate PRs later to cover this PR.

For me it will simpler to review smaller PR's - but I can manage the full PR as well, however that will take longer time. It is really up to you - I will continue reviewing the full PR now at least.

opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
joakim-hove: > I guess you mean ParserKeyword.hpp.

No - I actually mean `#include <opm/parser/eclipse/Parser/ParserKeywords.hpp>` - that is a header which is generated as part of the build process; you should not need to read it - but feel free to take a  look in `${build_dir}/generated-source/include/..`

GitPaean:
Thanks. Let us keep the PR this way for the moment. I will go ahead to add testing for the PR now. 
Feel free to suggest to split the PR anytime later.

opm/parser/eclipse/EclipseState/Schedule/SegmentSet.cpp
```diff
@@ -0,0 +1,277 @@
...
+        } else if (m_length_depth_type == WellSegment::ABS) {
+            SegmentPtr top_segment(new Segment(1, 1, 0, m_length_top, m_depth_top, meaningless_value, meaningless_value,
+                                               meaningless_value, m_volume_top, m_x_top, m_y_top, true));
+            new_segments[0] = top_segment;
+        }
+
+        // read all the information out from the DECK first then process to get all the required information
+        for (size_t recordIndex = 1; recordIndex < welsegsKeyword->size(); ++recordIndex) {
+            DeckRecordConstPtr record = welsegsKeyword->getRecord(recordIndex);
+            int K1 = record->getItem("SEGMENT1")->getInt(0);
```
joakim-hove: Please do not use variable names `K1` and `K2` - it gives associations to the `z` coordinate.

opm/parser/eclipse/EclipseState/Schedule/SegmentSet.cpp
```diff
@@ -0,0 +1,277 @@
...
+        copy->m_segments.resize(m_segments.size());
+        for (size_t i = 0; i < m_segments.size(); ++i) {
+            copy->m_segments[i] = m_segments[i];
+        }
+        return copy;
+    }
+
+    void SegmentSet::segmentsFromWELSEGSKeyword(DeckKeywordConstPtr welsegsKeyword, DeckKeywordConstPtr nsegmxKeyword) {
+        // get the required information for multi-segment wells
+        const int nsegmx = nsegmxKeyword->getRecord(0)->getItem("NSEGMX")->getInt(0);
```
joakim-hove: Are these variables just Eclipse style memory management, or do we actually need them for semantic correctness?

opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
```diff
@@ -0,0 +1,136 @@
...
+            } else {
+                // only one completion is specified here
+                end_IJK = -1;
+            }
+
+            double center_depth;
+            if (record->getItem("CENTER_DEPTH")->hasValue(0)) {
+                center_depth = record->getItem("CENTER_DEPTH")->getRawDouble(0);
+            } else {
+                center_depth = 0.;
```
joakim-hove: Hmmm - `center_depth == 0` does not seem to be a valid value? Unless simulation will continue happily with `center_depth == 0` I really think you should raise an exception here.

GitPaean: `center_depth == 0.0` means defaulted. it will be obtained based on the information of segments later.

opm/parser/eclipse/EclipseState/Schedule/Well.hpp
```diff
@@ -108,7 +112,19 @@ namespace Opm {
 
 
 
+        // for multi-segment wells
+        bool isMultiSegment() const;
+        void setMultiSegment(const bool is_multi_segment);
```
joakim-hove: Most these segments should be private.

opm/parser/eclipse/EclipseState/Schedule/Well.cpp
```diff
@@ -303,11 +526,109 @@ namespace Opm {
...
+        SegmentSetPtr currentSegmentSet = m_segmentset->get(time_step);
+
+        for (size_t i_comp = 0; i_comp < compsegs.size(); ++i_comp) {
+            if (compsegs[i_comp]->m_segment_number == 0) { // need to determine the segment number first
+                const double center_distance = (compsegs[i_comp]->m_distance_start + compsegs[i_comp]->m_distance_end) / 2.0;
+                const int branch_number = compsegs[i_comp]->m_branch_number;
+
+                int segment_number = 0;
+                double min_distance_difference = 1.e100; // begin with a big value
+                for (int i_segment = 0; i_segment < currentSegmentSet->numberSegment(); ++i_segment) {
```
joakim-hove: Use `size_t` in `for( ... )` loops.

opm/parser/eclipse/EclipseState/Schedule/Well.cpp
```diff
@@ -303,11 +526,109 @@ namespace Opm {
...
+                    if (branch_number == (*currentSegmentSet)[i_segment]->branchNumber()) {
+                        const double distance = (*currentSegmentSet)[i_segment]->length();
+                        const double distance_difference = std::abs(center_distance - distance);
+                        if (distance_difference < min_distance_difference) {
+                            min_distance_difference = distance_difference;
+                            segment_number = (*currentSegmentSet)[i_segment]->segmentNumber();
```
joakim-hove: Looks like you are hijacking the segment number from an existing segment?

GitPaean: You mean the name confusing here? What things are done here is assign each completion a segment by comparing the distance.

opm/parser/eclipse/EclipseState/Schedule/Well.cpp
```diff
@@ -303,11 +526,109 @@ namespace Opm {
...
+    SegmentSetConstPtr Well::getSegmentSet(size_t time_step) const {
+        return m_segmentset->get(time_step);
+    }
+
 
     bool Well::getAllowCrossFlow() const {
         return m_allowCrossFlow;
     }
 
+    void Well::processCOMPSEGS(size_t time_step, std::vector<CompsegsPtr>& compsegs) {
```
joakim-hove: I do not think the `Well` class is correct place for this implementation.

opm/parser/eclipse/EclipseState/Schedule/Well.cpp
```diff
@@ -303,11 +526,109 @@ namespace Opm {
...
+            }
+
+            CompletionPtr new_completion = std::make_shared<Completion>(newCompletionSet->get(ic));
+            new_completion->setSegmentNumber(compsegs[i_comp]->m_segment_number);
+            new_completion->setCenterDepth(compsegs[i_comp]->m_center_depth);
+            newCompletionSet->add(new_completion);
+        }
+
+        for (size_t ic = 0; ic < newCompletionSet->size(); ++ic) {
+            if (newCompletionSet->get(ic)->getSegmentNumber() == -1) {
```
joakim-hove: Can we instead have a new method `bool Completion::attachedToSegment()` - I strongly dislike spreading arbitrary invalid values like '-1' to the rest of the code.

opm/parser/eclipse/EclipseState/Schedule/Well.cpp
GitPaean: fixed. I was wondering how to remove the warnings about comparison between int and size_t.

opm/parser/eclipse/EclipseState/Schedule/Well.cpp
joakim-hove: I am struggling to understand what happens here (please: clearer code - not comments!). First of all - what is the main purpose of this method? Further questions:

1. What is the interaction between the input `std::vector<CompsegsPtr>& compsegs` and the state field `m_segmentset`?
2. Method seems to contain roughly two parts: 

   1. go through all the segments and update `segment_number` and `center_depth`.
   2. update the completions base on the updates segments.

   Could these two main tasks be split in two methods?

opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
```diff
@@ -0,0 +1,136 @@
...
+                    throw std::runtime_error("the direction has to be specified when END_IJK in the record is specified");
+                }
+            } else {
+                // only one completion is specified here
+                end_IJK = -1;
+            }
+
+            double center_depth;
+            if (record->getItem("CENTER_DEPTH")->hasValue(0)) {
+                center_depth = record->getItem("CENTER_DEPTH")->getRawDouble(0);
```
joakim-hove: I guess this should use `getSiDouble()`- the depth has dimension length.

opm/parser/eclipse/EclipseState/Schedule/Well.cpp
GitPaean: 1. `m_segmentset` contains the segment structure and some information about each segment. 
2. `std::vector<CompsegsPtr>& compsegs` contains the location information of each completion in the multi-segment wells. 
3. The function firstly based on the location information from `std::vector<CompsegsPtr>& compsegs` find the segment that the completion should belong to and update the center-depth based on the segment information. (in multi-segment wells, the depth of the perforation is not that related to the cell center anymore). 
4. With updated information from `std::vector<CompsegsPtr>& compsegs`, update `m_completions`.

opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
joakim-hove: OK - but observe that the test `if (record->getItem("CENTER_DEPTH")->hasValue(0))` will **always** return true. The point is that `hasValue( )` does *not* differentiate between the cases "value set in input deck" and "using default value" - they are both valid values; the `hasValue( )` method is mostly relevant when an item does not have a default value. I would instead reccommend:

```C++
// Use Type value for center_depth: https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Util/Value.hpp - to get some support for default handling.

{
     const auto depthItem = record->getItem<ParserKeywords::COMPSEGS::CENTER_DEPTH>(0);
     if (!depthItem->defaultApplied(0))
          center_depth.setValue( depthItem->getSIDouble(0) );
}
```

GitPaean: You are right. It is possible some people can set a zero value there, although it can be a little abnormal.

opm/parser/eclipse/EclipseState/Schedule/SegmentSet.hpp
```diff
@@ -0,0 +1,106 @@
...
+        double depthTopSegment() const;
+        double lengthTopSegment() const;
+        double volumeTopSegment() const;
+        double xTop() const;
+        double yTop() const;
+
+        WellSegment::LengthDepthEnum lengthDepthType() const;
+        WellSegment::CompPresureDropEnum compPressureDrop() const;
+        WellSegment::MultiPhaseModelEnum multiPhaseModel() const;
+        std::vector<SegmentPtr>& Segments();
```
joakim-hove: This class exposes both `operator[]` and `std::vector<SegmentPtr>` - that is one too much. I have come to dislike exposing stl containers in a public api - as a general rule. So I would strongly prefer that only `operator[]` is supported.

opm/parser/eclipse/EclipseState/Schedule/SegmentSet.hpp
```diff
@@ -0,0 +1,106 @@
...
+#include <opm/parser/eclipse/EclipseState/Schedule/ScheduleEnums.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/Segment.hpp>
+
+namespace Opm {
+
+    class SegmentSet {
+    public:
+        SegmentSet();
+
+        std::string wellName() const;
```
joakim-hove: Now - this is obviosuly a collection og segments; what defines the collection? Is it all the segments for one well? Or all the segments in a branch? Or?

GitPaean: All the segments are in one well, since one WELSEGS can only handle one well. Re-entering WELSEGS sometimes may only contain some of the changed segments. I am not sure if it will be all the segments for the well.

opm/parser/eclipse/EclipseState/Schedule/SegmentSet.hpp
GitPaean: I forgot the reason, probably to reduce the length of each line of the code. I will fix it.

opm/parser/eclipse/EclipseState/Schedule/SegmentSet.cpp
```diff
@@ -0,0 +1,277 @@
...
+            }
+
+            // how to handle the logical relations between lateral branches and parent branches.
+            // so far, the branch number has not been used.
+            int branch = record->getItem("BRANCH")->getInt(0);
+            if ((branch < 1) || (branch > nlbrmx)) {
+                throw std::logic_error("illegal branch number input is found in WELSEGS!\n");
+            }
+            int outlet_segment = record->getItem("JOIN_SEGMENT")->getInt(0);
+            double diameter = record->getItem("DIAMETER")->getRawDouble(0);
```
joakim-hove: `getSIDouble()`

opm/parser/eclipse/EclipseState/Schedule/SegmentSet.cpp
```diff
@@ -0,0 +1,277 @@
...
+            DeckItemConstPtr itemArea = record->getItem("AREA");
+            double area;
+            if (itemArea->hasValue(0)) {
+                area = itemArea->getRawDouble(0);
```
joakim-hove: `getSIDouble()`

opm/parser/eclipse/EclipseState/Schedule/SegmentSet.cpp
```diff
@@ -0,0 +1,277 @@
...
+            double volume;
+            if (itemVolume->hasValue(0)) {
+                volume = itemVolume->getRawDouble(0);
+            } else if (m_length_depth_type == WellSegment::INC) {
+                volume = area * segment_length;
+            } else {
+                volume = meaningless_value; // A * L, while L is not determined yet
+            }
+
+            double roughness = record->getItem("ROUGHNESS")->getRawDouble(0);
```
joakim-hove: `getSIDouble()`

opm/parser/eclipse/EclipseState/Schedule/SegmentSet.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+            double length_x;
+            double length_y;
+
+            length_x = record->getItem("LENGTH_X")->getRawDouble(0);
+            length_y = record->getItem("LENGTH_Y")->getRawDouble(0);
```
joakim-hove: `getSIDouble()`

opm/parser/eclipse/EclipseState/Schedule/SegmentSet.cpp
```diff
@@ -0,0 +1,277 @@
...
+        // create a temporary vector to store all the segments information in a sparse way
+        // then compress the vector in a orderly way to m_segments
+        std::vector<SegmentPtr> new_segments;
+        new_segments.resize(nsegmx);
+
+        for (int i = 0; i < int(new_segments.size()); ++i) {
+            new_segments[i] = NULL;
+        }
+
+        const double meaningless_value = -1.e100; // meaningless value to indicate unspecified values
```
joakim-hove: I *really* dislike such meaningless values - and I am willing to suggest quite significant rewrites to avoid it - although I do not see how just now.

opm/parser/eclipse/EclipseState/Schedule/SegmentSet.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+        m_depth_top = record1->getItem("DEPTH")->getRawDouble(0);
+        m_length_top = record1->getItem("LENGTH")->getRawDouble(0);
+        m_length_depth_type = WellSegment::LengthDepthEnumFromString(record1->getItem("INFO_TYPE")->getTrimmedString(0));
+        m_volume_top = record1->getItem("WELLBORE_VOLUME")->getRawDouble(0);
```
joakim-hove: `getSIDouble()`

opm/parser/eclipse/EclipseState/Schedule/SegmentSet.cpp
```diff
@@ -0,0 +1,277 @@
...
+        m_comp_pressure_drop = WellSegment::CompPressureDropEnumFromString(record1->getItem("PRESSURE_COMPONENTS")->getTrimmedString(0));
+        m_multiphase_model = WellSegment::MultiPhaseModelEnumFromString(record1->getItem("FLOW_MODEL")->getTrimmedString(0));
+        m_x_top = record1->getItem("TOP_X")->getRawDouble(0);
+        m_y_top = record1->getItem("TOP_Y")->getRawDouble(0);
```
joakim-hove: `getSIDouble()`

opm/parser/eclipse/EclipseState/Schedule/SegmentSet.cpp
```diff
@@ -0,0 +1,277 @@
...
+                                               m_volume_top, 0., 0., false));
+            new_segments[0] = top_segment;
+        } else if (m_length_depth_type == WellSegment::ABS) {
+            SegmentPtr top_segment(new Segment(1, 1, 0, m_length_top, m_depth_top, meaningless_value, meaningless_value,
+                                               meaningless_value, m_volume_top, m_x_top, m_y_top, true));
+            new_segments[0] = top_segment;
+        }
+
+        // read all the information out from the DECK first then process to get all the required information
+        for (int recordIndex = 1; recordIndex < int(welsegsKeyword->size()); ++recordIndex) {
```
joakim-hove: Use `size_t`

opm/parser/eclipse/EclipseState/Schedule/SegmentSet.cpp
```diff
@@ -0,0 +1,277 @@
...
+                area = M_PI * diameter * diameter / 4.0;
+            }
+
+            double segment_length;
+            double depth_change;
+            // if the values are incremental values, then we can just use the values
+            // if the values are absolute values, then we need to calculate them during the next process
+            // only the value for the last segment in the range is recorded
+            segment_length = record->getItem("SEGMENT_LENGTH")->getRawDouble(0);
+            depth_change = record->getItem("DEPTH_CHANGE")->getRawDouble(0);
```
joakim-hove: `getSIDouble()`

opm/parser/eclipse/EclipseState/Schedule/SegmentSet.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+            DeckItemConstPtr itemVolume = record->getItem("VOLUME");
+            double volume;
+            if (itemVolume->hasValue(0)) {
+                volume = itemVolume->getRawDouble(0);
```
joakim-hove: `getSIDouble()`

opm/parser/eclipse/EclipseState/Schedule/SegmentSet.cpp
GitPaean: Just to be clear. Are you suggesting to use `size_t` for for loops here?

joakim-hove: Yes
23. okt. 2015 15:55 skrev "Kai Bao" <notifications@github.com> følgende:

opm/parser/eclipse/EclipseState/Schedule/Well.cpp
```diff
@@ -303,11 +526,109 @@ namespace Opm {
         }
         return wellNameInPattern;
     }
+    bool Well::isMultiSegment() const {
+        return m_is_multi_segment;
+    }
+
+    void Well::setMultiSegment(const bool is_multi_segment) {
+        m_is_multi_segment = is_multi_segment;
```
joakim-hove: Can we infer the multisegment status of the well by just looking at the size of the segmenteset:
```C++
bool Well::isMultiSegemnt() const {
    // Do not think the DynamicState->getLast() exists - you can add it.
    if (m_segmentset->getLast()->size() > 0)
       return true;
    else
      return false;
}
```

joakim-hove:
General comment: I have recently read a bit about functional programming; in that paradigm there is strong focus on immutable data structures - and change of state is seen as the root av all evil. Now I am fully aware of the fact that we are using the non-functional language C++ here, but I still think we can learn some principles from the functional community:

- Prefer immutable to mutable.
- Isolate the mutable parts as a sub object which can be *replaced* instead of *updated*

For this PR a general comment is that the amount of state change is *high*.

GitPaean:
As requested, this PR will  be split to two or three PRs to reduce the efforts of reviewing. Closing now.

PR #619, opened 2015-10-27
Test for changing BhpLimit in history mode using weltarg

totto82:
This test does not currently work but points to an issue with the current handling of weltarg in combination with historical mode. The test is based on my interpretation of the manual and tests in Eclipse.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/tests/ScheduleTests.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/tests/ScheduleTests.cpp
@@ -1347,3 +1347,80 @@ BOOST_AUTO_TEST_CASE(createDeckWithOutOilVaporizationProperties) {
 
 
 }
+
+BOOST_AUTO_TEST_CASE(changeBhpLimitInHistoryModeWithWeltarg) {
+    Opm::Parser parser;
+    std::string input =
+            "START             -- 0 \n"
+            "19 JUN 2007 / \n"
+            "SCHEDULE\n"
+            "DATES             -- 1\n"
+            " 10  OKT 2008 / \n"
+            "/\n"
+            "WELSPECS\n"
+            "    'P'       'OP'   9   9 1*     'OIL' 1*      1*  1*   1*  1*   1*  1*  / \n"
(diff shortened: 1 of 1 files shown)


joakim-hove:
First of all - let me congratulate you with your first(?) *test first experience*; now you have a failing test - just hack away until the test passes - and wham bang you are done - never to see the bug again!

I would assume that the first `BOOST_CHECK()` statements are ok? But I agree that resetting of default BHP limit when changing between injector and producer is not implemented - at all. Is the behavior you ask for in the test - what you consider the Right&trade; way to do it?

opm/parser/eclipse/EclipseState/Schedule/tests/ScheduleTests.cpp
```diff
@@ -1347,3 +1347,80 @@ BOOST_AUTO_TEST_CASE(createDeckWithOutOilVaporizationProperties) {
...
+    DeckPtr deck = parser.parseString(input, parseMode);
+    std::shared_ptr<const EclipseGrid> grid = std::make_shared<const EclipseGrid>(10, 10, 10);
+    IOConfigPtr ioConfig;
+    Schedule schedule(parseMode , grid, deck, ioConfig);
+    WellPtr well_p = schedule.getWell("P");
+
+    BOOST_CHECK_EQUAL(well_p->getProductionProperties(0).BHPLimit, 0); //start
+    BOOST_CHECK_EQUAL(well_p->getProductionProperties(1).BHPLimit, 50 * 1e5); // 1
+    // The BHP limit should not be effected by WCONHIST
+    BOOST_CHECK_EQUAL(well_p->getProductionProperties(2).BHPLimit, 50 * 1e5); // 2
```
totto82: This fails

totto82:
> Is the behavior you ask for in the test - what you consider the Right™ way to do it?
Yes. 

I believe that the flowing changes will fix the two failures
1) Don't reset the bhp limit every WCONHIST
2) Reset the bhp limit when moving between producer and injector.

joakim-hove:
See: https://github.com/OPM/opm-parser/pull/621

PR #623, opened 2015-10-30
fix build when using newish Boost libraries

andlaus:
it seems that recent Boost versions (I used 1.59) enforce a semicolon
behind the BOOST_CHECK macros, while older ones did not.


joakim-hove:
Well a ';' has seldom hurt - thanks.

PR #624, opened 2015-11-02
Geo mini deck

joakim-hove:
Store small per-timestep 'minideck' for Schedule modifiers. With this PR the parser part of support for MULTFLT in Schedule section should be complete.

### DIFF
--- a/opm/parser/eclipse/EclipseState/EclipseState.cpp
+++ b/opm/parser/eclipse/EclipseState/EclipseState.cpp
@@ -1361,4 +1361,46 @@ namespace Opm {
         }
     }
 
+
+
+
+    void EclipseState::applyModifierDeck( std::shared_ptr<const Deck> deck) {
+        using namespace ParserKeywords;
+        for (const auto& keyword : *deck) {
+
+            if (keyword->isKeyword<MULTFLT>()) {
+                for (const auto& record : *keyword) {
+                    const std::string& faultName = record->getItem<MULTFLT::fault>()->getString(0);
+                    auto fault = m_faults->getFault( faultName );
+                    double tmpMultFlt = record->getItem<MULTFLT::factor>()->getRawDouble(0);
+                    double oldMultFlt = fault->getTransMult( );
+                    double newMultFlt = oldMultFlt * tmpMultFlt;
+
+                    /*
+                      This extremely contrived way of doing it is because of difference in
+                      behavior and section awareness between the Fault object and the
+                      Transmult object:
+
+                      1. MULTFLT keywords found in the SCHEDULE section should apply the
+                         transmissibility modifiers cumulatively - i.e. the current
+                         transmissibility across the fault should be *multiplied* with the
+                         newly entered MULTFLT value, and the resulting transmissibility
+                         multplier for this fault should be the product of the newly
+                         entered value and the current value.
+
+                      2. The TransMult::applyMULTFLT() implementation will *multiply* the
+                         transmissibility across a face with the value in the fault
+                         object. Hence the current value has already been multiplied in;
+                         we therefor first *set* the MULTFLT value to the new value, then
+                         apply it to the TransMult object and then eventually update the
+                         MULTFLT value in the fault instance.
+
+                    */
+                    fault->setTransMult( tmpMultFlt );
+                    m_transMult->applyMULTFLT( fault );
+                    fault->setTransMult( newMultFlt );
+                }
+            }
+        }
+    }
 }
--- a/opm/parser/eclipse/IntegrationTests/TransMultIntegrationTests.cpp
+++ b/opm/parser/eclipse/IntegrationTests/TransMultIntegrationTests.cpp
@@ -1,0 +1,59 @@
+/*
+  Copyright 2013 Statoil ASA.
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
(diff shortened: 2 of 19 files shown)


opm/parser/eclipse/EclipseState/EclipseState.cpp
```diff
@@ -1361,4 +1361,46 @@ namespace Opm {
...
+
+                      1. MULTFLT keywords found in the SCHEDULE section should apply the
+                         transmissibility modifiers cumulatively - i.e. the current
+                         transmissibility across the fault should be *multiplied* with the
+                         newly entered MULTFLT value, and the resulting transmissibility
+                         multplier for this fault should be the product of the newly
+                         entered value and the current value.
+
+                      2. The TransMult::applyMULTFLT() implementation will *multiply* the
+                         transmissibility across a phase with the value in the fault
```
atgeirr: Is phase a typo for face? I appreciate the comment!

opm/parser/eclipse/IntegrationTests/TransMultIntegrationTests.cpp
```diff
@@ -0,0 +1,58 @@
...
+    ParserPtr parser(new Parser());
+    boost::filesystem::path scheduleFile("testdata/integration_tests/TRANS/Deck1");
+    DeckPtr deck =  parser->parseFile(scheduleFile.string(), parseMode);
+    std::shared_ptr<EclipseState> eclState = std::make_shared<EclipseState>( deck , parseMode );
+    std::shared_ptr<const TransMult> trans = eclState->getTransMult();
+    std::shared_ptr<const Schedule> schedule = eclState->getSchedule();
+    const Events& events = schedule->getEvents();
+
+    BOOST_CHECK_EQUAL( 0.10 , trans->getMultiplier( 3,2,0,FaceDir::XPlus ));
+    BOOST_CHECK_EQUAL( 0.10 , trans->getMultiplier( 2,2,0,FaceDir::XPlus ));
```
atgeirr: The code below seems to illustrate how this feature should be used, and it makes me wonder: would it not be better if the application of modifiers happened inside the EclipseState? Then we would have to change some APIs: there would no longer be a getTransMult() method but a getTransMult(int report_step) method for example. One would still need to check the events to find out if an update is needed (using getTransMult(0) would be identical to current behaviour).

opm/parser/eclipse/EclipseState/EclipseState.cpp
joakim-hove: > Is phase a typo for face?

Yes.

opm/parser/eclipse/IntegrationTests/TransMultIntegrationTests.cpp
joakim-hove: > would it not be better if the application of modifiers happened inside the EclipseState?

Yes - that is certainly possible. The main point as I see it is that calling scope, e.g. `flow` must be responsible for checking for events - and then ask the EclipseState to recalculate. Either as here by calling an `apply()`method or asking for a new version of the `TransMult` field - I guess the latter is cleaner.

joakim-hove:
Observe that the commit 5df8926 illustrates what is possible; it is not clear to me what is the 'best' solution here?!

opm/parser/eclipse/IntegrationTests/TransMultIntegrationTests.cpp
blattms: On Tue, Nov 03, 2015 at 01;2802;0c3:19:25AM -0800, Joakim Hove wrote:
> 
> > would it not be better if the application of modifiers happened inside the EclipseState?
> 
> Yes - that is certainly possible. The main point as I see it is that calling scope, e.g. `flow` must be responsible for checking for events - and then ask the EclipseState to recalculate. Either as here by calling an `apply()`method or asking for a new version of the `TransMult` field - I guess the latter is cleaner.
> 

Currently the eclipse parser works on the whole grid while flow might
only use a part of it (e.g. in a parallel run). If flow calls to the
eclipse parser to recompute something, then this should not need to
happen on the whole grid. Otherwise it does not scale.

joakim-hove:
> If flow calls to the eclipse parser to recompute something, then this should not need to
happen on the whole grid. Otherwise it does not scale.

Well - I really think this a quite uncommon feature; our Model 2 is the first time ever I see geology modifiers in the Schedule section. My suggestion is therefor that we allow *one* 'stop-the-world' to recalculate the transmissibilities on one PE and distribute to the remaining PE's during the execution of Model 2.

joakim-hove:
> Then we would have to change some APIs: there would no longer be a getTransMult() method but a getTransMult(int report_step) method for example.

I have thought more about this; and would actually like to retain the current approach with downstream modules calling `EclipseState::applyModifierDeck()` instead of `getTrasnmult( reportStep )`. The reasons are:

1. Using the method signature `getTransMult( reportStep )` analogous to e.g. the wellrates hints that we have full 'random access' to the Transmissibility multtiplier in the time domain. That is not the case - the current `EclipseState::applyModifierDeck()` is a destructive operation - i.e. you can not easily get back to the initial transmissibility multiplier. Since I believe this (geo modifiers in the Schedule section) is quite close to a measure zero event I *really* think the destructive approach is OK.

2. If other geo modifiers than pure transmissibility modifiers come in the Schedule section (do not know if that is possible) - that will still fit in the `EclipseState::applyModifierDeck()`.

PS: I realize I am arguing from two persepctives when it comes to frequency of this phenomenon.

atgeirr:
> I have thought more about this; and would actually like to retain the current approach

I can certainly accept your arguments. I would think that it would not be very hard to support non-destructiveness of this though. The Schedule class does this for well controls for example, so I would think that using the same utilities that make this doable for Schedule items would make it doable also for grid properties? But if this means a lot more work, I agree that we could let the (low) frequency win...

joakim-hove:
> The Schedule class does this for well controls for example, so I would think that using the same utilities that make this doable for Schedule items would make it doable also for grid properties?

Hehe - you know how to get your way:grin: I certainly agree with you what would be most elegant; I'll let it linger a bit.

joakim-hove:
Mainly due to the perceived frequency argument I will merge this in the current state. If / when we see the need for full random access in the time direction we will expand on this.

PR #625, opened 2015-11-02
Support for the keyword WELSEGS

GitPaean:
which is split from the PR #615 manually. 

Many of the comments PR #615 have been addressed, while not all of them yet. 

I hope the new PR is easier to review.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
@@ -1267,7 +1267,7 @@ namespace Opm {
         SegmentSetPtr newSegmentset= std::make_shared<SegmentSet>();
         newSegmentset->segmentsFromWELSEGSKeyword(keyword);
 
-        std::string well_name = newSegmentset->wellName();
+        const std::string& well_name = newSegmentset->wellName();
         WellPtr well = getWell(well_name);
 
         // update multi-segment related information for the well
--- a/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
@@ -1271,6 +1271,7 @@ namespace Opm {
         WellPtr well = getWell(well_name);
 
         // update multi-segment related information for the well
+        well->addSegmentSet(currentStep, newSegmentset);
     }
 
     void Schedule::handleWGRUPCON(DeckKeywordConstPtr keyword, size_t currentStep) {
--- a/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
@@ -157,6 +157,9 @@ namespace Opm {
             if (keyword->name() == "COMPDAT")
                 handleCOMPDAT(keyword, currentStep);
 
+            if (keyword->name() == "WELSEGS")
+                handleWELSEGS(keyword, currentStep);
+
             if (keyword->name() == "WELOPEN")
                 handleWELOPEN(keyword, currentStep , section->hasKeyword("COMPLUMP"));
 
@@ -1258,6 +1261,16 @@ namespace Opm {
             well->addCompletions(currentStep, iter->second);
         }
         m_events->addEvent(ScheduleEvents::COMPLETION_CHANGE, currentStep);
+    }
+
+    void Schedule::handleWELSEGS(DeckKeywordConstPtr keyword, size_t currentStep) {
+        SegmentSetPtr newSegmentset= std::make_shared<SegmentSet>();
+        newSegmentset->segmentsFromWELSEGSKeyword(keyword);
+
+        std::string well_name = newSegmentset->wellName();
+        WellPtr well = getWell(well_name);
+
+        // update multi-segment related information for the well
     }
 
     void Schedule::handleWGRUPCON(DeckKeywordConstPtr keyword, size_t currentStep) {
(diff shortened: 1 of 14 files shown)


joakim-hove:
This looks much better from the point of a reviewer - thank you!

joakim-hove:
>  It is just more clear to me to give an explicit flag

I agree explicit is good - but in this case the price is to maintain two variables with the same content. So I would prefer it were changed; it could get in addition get a `reportStep` argument - then you will for free get change with time as well.

GitPaean:
Okay. I will fix it. 

> then you will for free get change with time as well.

You mean `isMultiSegment(size_t time_step)`?

joakim-hove:
Yes
2. nov. 2015 16:15 skrev "Kai Bao" <notifications@github.com> følgende:

GitPaean:
fixed `isMultiSegment()` function. Thanks.

joakim-hove:
> Even m_length_x and m_length_y are not useful for simulation at all. They are for plotting purpose.

Then I strongly suggest they are deleted - plotting support is way into the future.

joakim-hove:
How about `m_volume, m_length` and `m_area`? I assume they are related by:
```C++
m_volume = m_area * m_length
```
would it be possibel to only have two of them as members?

joakim-hove:
> The problem with WELSEGS is that under some circumstances (unless specifying all the required information completely), the required information can not be obtained by one single record. We have to have all the segment information related to the well first (existing and newly entered ones), then do some processing to make the required information complete.

Yes - I understand. The `SegmentRange`class was just a wild idea - the point was that if possible it would be beneficial to do the special case processing a 'special place' - so that the final `Segments / SegementSet` can be immutable.

GitPaean:
> How about m_volume, m_length and m_area? I assume they are related by:

>> m_volume = m_area * m_length

> would it be possibel to only have two of them as members?

I prefer not. 
1. The relation only applies when they are defaulted. They can be specified with arbitrary values and the relation may not work there. 
2. Even the relation applies always, the three values will be used in calculation in different parts later in the simulation, literally. We do not want to calculate them again later.

GitPaean:
>> Even m_length_x and m_length_y are not useful for simulation at all. They are for plotting purpose.

> Then I strongly suggest they are deleted - plotting support is way into the future.

Done deleting all the length_x and length_y related. I wanted to keep them since I feel they are something we might need later. Anyway, I think they will not be difficult to add again, they are following the exactly same strategy with length and depth.

joakim-hove:
>1. The relation only applies when they are defaulted. They can be specified with arbitrary values and the relation may not work there. 

Well - if the simulator actually allows violation of `Volume = Length * Area` we must of course keep all three.

> 2. Even the relation applies always, the three values will be used in calculation in different parts later in the simulation, literally. We do not want to calculate them again later.

Your argument 1) wins - but this argument carries no weight; if one of the fields were removed we would still retain all three `getXxx()` methods - and the recalculation of one of the properties would be free for all practical purposes.

GitPaean:
>> The problem with WELSEGS is that under some circumstances (unless specifying all the required information completely), the required information can not be obtained by one single record. We have to have all the segment information related to the well first (existing and newly entered ones), then do some processing to make the required information complete.

> Yes - I understand. The SegmentRangeclass was just a wild idea - the point was that if possible it would be beneficial to do the special case processing a 'special place' - so that the final Segments / SegementSet can be immutable.

I understand the concern. The only approach to update something is then always making a copy, updating it and putting it back. 

But considering, we need to update the completion information for the segments soon later, and we probably need to update the segment information for the completions. And then we need to re-entering a few segment information with different `INC` or `ABS` way to update the existing segment structures. A lot of copying operations need to be done.  It is just something I am thinking at the moment.

joakim-hove:
> It is just something I am thinking at the moment.

Let us continue the discussion tomorrow.

GitPaean:
> Yes - I understand. The SegmentRangeclass was just a wild idea - the point was that if possible it would be beneficial to do the special case processing a 'special place' - so that the final Segments / SegementSet can be immutable.

Please help to check if the current status is closer to what you want. Thanks.

GitPaean:
done updating. Thanks for all the comments.

GitPaean:
Done re-basing to solve the conflict with the updated master branch.

joakim-hove:
Thanks for rebase!

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -1260,6 +1263,17 @@ namespace Opm {
         m_events->addEvent(ScheduleEvents::COMPLETION_CHANGE, currentStep);
     }
 
+    void Schedule::handleWELSEGS(DeckKeywordConstPtr keyword, size_t currentStep) {
+        SegmentSetPtr newSegmentset= std::make_shared<SegmentSet>();
+        newSegmentset->segmentsFromWELSEGSKeyword(keyword);
+
+        std::string well_name = newSegmentset->wellName();
```
joakim-hove: `const std::string& well_name`

joakim-hove:
OK - I am ready to merge this now - but before I merge: 

```txt:
What is the public api of these changes; how do you intend to get 
segment properties into downstream modules?
```

GitPaean:
> What is the public api of these changes; how do you intend to get segment properties into downstream modules?

It will be used in the autodiff module through `Well::getSegmentSet(size_t time_step)` to initialize the `WellMultiSegment` related information.

joakim-hove:
> It will be used in the autodiff module through Well::getSegmentSet(size_t time_step) to initialize the WellMultiSegment related information.

OK - thank you.

PR #627, opened 2015-11-05
Prevent wells with zero rate and crossflow disabled to open

totto82:
This mimics behaviour observed from Eclipse results on the new model.


joakim-hove:
Can you factor the check out in å Well::canOpen(step) method

totto82:
Well::canOpen(step) and a test is added.

joakim-hove:
Good - thank you!

PR #630, opened 2015-11-10
Rename from HO to H0

totto82:
It is hard to judge from the manual, but H0 and not HO is used in the new model.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/ScheduleEnums.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/ScheduleEnums.cpp
@@ -653,8 +653,8 @@ namespace Opm {
 
         const std::string MultiPhaseModelEnumToString(MultiPhaseModelEnum enumValue) {
             switch (enumValue) {
-                case HO:
-                    return "HO";
+                case H0:
+                    return "H0";
                 case DF:
                     return "DF";
                 default:
@@ -667,8 +667,8 @@ namespace Opm {
             std::string trimmedString(stringValue);
             boost::algorithm::trim(trimmedString);
 
-            if (trimmedString == "HO") {
-                return HO;
+            if (trimmedString == "H0") {
+                return H0;
             } else if (trimmedString == "DF") {
                 return DF;
             } else {
--- a/opm/parser/eclipse/EclipseState/Schedule/ScheduleEnums.hpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/ScheduleEnums.hpp
@@ -264,7 +264,7 @@ namespace Opm {
         CompPressureDropEnum CompPressureDropEnumFromString(const std::string& stringValue);
 
         enum MultiPhaseModelEnum {
-            HO = 0,
+            H0 = 0,
             DF = 1
         };
         const std::string MultiPhaseModelEnumToString(MultiPhaseModelEnum enumValue);


GitPaean:
We saw that from the model file some while ago and reported. I prefer it is something wrong with the new model. `HO` is the short for `homogeneous`. And if you check in computer in the manual file, it is `HO` not `H0`. `H0` is not a very meaningful name to me.

totto82:
Eclipse runs the new model without complaining. Maybe this is a common mix-up and Eclipse reads both of them.

GitPaean:
It is possible. It is also possibly because `HO` is the default option. 
My examples with `HO` did not get complaining either. You can try with `HO`. 

In the report file from the new model, it is `HO` even in the deck it is `H0`.

```
------------------------------------------------------------------------------------------------------------------------------------
:WELLNAME: SEG : BRN : MAIN  :       : SEGMENT :TOT LENGTH:  DEPTH   :T.V. DEPTH:DIA OR F:VFP TAB OR:   AREA  :  VOLUME :  P DROP  :
:  AND   : NO  : NO  : INLET :OUTLET : LENGTH  :  TO END  :  CHANGE  :  AT END  :SCALING :ABS ROUGHN: X-SECTN :         :   MULT   :
:SEG TYPE:     :     :SEGMENT:SEGMENT: METRES  :  METRES  :  METRES  :  METRES  : METRES :  METRES  :   M**2  :    M3   :  FACTOR  :
------------------------------------------------------------------------------------------------------------------------------------
:F-29Y1Y2:   1 :   1 :    2  :    0  :  3748.0 :   3748.0 :   2798.0 :   2798.0 :      0 :        0 :       0 :   0.000 :  1.00E+00:
: HFA  HO:   2 :     :    3  :    1  :    24.9 :   3772.9 :      4.5 :   2802.5 : 0.1550 : 0.000100 : 0.01887 :   0.470 :  1.00E+00:
:      HO:   3 :     :    4  :    2  :   650.8 :   4423.7 :     21.3 :   2823.8 : 0.1550 : 0.000100 : 0.01887 :  12.279 :  1.00E+00:
:      HO:   4 :     :    5  :    3  :    16.6 :   4440.2 :      0.9 :   2824.7 : 0.1550 : 0.000100 : 0.01887 :   0.313 :  1.00E+00:
:      HO:   5 :     :    6  :    4  :     6.8 :   4447.0 :      0.3 :   2825.0 : 0.1550 : 0.000100 : 0.01887 :   0.129 :  1.00E+00:
:      HO:   6 :     :    7  :    5  :    45.8 :   4492.8 :      1.8 :   2826.8 : 0.1550 : 0.000100 : 0.01887 :   0.864 :  1.00E+00:
:      HO:   7 :     :    8  :    6  :    61.0 :   4553.8 :      2.0 :   2828.8 : 0.1550 : 0.000100 : 0.01887 :   1.152 :  1.00E+00:
```

totto82:
@GitPaean What happens if you use some random enum like AA

I do not have any strong opinion on how this should be solved. I am happy as long as the new model runs. So either changing the deck or changing the parser to use HO as long as the input is not DF.

GitPaean:
I do not know how Eclipse will handle that. I guess probably in Eclipse the default option will be used if a non-recognized keyword is input there based on how it handles `H0`. 

For OPM, the current implementation if the keyword is not recognized, it will throw an exception. We can handle with the similar way with Eclipse though.

GitPaean:
How about we accept both `HO` or `H0` for `homegenous flow model`? `H0` is basically only for new model and some other typos.  We do not accept other random input for `homegenous flow model`. 

I do not have strong opinion either, I just think `HO` is a more meaningful name.

joakim-hove:
I guess we will have to accept H0 and HO - right now a test fails :angry:

joakim-hove:
> I am happy as long as the new model runs

That is not enough; if we are to achieve anything close to long term stability and maintainability of the opm codebase we must look a bit beyond the ECLIPSE deck of the day.

PR #632, opened 2015-11-11
handling both HO and H0 for mutliphase model input

GitPaean:
To fix a failed test.

Internally and output we still use `HO`.


joakim-hove:
Thank you - merging.

PR #631, opened 2015-11-10
Compsegs has dimension

joakim-hove:

PR #633, opened 2015-11-11
Support of keyword COMPSEGS

GitPaean:
This PR needs the PR #631. 

The support of keyword COMPSEGS is supposed to handle the current needs of the keyword.

limitations for now:
1.  The input with a range is not supported yet.
2.  More than one completion for a segment is not supported yet.
3.  Calculating completion depth based on COMPDAT is not supported yet.
4.  Not handling the thermal length.


opm/parser/share/keywords/000_Eclipse100/C/COMPSEGS
```diff
@@ -9,6 +9,6 @@
  {"name" : "DIRECTION" , "value_type" : "STRING"},
  {"name" : "END_IJK" , "value_type" : "INT"},
  {"name" : "CENTER_DEPTH" , "value_type" : "DOUBLE" , "dimension" : "Length" , "default" : 0},
- {"name" : "THERMAL_LENGTH" , "value_type" : "DOUBLE" , "dimension" : "Length" , "default" : 0},
+ {"name" : "THERMAL_LENGTH" , "value_type" : "DOUBLE" , "dimension" : "Length"},
  {"name" : "SEGMENT_NUMBER" , "value_type" : "INT"}
- ]]}
```
GitPaean: I have no idea about how things work here. Is it something related to the editor used?

joakim-hove: > I have no idea about how things work here. Is it something related to the editor used?

Yes probably - what has happened is that the line: `]]}` is replaced with `]]}\n` - i.e. with a final newline - that is OK!

opm/parser/eclipse/EclipseState/Schedule/Completion.hpp
```diff
@@ -71,6 +78,10 @@ namespace Opm {
         WellCompletion::DirectionEnum m_direction;
         Value<double> getDiameterAsValueObject() const;
         Value<double> getSkinFactorAsValueObject() const;
+	// related segment number
+	// -1 means the completion is not related to segment
```
joakim-hove: I want a `bool attachedToSegment()` method

opm/parser/eclipse/EclipseState/Schedule/Completion.cpp
```diff
@@ -212,6 +242,22 @@ namespace Opm {
         return m_wellPi;
     }
 
+    int Completion::getSegmentNumber() const {
+        return m_segment_number;
```
joakim-hove: This should throw an exception if `m_segment_number == -1` -  if calling scope needs to check the (not yet created) `attachedToSegment()` method should be consulted.

opm/parser/eclipse/EclipseState/Schedule/Completion.hpp
```diff
@@ -55,6 +58,10 @@ namespace Opm {
         double getDiameter() const;
         double getSkinFactor() const;
         void   fixDefaultIJ(int wellHeadI , int wellHeadJ);
+        int getSegmentNumber() const;
+        void setSegmentNumber(const int segment_number);
+        double getCenterDepth() const;
+        void setCenterDepth(const double center_depth);
```
joakim-hove: I think the two methods `setSegmentNumber()` and `setCenterDepth()` could be combined to one method:
```C++
void attachSegment( int segmentNumber , double centerDepth) {
    m_segmentNumber = segmentNumber;
    m_center_depth = centerDepth;
}
```

opm/parser/eclipse/EclipseState/Schedule/Completion.cpp
GitPaean: If we throw an exception here, it will make the following situation a little more complicated. 
```
m_segment_number(oldCompletion->getSegmentNumber())
```
And `-1` is unusable as a segment number. How about we just return `-1` for completions of non-multi-segment wells and completions of multi-segment wells before finishing the attaching process.

GitPaean: Never mind. Done addressing the comments. Please check.

opm/parser/eclipse/EclipseState/Schedule/Compsegs.hpp
```diff
@@ -0,0 +1,71 @@
...
+#include <opm/parser/eclipse/EclipseState/Schedule/SegmentSet.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/CompletionSet.hpp>
+
+namespace Opm {
+
+    struct Compsegs {
+        int m_i;
+        int m_j;
+        int m_k;
+        int m_branch_number;
```
joakim-hove: Maybe a comment that the `branch_number` has offset 1 - i.e. the same as Eclipse.

GitPaean: Okay. I will add some comments here. Basically, the top segment will always be segment 1. The other segments begins with 2.

opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
```diff
@@ -0,0 +1,206 @@
...
+            }
+
+            WellCompletion::DirectionEnum direction;
+            if (record->getItem("DIRECTION")->hasValue(0)) {
+                direction = WellCompletion::DirectionEnumFromString(record->getItem("DIRECTION")->getString(0));
+            } else if (!record->getItem("DISTANCE_END")->hasValue(0)) {
+                throw std::runtime_error("the direction has to be specified when DISTANCE_END in the record is not specified");
+            }
+
+            int end_IJK;
```
joakim-hove: No need to properly internalize the `END_IJK`value; just check if it is set or not:
```C++
if (!record->getItem<ParserKeywords::COMPSEGS::END_IJK>()->hasValue(0)) {
    CompsegsPtr new_compsegs = std::make_shared<Compsegs>(I, J, K, branch, distance_start, distance_end,
                                                                                                         direction, center_depth, segment_number);
    compsegs.push_back(new_compsegs);
} else { // a range is defined. genrate a range of Compsegs
                throw std::runtime_error("entering COMPSEGS entries with a range is not supported yet!");
}
```

GitPaean:
Do I need to add a new header file to use something like `ParserKeywords::COMPSEGS::I`?

GitPaean:
The compiling is ok. I got problem when running. 
```
terminate called after throwing an instance of 'std::invalid_argument'
 what():  Itemname: ParserKeywords::COMPSEGS::I does not exist.
```

GitPaean:
I saw the problem now. Sorry for the spamming.

opm/parser/eclipse/EclipseState/Schedule/Compsegs.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+    void Compsegs::updateCompletionsWithSegment(const std::vector<CompsegsPtr>& compsegs,
+                                                CompletionSetPtr completion_set) {
+
+        for (size_t i_comp = 0; i_comp < compsegs.size(); ++i_comp) {
+            const int i = compsegs[i_comp]->m_i;
+            const int j = compsegs[i_comp]->m_j;
+            const int k = compsegs[i_comp]->m_k;
+
+            size_t ic;
```
joakim-hove: Rather implement a new method `CompletionSet::getFromIJK()`

joakim-hove:
OK - I am ready to merge this; but first I would like the `COMPSEGS` and `WELSEGS` specific implementation to be moved to a new directory `EclipseState/Schedule/MSW`. Hopefully that will contribute to readability++

GitPaean:
done updating. Thanks for all the comments.

PR #635, opened 2015-11-13
Support for minpv-pinch-nnc 

totto82:
Adds 
- default constructor
- non-const access of the nncs
- test

### DIFF
--- a/opm/parser/eclipse/EclipseState/Grid/NNC.cpp
+++ b/opm/parser/eclipse/EclipseState/Grid/NNC.cpp
@@ -50,6 +50,11 @@ namespace Opm
         }
     }
 
+    // default constructor
+    NNC::NNC() {
+    }
+
+
     void NNC::addNNC(const size_t NNC1, const size_t NNC2, const double trans) {
         m_nnc1.push_back(NNC1);
         m_nnc2.push_back(NNC2);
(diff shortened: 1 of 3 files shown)


joakim-hove:
> non-const access of the nncs

Hmmm - why do need that?

> When this code is revisited I would strongly prefer that it was implemented as a vector of:
struct NNC {
    size_t cell1;
    size_t cell2;
    double trans;
}
```

Instead of three `std::vector<>` instances.

atgeirr:
> When this code is revisited I would strongly prefer that it was implemented as a vector of:
> 
> ```
> struct NNC {
>     size_t cell1;
>     size_t cell2;
>     double trans;
> }
> ```
> Instead of three std::vector<> instances.

I think that where the class is used (in opm-autodiff) it is more convenient to have the separate vectors, at least for the transmissibilities. I do agree that the structure you propose is a little more intuitive though.

totto82:
The nnc data is stored in  a structure, NNCdata, as you suggested.

joakim-hove:
> The nnc data is stored in a structure, NNCdata, as you suggested.

Thank you. I am ready to merge this - but I will hold back until someone gives a green light for the downstream module. The downstream reviewer @atgeirr / @bska ?? is welcome to push green on this as well.

atgeirr:
The whole chain of PRs should now be good, I will therefore merge this, since @joakim-hove gave his OK. Thanks for the effort!

PR #636, opened 2015-11-16
Sonar config

flikka:
Config file for sonarqube analysis - needs a runner and server in order to actually do anything.


flikka:
WE choose not to follow this sonarqube track

PR #637, opened 2015-11-24
Change return type from tuple to array for getCellCenter().

atgeirr:
This makes it easier to write code using this function (see OPM/opm-autodiff#543).

The tests needed no changes since they used `auto` for the return type and an std::array supports the std::tuple interface.


joakim-hove:
Looks fine - @atgeirr merge when downstream is ready.

atgeirr:
Actually, it can be merged immediately, since no downstream code breaks (at least as far as autodiff).

PR #642, opened 2015-12-01
Miscible keywords

totto82:
Add table support for the following keywords
MSFN
PMISC
MISC
SGCWMIS
SORWMIS

Also checkRange() is added to SimpleTable to check whether the input span [0,1]

### DIFF
--- a/opm/parser/eclipse/CMakeLists.txt
+++ b/opm/parser/eclipse/CMakeLists.txt
@@ -278,6 +278,7 @@ EclipseState/Tables/TableManager.hpp
 EclipseState/Tables/TableContainer.hpp
 EclipseState/Tables/SorwmisTable.hpp
 EclipseState/Tables/SgcwmisTable.hpp
+EclipseState/Tables/MiscTable.hpp
 #
 Utility/WconinjeWrapper.hpp
 Utility/CompdatWrapper.hpp
--- a/opm/parser/eclipse/CMakeLists.txt
+++ b/opm/parser/eclipse/CMakeLists.txt
@@ -280,6 +280,7 @@ EclipseState/Tables/SorwmisTable.hpp
 EclipseState/Tables/SgcwmisTable.hpp
 EclipseState/Tables/MiscTable.hpp
 EclipseState/Tables/PmiscTable.hpp
+EclipseState/Tables/MsfnTable.hpp
 #
 Utility/WconinjeWrapper.hpp
 Utility/CompdatWrapper.hpp
--- a/opm/parser/eclipse/CMakeLists.txt
+++ b/opm/parser/eclipse/CMakeLists.txt
@@ -276,6 +276,7 @@ EclipseState/Tables/VFPProdTable.hpp
 EclipseState/Tables/VFPInjTable.hpp
 EclipseState/Tables/TableManager.hpp
 EclipseState/Tables/TableContainer.hpp
+EclipseState/Tables/SorwmisTable.hpp
 #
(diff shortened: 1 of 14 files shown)


joakim-hove:
Good - tests and all!

opm/parser/eclipse/CMakeLists.txt
```diff
@@ -276,6 +276,11 @@ EclipseState/Tables/VFPProdTable.hpp
 EclipseState/Tables/VFPInjTable.hpp
 EclipseState/Tables/TableManager.hpp
 EclipseState/Tables/TableContainer.hpp
+EclipseState/Tables/SorwmisTable.hpp
```
joakim-hove: The new header `SorwmisTable.hpp` is not part of the PR; so right now it does not compile.

joakim-hove:
Actually none of the new headers are added to the PR.

opm/parser/eclipse/EclipseState/Tables/SimpleTable.cpp
```diff
@@ -153,6 +153,16 @@ void SimpleTable::checkMonotonic(const std::string& columnName,
     }
 }
 
+void SimpleTable::checkRange(const std::string& columnName)
```
joakim-hove: If you indeed intend to throw I would prefer if the method were named e.g. `assertRange()` - it should be safe to call a method named `checkRange()` without catching fire.

joakim-hove: The range `[0,1]` is not *that* canonical; rename the method `assertUnitRange()` - or alternatively take limits as input arguments to the method.

opm/parser/eclipse/EclipseState/Tables/TableManager.cpp
```diff
@@ -145,6 +146,21 @@ namespace Opm {
         addTables( "RVVD", m_eqldims->getNumEquilRegions());
 
         {
+            size_t numMiscibleTables = ParserKeywords::MISCIBLE::NTMISC::defaultValue;
+            if (deck.hasKeyword<ParserKeywords::MISCIBLE>()) {
+                auto keyword = deck.getKeyword<ParserKeywords::MISCIBLE>();
+                auto record = keyword->getRecord(0);
+                numMiscibleTables =  static_cast<size_t>(record->getItem<ParserKeywords::MISCIBLE::NTMISC>()->getInt(0));
+            }
```
joakim-hove: I'm glad you are using the `ParserKeywords` classes - I hope it feels reasonably natural!

opm/parser/eclipse/IntegrationTests/ParseMiscible.cpp
```diff
@@ -0,0 +1,262 @@
...
+std::strcpy(tooFewTables,miscibleTightData);
+std::strcat(tooFewTables,sorwmisData);
+
+// missing miscible keyword
+BOOST_CHECK_THROW (parser->parseString(sorwmisData, ParseMode()), std::invalid_argument );
+
+//too many tables
+BOOST_CHECK_THROW( parser->parseString(tooFewTables, ParseMode()), std::invalid_argument);
+
+char data[10000];
```
joakim-hove: Can't you just use `std::string`?

opm/parser/eclipse/CMakeLists.txt
totto82: Sorry. I didn't notice.

joakim-hove:
Good - running full build test now; will merge when all is green.

PR #644, opened 2015-12-08
OPM-250: Minor changes to support restart

chflo:
Added a new method getRestartFileName(const std::string& restart_base, int report_step) to IOConfig. 
The parameters are the restart_base and report_step given in the RESTART keyword in the DATA file, located in the initConfig object. 
I will be using this method from the restart functionality code in autodiff. (not yet a PR for this).

 There are some things I wonder about here though , - what about support for higher restart numbers than 9999? As you can see from the code this is not supported here, and I can't see that Eclipse supports it either, but I am sure that I am wrong. So please enlighten me on this subject. 

Or maybe the ert code has functionality for getting the name of a restart file?

### DIFF
--- a/opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
+++ b/opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
@@ -22,6 +22,9 @@
 
 #include <opm/parser/eclipse/EclipseState/IOConfig/IOConfig.hpp>
 #include <opm/parser/eclipse/Deck/Section.hpp>
+#include <ert/ecl/ecl_util.h>
+
+
 
 
 namespace Opm {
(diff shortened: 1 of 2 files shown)


joakim-hove:
> There are some things I wonder about here though , - what about support for higher restart numbers than 9999? 

You are right - according to the Eclipse manual you should use unified files if you need more than 9999 report steps - so for good or for bad we are perfectly aligned with Eclipse.

joakim-hove:
> Or maybe the ert code has functionality for getting the name of a restart file?

There is an ert function for this - `ecl_util_alloc_filename()` - so the internals of this function could be implemented as:
```C++
{
     ecl_file_enum file_type = (unified_file) ? ECL_UNIFIED_RESTART_FILE : ECL_RESTART_FILE;
     char * c_str = ecl_util_alloc_filename( NULL , restart_base.c_str() , file_type , fmt_file , report_nr);
     restart_filename = c_str;
     free( c_str );
}
``` 

Where boolean variables `unified_file` and `fmt_file` are defined based on the correct *in* or *out* variable. This is the way it is done in the rest of opm - so in a way I would prefer that; I guess we should add nice C++ wrappers to: https://github.com/Ensembles/ert/tree/master/devel/libeclxx/

PR #653, opened 2016-01-13
Basic Python wrappers for opm-parser

joakim-hove:
With this PR a basic infrastructure for Python wrapping of (some of) the classes in opm-parser are added.

**How it works:** The wrapping is based on the standard Python ctypes which is used to load and bind to a shared library at runtime. On top of that we are using a framework from ert to "prototype" the C functions in the shared library. Due to name mangling an extremely thin C layer is created around the C++ code, the Python code binds to these C functions.

There are other alternative frameworks for binding Python to C++, most notably Swig and Boost Python. The advantage with the current approach is that the Python classes are implemented in Python - not in some intermediate definition language. The approach based on ctypes and `ert.cwrap` has been heavily tested in Statoil for several years. The current wrappers have also seen some use in Statoil as an installed feature branch.

**To try it out:** Everything is controlled by the cmake switch `ENABLE_PYTHON` - that switch defaults to `OFF`. If `ENABLE_PYTHON` is set to `ON` the following will happen:
 
 1. The build system will force a shared build - the ctypes system requires shared libraries.
 2. A very thin C wrapper library `libcopmparser.so` is created.
 3. A tree of Python files is marked for installation, and a simple python "compilation" is performed.
 4. Several Python tests are added and i ntegrated into the `ctest` framework.

As long as the `ENABLE_PYTHON` setting is left at the default value `OFF` the opm-parser product will not be affected *at all*; and even when `ENABLE_PYTHON` is set to `ON` the additions will be as *new* files.

**Requirements:** This requires a brand new version of ert, no other external dependencies.

**Why on earth:** I am willing to bet there are more than 10 half-baked Eclipse input parser written in everything from Fortran to Perl in more or less active use in Statoil today. I feel that the current opm-parser implementation has the potential to "rule them all" - but then it must be simpler for the average reservoir engineer to use what is currently the case.

```Python
#!/usr/bin/env python
from opm import Parser
import sys

data_file = sys.argv[1]
parser = Parser( )
deck = parser.parseFile( data_file )

print("List of keywords in %s" % data_file)
for keyword in deck:
    print( keyword )
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,17 @@ if(NOT OPM_COMMON_ROOT)
   find_package(opm-common QUIET)
 endif()
 
+#-----------------------------------------------------------------
+option (ENABLE_PYTHON "Enable simple python wappers" OFF)
+if (ENABLE_PYTHON)
+   if (NOT BUILD_SHARED_LIBS)
+      set( BUILD_SHARED_LIBS ON)
+      message( WARNING "When ENABLE_PYTHON is ON we must use shared libraries - forcing BUILD_SHARED_LIBS to ON")
+   endif()
+endif()
+#-----------------------------------------------------------------
+
+
 if (NOT opm-common_FOUND)
    unset(opm-common_FOUND)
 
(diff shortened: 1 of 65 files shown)


opm/parser/eclipse/python/python/opm/deck/keyword.py
```diff
@@ -0,0 +1,12 @@
...
+
+class DeckKeyword(BaseCClass):
+    def __init__(self):
+        raise NotImplementedError("Class can not be instantiated directly!")
+
+    
+
+CWrapper.registerObjectType("deck_keyword", DeckKeyword)
+
+cwrapper = CWrapper(OPM_LIB)
```
jepebe: Old style class?

andlaus:
cool feature IMO. based on your PR description: does this work with alternative operating systems which may use different mangling conventions? I'm basically thinking on MacOS X and to a lesser extend Windows...

akva2:
hence the c wrapper. c is mangled the same on all platforms (no mangling - no need to encode parameter types in function names). calling convention is the only thing to watch out for (on windows).

PR #654, opened 2016-01-18
Internalized GEFAC keyword, added support for WCONPROD controlmode option GRUP

chflo:
- Added and internalized GEFAC keyword
- Added support for WCONPROD controlmode option "GRUP"

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Group.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Group.cpp
@@ -24,6 +24,7 @@
 #include <opm/parser/eclipse/EclipseState/Schedule/WellSet.hpp>
 
 #define INVALID_GROUP_RATE -999e100
+#define INVALID_EFFICIENCY_FACTOR 0.0
 
 
 namespace Opm {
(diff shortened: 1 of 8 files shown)


joakim-hove:
OK; good this is essentially ready to be merged. But as part of this PR I would like you to fix up the old names `controlModes` here:https://github.com/chflo/opm-parser/blob/3e498781439037cb4be868c94503952ecbdf3e66/opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.cpp#L86 and https://github.com/chflo/opm-parser/blob/3e498781439037cb4be868c94503952ecbdf3e66/opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.cpp#L33. These `std::vector<std::string>` instances are not really *controlModes* - but rather *targetRates*. And around line ~33 the same `defaultApplied()` test as further down should be implemented.

bska:
> really *controlModes* - but rather *targetRates*

That's still slightly misleading, though.  Pressures (`BHP`, and `THP`) aren't rates.  I chose the name `controlModes` because the subsequent code checks whether or not the well actually supports being controlled by a particular type of constraints.

joakim-hove:
> That's still slightly misleading, though

OK - I agree. How about:
```C++
for (auto const& modeKey : {"BHP" , "LRAT" , "GRAT", ...}) {
     if (!record->getItem(modeKey)->defaultApplied(0)) {
         auto cmode = WellProducer::ControlModeFromString( modeKey );
         p.addProductionControl( cmode );
    }
}
```

bska:
>> That's still slightly misleading, though
>
> OK - I agree. How about [code snipped]

Perfect!  That's an excellent suggestion.

joakim-hove:
Good - merging :-)

PR #656, opened 2016-01-19
Build system improvements & dependency/header cleanup

jokva:
Hi,

These patches introduce some modifications to the build system to reduce time spent compiling by cleaning up unused or too-heavy headers. Notable changes:

* Generated source code has been tweaked for build performance, but are indistinguishable for users and developers alike (except for reduced build times).
* You no longer need to include all keyword classes - rather, they are offered on a first-letter-basis, now in Parser/ParserKeywords/A.hpp etc. For compability and occasional convenience, the old ParserKeywords.hpp still includes all keyword classes.
* Headers have been tuned slightly by only including other headers that -need- inclusion (typically because of value members). This  speeds up compilation and makes definition dependencies more obvious, but introduces some soft breakage - if you previously relied on some header to include everything you use for you, your code won't compile. This will be fixed by simply including the correct headers.

Downstream PRs:
* opm-core: https://github.com/OPM/opm-core/pull/944
* opm-material: https://github.com/OPM/opm-material/pull/111
* opm-porsol: https://github.com/OPM/opm-porsol/pull/141
* opm-autodiff: https://github.com/OPM/opm-autodiff/pull/573


joakim-hove:
This is good! I will merge as soon as the small details have been updated.

joakim-hove:
Get ready for high-speed parser builds!

PR #658, opened 2016-01-22
Fixes autogen'd .hpp installed into wrong dir

jokva:
Pretty self-explanatory. The makefile has a bug which installs autogenerated headers into the wrong directory.

### DIFF
--- a/opm/parser/eclipse/CMakeLists.txt
+++ b/opm/parser/eclipse/CMakeLists.txt
@@ -364,12 +364,11 @@ include( ${PROJECT_SOURCE_DIR}/cmake/Modules/install_headers.cmake )
 install_headers( "${HEADER_FILES}" "${CMAKE_INSTALL_PREFIX}" )
 install( TARGETS opmparser DESTINATION ${CMAKE_INSTALL_LIBDIR} )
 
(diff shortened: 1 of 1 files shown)


jokva:
This is not yet completely verified, so it's not ready for merge.

joakim-hove:
Hmmm - opm-autodiff still fails:

```txt
/private/jenkins/jenkins/OPM/install/RH6/include/opm/parser/eclipse/EclipseState/Tables/PlyshlogTable.hpp:22:58: fatal error: opm/parser/eclipse/Parser/ParserKeywords/P.hpp: No such file or directory
 #include <opm/parser/eclipse/Parser/ParserKeywords/P.hpp>
 ```

jokva:
Yeah, autodiff failes, but I suspect this is a jenkins issue -or- some peculiar race condition. I fetched a local copy and tried make install, and the headers were put in the right directory (with this patch).

joakim-hove:
> some peculiar race condition.

I have always found the more custom dependency management in cmake quite incomprehensible, but I think race condition explains it. To me it actually seems like the `GLOB` operation is performed at configure time - before a single header is generated? Add a manual/custom dependency between the `GLOB` operation and the running of `createDefaultKeywordList`?

jokva:
It should work now - directory install is a thing it turns out, and it is marvellous in our eyes.

PR #662, opened 2016-01-28
Deck iterator

jokva:
foo.


jokva:
This was an infrastructure test - please ignore.

PR #664, opened 2016-02-01
Iterator utilities

jokva:
Patch set that provides iterator support.

Previously Deck* classes that had begin/end iterator support returned std::vector::iterator, which directly exposed the underlying data structures. These new iterators changes that behaviour to rather match the expected behaviour of std::iterator, in addition to implement begin/end consistently over all Deck* classes.

In the future more beefed-up containers might get iterator or revised iterator support, in which case this implementation can also be used.

### DIFF
--- a/opm/parser/eclipse/CMakeLists.txt
+++ b/opm/parser/eclipse/CMakeLists.txt
@@ -286,6 +286,7 @@ EclipseState/Tables/TableEnums.hpp
 EclipseState/Tables/TableSchema.hpp
 EclipseState/Tables/TableIndex.hpp
 #
+Utility/Iterator.hpp
 Utility/WconinjeWrapper.hpp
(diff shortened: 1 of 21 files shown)


atgeirr:
Now that was closed before I had a chance to look much at it, but would this enable using range-for such as below?
```
Deck deck;
// populate deck
for (const auto& keyword : deck) {
    // do stuff to keywords
}
```

jokva:
@atgeirr New request: https://github.com/OPM/opm-parser/pull/665

Yes, 100% possible, and why it's supported.

blattms:
Wow custom iterators. That is a can of worms.

You might want to look into the [iterator adaptors of boost](http://www.boost.org/doc/libs/1_55_0/libs/iterator/doc/iterator_adaptor.html), to save you some time (code duplication) and prevent some mistakes.

PR #665, opened 2016-02-01
[WIP] Iterator utilities

jokva:
Patch set that provides iterator support.

Previously Deck* classes that had begin/end iterator support returned std::vector::iterator, which directly exposed the underlying data structures. These new iterators changes that behaviour to rather match the expected behaviour of std::iterator, in addition to implement begin/end consistently over all Deck* classes.

In the future more beefed-up containers might get iterator or revised iterator support, in which case this implementation can also be used.


jokva:
This will likely be rendered obsolete by https://github.com/OPM/opm-parser/pull/677 or subject to heavy modifications. I recommend not merging at this time.

jokva:
This have been rendered obsolete by https://github.com/OPM/opm-parser/pull/677. Closing.

PR #669, opened 2016-02-02
Reimplement Deck<T>Item with templates

jokva:
To reduce code duplication, Deck<T>Item, this patch set implement the Deck* interfaces with templates.


joakim-hove:
The travis build has failed to incorrect ert build; if you rebase and force-push you will get a real test of this PR.

atgeirr:
Net reduction of >300 lines is good! It looks to me like you managed to still put implementation details in the cpp file too, avoiding long compile times, or did I misunderstand it?

jokva:
@joakim-hove I'll do the rebase-push tomorrow.

@atgeirr All implementation details are in C++ files yes. Deck.hpp is slightly larger and more complex (but still a lot simpler than it was just three weeks ago, so net gain), but project compilation time should be improved. Obviously Deck.cpp is now slightly heavier and my rough guess that it would be somewhat shorter than the sum of all the previous DeckItem files.

Of course that's just a nice benefit, the code simplification is the real gain here.

PR #675, opened 2016-02-03
Using unique_ptr in EclipseGrid.

joakim-hove:


joakim-hove:
Waiting for: https://github.com/OPM/opm-common/pull/67

PR #677, opened 2016-02-04
Unique and automatically managed keywords

jokva:
This branch is submitted for early reviews & discussions.

First of all, this is a blocker to https://github.com/OPM/opm-parser/pull/665 - should this direction be approved, some substantial work must be done on the iterator patch, and hopefully most of it will be obsolete in favour of a simpler implementation derived from this branch.

The goal of this patch set is to clarify resource ownership in opm-parser, and slowly replace the (over)use of std::shared_ptr in favour of automatically managed values or unique_ptr where necessary. This also introduces API cleanups, favouring references wherever possible.

Comments & discussion welcome!


jokva:
This patch set has now seen some serious improvements and is reaching maturity.

Short summary:
* Deck has been refactored, but exhibits the same behaviour as before.
* The Deck family of classes now properly owns their resources, and no longer exposes shared_ptrs
* Some unused or unnecessary implementation details removed.

Unfortunately this patch set is not bisectable, but this was done to make the review process simpler. The interesting patches are smaller and nimbler and concentrated on the actual changes, whereas the final patch fixes "the rest of the project" that broke due to the changes.

jokva:
Downstream PRs:

* https://github.com/OPM/opm-autodiff/pull/579
* https://github.com/OPM/opm-core/pull/952
* https://github.com/OPM/opm-porsol/pull/143
* https://github.com/OPM/opm-upscaling/pull/168
* https://github.com/OPM/opm-material/pull/122
* https://github.com/OPM/dune-cornerpoint/pull/205

joakim-hove:
First of all; I think this effort to put the opm-parser code on `std::shared_ptr<T>` is a good thing, so thank you for the effort. This work will certainly be merged; but hopefully I can get some changes in first. I have just started the review - and will continue tomorrow.

joakim-hove:
An advantage of the `std::shared_ptr<T>`approach was that it was not necessary to consider lifetime issues - things just worked in 'java-like' fashion. Now lifetimes must be considered, in the:
```C++
Deck -> DeckKeyword -> DeckRecord -> DeckItem
```
chain the container class (i.e. the class to the 'left') will always assume ownership. In most cases this will eventually lead back to a `Deck` instance owning memory. I think this ownership semantics - i.e. that the container class must stay in town, should be documented.

joakim-hove:
I do not like the `DeckView`class which I feel is a mixture of container (i.e. memory manager - not exactly a view); and iterator/access functionality which is what I would expect to find in a view. I would prefer if the two could be split, possibly using `has a` instead of `is a` relationships.

joakim-hove:
The commit. https://github.com/jokva/opm-parser/commit/c748a058a3a21aeae8164877b63530d479adb4e1 should be removed. Right now this is a lifetime bug; when the Parser goes out of scope the `DeckKeyword` instances will be left with invalid `ParserKeyword` pointers. My suggestion is that we short term let the `ParserKeyword` pointer in the `DeckKeyword` remain a `std::shared_ptr<ParserKeyword>`. This coupling between `DeckKeyword` and `ParserKeyword` is ugly anyway - and I hope that slightly longer term we can remove it completely.

jokva:
> An advantage of the `std::shared_ptr<T>` approach was that it was not necessary to consider lifetime issues - things just worked in 'java-like' fashion.

I'm not entirely convinced that's a good thing, to be honest - even in Java lifetimes must be considered, even though it's less obvious there.

> In most cases this will eventually lead back to a Deck instance owning memory. I think this ownership semantics - i.e. that the container class must stay in town, should be documented.

I don't mind this at all. These rules also apply to all other container oriented code (std::vector etc.), so no surprises there. Furthermore, with the redesigned API this is actually communicated well - only references are returned, but items are -moved- (or copied) into the container.

> I do not like the DeckViewclass which I feel is a mixture of container (i.e. memory manager - not exactly a view)

The `DeckView` does not own any memory, and consequently does not manage it. It is no less a view than an iterator is a view.

> I would prefer if the two could be split, possibly using `has a` instead of `is a` relationships.

What exactly are you thinking of here? If you're referring to `Deck : private DeckView` then the inheritance is private, meaning it is a `has a` relationship already.

> The commit. jokva@c748a05 should be removed. Right now this is a lifetime bug; when the Parser goes out of scope the DeckKeyword instances will be left with invalid ParserKeyword pointers. My suggestion is that we short term let the ParserKeyword pointer in the DeckKeyword remain a std::shared_ptr<ParserKeyword>. This coupling between DeckKeyword and ParserKeyword is ugly anyway - and I hope that slightly longer term we can remove it completely.

This method has one (1) usage, in section.cpp, so I'm confident this dependency can be resolved.

atgeirr:
Just two comments (have not done a thorough review):
 - for some of the objects dealt with in this library, shared ownership may actually be appropriate, even if for most objects it is not,
 - private inheritance should be considered an exotic feature better avoided, using containment is usually better (although I have been guilty of using it in the past myself).

joakim-hove:
Regarding view: I was wrong with the ownership.

jokva:
@joakim-hove 
> The commit. jokva@c748a05 should be removed. 

This dependency has been removed and the lifetime issues should disappear.

joakim-hove:
Ok - this is good work; going in.

PR #679, opened 2016-02-04
Add ADSALNOD keyword.

joakim-hove:


andlaus:
stupid question: what does this keyword do?

it seems to be pretty new because in my RM (for 2011.1) it is not featured...

joakim-hove:
>  what does this keyword do?

I haven't got a clue; but it appears in a deck I have tried to parse.

> it seems to be pretty new because in my RM (for 2011.1)

Probably - all I know is that it figures in the 2014.2 manual.

alfbr:
The keyword is used to specify salt concentrations, which in turn is used to calculate absorption of polymer.

PR #681, opened 2016-02-10
Fix minor issues

atgeirr:
This fixes:
 - missing includes after refactoring (triggered on my system at least)
 - two minor warnings.

### DIFF
--- a/opm/parser/eclipse/Deck/Deck.hpp
+++ b/opm/parser/eclipse/Deck/Deck.hpp
@@ -23,6 +23,7 @@
 #include <map>
 #include <memory>
 #include <vector>
+#include <string>
 
 namespace Opm {
 
--- a/opm/parser/eclipse/Deck/DeckStringItem.hpp
+++ b/opm/parser/eclipse/Deck/DeckStringItem.hpp
@@ -35,7 +35,7 @@ namespace Opm {
             const std::string& getString( size_t index ) const override;
             const std::vector< std::string >& getStringData() const override;
 
-            std::string getTrimmedString( size_t index ) const;
+            std::string getTrimmedString( size_t index ) const override;
     };
 
     typedef std::shared_ptr< DeckStringItem > DeckStringItemPtr;
--- a/opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.cpp
@@ -25,7 +25,7 @@ namespace Opm {
     }
 
 
-    WellProductionProperties WellProductionProperties::history(double BHPLimit , DeckRecordConstPtr record, bool addGroupProductionControl)
+    WellProductionProperties WellProductionProperties::history(double BHPLimit , DeckRecordConstPtr record, bool /* addGroupProductionControl */)
     {
         // Modes supported in WCONHIST just from {O,W,G}RAT values
         //
--- a/opm/parser/eclipse/IntegrationTests/ParseTOPS.cpp
+++ b/opm/parser/eclipse/IntegrationTests/ParseTOPS.cpp
@@ -31,6 +31,8 @@
 #include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
 #include <opm/parser/eclipse/EclipseState/Grid/EclipseGrid.hpp>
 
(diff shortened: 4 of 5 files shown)


atgeirr:
There is one more group of warnings I get now from opm-parser that I did not try to fix:
```
/Users/atgeirr/opm-build-debug/opm-parser/opm/parser/eclipse/Deck/DeckItem.cpp:70:56: warning: declaration shadows a field of 'DeckTypeItem<T>' [-Wshadow]
    void DeckTypeItem< T >::push_back( std::deque< T > data, size_t items ) {
                                                       ^
/Users/atgeirr/opm-build-debug/opm-parser/opm/parser/eclipse/Deck/DeckItem.hpp:141:30: note: previous declaration is here
            std::vector< T > data;
                             ^
```
The root cause is simple, using `data` as name for both function parameter and member variable. Perhaps @jokva could have a brief look? I believe the simplest fix is probably to use the established convention of marking member variables in some way (usually m_name in opm-parser, name_ in some of the other places).

akva2:
and why is data copied there in the first place?

jokva:
@akva2 This is an artefact of the "old" Deck* family of classes (which are under heavy review right now). See https://github.com/OPM/opm-parser/pull/677 for the first iteration.

@atgeirr The default build is without -Wshadow, so these never show up. It is pseudo-intentional on my end anyway, as I find the hungarianesque notation of m_ horrible. I am willing to reconsider this stance, but I find it good practice to use this-> always anyway as this makes it *very* clear where some symbol belongs.

akva2:
right. i don't immediately see how that pr adresses my question, but i trust you on that. to be explicit the function should eat a const std::deque<T>&, not a std::deque<T>. right now this whole class (DeckTypeItem) screams JAVA PROGRAMMER, since everything is call by value. (for primitives that's ofc fine, but there's no guarantee T is a primitive here, and a deque is never a primitive).

atgeirr:
I think -Wshadow is (usually) worthwhile, it has uncovered unclear code in the past. I think you must reconsider your stance on the hungarianesque (fantastic word...) m_ if for no other reason than it is already established. The rest of us expect it in opm-parser and risk getting needlessly confused if some code stop using the convention.

GitPaean:
Can we add `-Wextra` and `-Wshadow` to the default compilation options instead of making extra efforts (often resulting in separate PRs) in pointing out or removing the warnings?

> I think -Wshadow is (usually) worthwhile, it has uncovered unclear code in the past. I think you must reconsider your stance on the hungarianesque (fantastic word...) m_ if for no other reason than it is already established. The rest of us expect it in opm-parser and risk getting needlessly confused if some code stop using the convention.

jokva:
@akva2 I didn't change the deque thing to keep the change "minimal" - everything -was- passed by value before, but that's about to change. It's a step in a larger, let's be honest, redesign of the Deck internals. This is cruft being trimmed down, but one change at the time to keep the transition smooth (er).

@akva2 The DeckItem should own whatever data is passed in. I believe the deque thing is a legacy artefact to pass more than one value at once, but it's under review. DeckItem does contain a vector of primitives (T in DeckTypeItem will always be a primitive; it's implementation is in a cpp file and it is explicitly instantiated in the same file to int, double and std::string (sort-of primitive anyway)). In fact, the final interface might change that to a vector that is passed by value to signalise that an underlying copy will take place, enabling some optimisations and in worst case costing nothing as opposed to passing reference (since the memory must be copied to the internal vector anyway). I'm rambling slightly, but the point is that while the Deck* still has some obvious legacy to it now, that is underway of being modernised.

@atgeirr While I can possibly agree on the already-established argument, I'm not so sure that I'm convinced it has a good basis. I preserved the name of the data argument here for transitional reasons, but I generally find that using a different name for the argument than the internal variable is good practice anyway, for more than just disambiguating reasons. It also makes code accessing the private vars have clearer names (m_ is *clunky*) which happens far more often than code that adds to it (typically set-type methods).

jokva:
That being said, I think the patch is fine and is ok to merge.

jokva:
@akva2 The `push_back( deque )` feature is not used at -all- in the projects. I have suggested it for removal: https://github.com/OPM/opm-parser/pull/683

joakim-hove:
Will merge: #684

PR #683, opened 2016-02-12
Removes push_back( deque ) from DeckItem

jokva:
This feature is not used at all. I suggest it is removed, and if a need arises it can be reintroduced.


joakim-hove:
This is a relic from a previous implementering,  good to get rid of. Merking.

PR #687, opened 2016-02-15
EclipseState: make the grid properties returned by get*GridProperty() constant

andlaus:
that's because they are not supposed to be modified outside of the
EclipseState. (if they are, why? I'd consider that *very* bad style
since it is also possible to copy these objects and modify the copy
and also this was used nowhere within the OPM project.)

also, the has*Property() is now working as expected: e.g.,

```c++
bool hasSatnum = eclipseState->hasIntProperty("SATNUM");
eclipseState->getIntProperty("SATNUM");
assert(hasSatnum == eclipseState->hasIntProperty("SATNUM"));
```

will now work for decks which do not explicitly specify
SATNUM. (before the getIntProperty() method silently created the
SATNUM property which caused hasIntProperty() to change its
opinion. With this patch, the property will still be silently created,
but has\*Property() ignores it, i.e., that method could be renamed to
hasExplicit*Property() which -- as far as I understand this -- was its
intention from start.)


andlaus:
be aware this PR needs a mop-up patch for opm-core (OPM/opm-core#953). all other modules seem to be unaffected...

joakim-hove:
> I'd consider that very bad style

I agree - that would be very bad style.

> the has*Property() is now working as expected: e.g.,

I agree that the `hasXXXGridProperty()` has always felt somewhat klunky - and semantically ~correct. I certainly agree with the main goal of this patch, i.e. to make the gridproperties returned constant; however when it comes to the `hasXXXProperty()`I would strongly prefer that you followed this approach:

> [....] i.e., that method could be renamed to hasExplicit*Property() which -- as far as I understand this -- was its intention from start.)

which is - as you say - roughly the intention from the start. Instead of `hasExplicitProperty()` I would suggest `hasInputProperty()` or `hasDeckProperty()`?

I have not really looked at `adHockeyword()` implementation - but it did seem a bit overkill to me? As for which properties have been assigned from the deck I would assume that could be maintained with a simple `std::set<std::string>` which was updated in the `EclipseState::loadGridPropertyFromDeckKeyword()` method.

andlaus:
> which is - as you say - roughly the intention from the start. Instead of hasExplicitProperty() I would suggest hasInputProperty() or hasDeckProperty()?

I don't mind renaming this to whatever you prefer. In the interest of not breaking downstream code I'd propose to keep the current `has*GridProperty()` methods and just mark them as deprecated (thus only causing warnings downstream).

> I have not really looked at adHockeyword() implementation - but it did seem a bit overkill to me? 

well, it is completely internal since it does not provide any public methods. (Or it should be that way.) alternatives to the implementations would be to return grid properties by value (potentially slow, but might require less memory once the input has been internalized by the simulator) or to use separate containers for ad-hoc and explicit grid properties (in that case, converting an ad-hoc to an explicit grid property is more elaborate).

> As for which properties have been assigned from the deck I would assume that could be flipped around - instead of keeping rack of which keywords have been auto created with the adHoc machinery - a std::set<std::string> which was updated in the EclipseState::loadGridPropertyFromDeckKeyword() method could be used to keep track of which keywords were explicitly provided in the deck. That would also form the essentials of the hasInputProperty() implementation?

true and I don't mind implementing it that way. I guess that this would be pretty much the same amount of effort from the perspective of the implementation, though. do you still insist?

joakim-hove:
>I don't mind renaming this to whatever you prefer.

OK - then I suggest `hasDeck*Property()`

> In the interest of not breaking downstream 

I assume the breakage will be minimal? - unless I am mistaken on this I think we should just take the breakage - and the required fixes right away.

>true and I don't mind implementing it that way. 

:smile: 

>I guess that this would be pretty much the same amount of effort from the perspective of the implementation, though. do you still insist?

Yes. I really think that would be quite a lot simpler; e.g.:
```
GridProperties::getDeckProperty(const std::string& keyword) const {
     m_deckProperties.insert( keyword );
     return getProperty( keyword );
}
```

andlaus:
>>  I don't mind renaming this to whatever you prefer.
>
> OK - then I suggest hasDeck*Property()

good. I added a renaming patch.

> I assume the breakage will be minimal? - unless I am mistaken on this I think we should just take the breakage - and the required fixes right away.

sure, but if it can be avoided. so far the patch makes the old names deprecated.

> Yes. I really think that would be quite a lot simpler; e.g.:
>
> GridProperties::getDeckProperty(const std::string& keyword) const {
>     m_deckProperties.insert( keyword );
>     return getProperty( keyword );
>}

that's what the current `getInitialized*GridProperty()` methods do? wouldn't the `get*GridProperty()` method be more like:

```c++
GridProperties::getDoubleDeckProperty(const std::string& keyword) const {
   if (m_deckProperties.count(keyword) > 0)
    return m_deckProperties.at(keyword);
   else if (m_nonDeckProperties.count(keyword) > 0)
    return m_nonDeckProperties.at(keyword);
  else {
    m_nonDeckProperties[keyword] = createNonDeckDoubleProperty_(keyword);
    return m_nonDeckProperties.at(keyword);
   }
```

?

also, you have to handle the case where a "non-deck" property becomes a deck property. (That happens if a property which was not mentioned in the deck is required to initialize a keyword which is explicitly mentioned and later the non-deck keyword is explicitly mentioned as well. e.g.:

```
PORV
100* /

NTG
100*0.99 /
```

andlaus:
(note that I don't know what E100 does for the example above: does it use `1.0` for `NTG` to initialize `PORV` or does it use `0.99`? Also, it's not clear what the user wanted to express with this...)

joakim-hove:
> (note that I don't know what E100 does for the example above: does it use 1.0 for NTG to initialize PORV or does it use 0.99? 

I *think* it would use 0.99 - but I am certainly not sure; opm would at least use 0.99.

joakim-hove:
I actually insist that as long is is used in the `EclipseState::loadGridPropertyFromDeckKeyword()` method  this:

```C++
GridProperties::getDeckProperty(const std::string& keyword) const {
   m_deckProperties.insert( keyword );
   
   // This should call the existing getProperty() which will autocreate if required:
   return getProperty( keyword ); 
}
```

should work. In the case you describe where a keyword appears in the deck at a stage where it has actually been autocreated already the unconditional `m_deckProperties.insert( keyword )` will fix the in deck property. 

Or do you disagree?

andlaus:
> I think it would use 0.99 - but I am certainly not sure; opm would at least use 0.99.

nope. I'm pretty sure it will use the default for NTG (which is 1.0 IIRC). That's because it will creates the NTG grid property ad-hoc (and thus using the default values) when it needs to calculate the defaults for initializing PORV. (see https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/EclipseState.cpp#L83 et al.)

andlaus:
(my "it" is OPM, not Eclipse ;)

andlaus:
> I actually insist that as long is is used in the EclipseState::loadGridPropertyFromDeckKeyword() method this: [..] should work.

I agree and if I did not screw up horribly, it still will work after this PR. if the requested grid property did not exist before, it still creates it. (the only difference to the current behavior is that this new property is flagged 'ad-hoc' and thus is not considered by the size() and has*GridProperty() methods.)

joakim-hove:
> I'm pretty sure it will use the default for NTG (which is 1.0 IIRC)

I back off - you are right, this is a bug. As already mentioned I have not really looked into your implementation (my gut feeling was just that this was too much fiddling to support a detail - but maybe I am wrong). Anyway - do you claim that your implementation fixes this?

andlaus:
(i.e., after this patch the only semantic difference to the current behavior is that it is still possible to find out if a grid property was explicitly specified by the deck if `get*GridProperty()` was called for a grid property that does not exist.)

andlaus:
> this is a bug.

I'm not so sure. but it is probably unintended. maybe the code should be changed so that grid properties cannot remove their "ad-hoc flag". (because I suppose most users will not expect this.)

andlaus:
> Anyway - do you claim that your implementation fixes this?

no, I'm not claiming this: this PR does not change anything in this regard. It is just a thing I stumbled over while coding it...

andlaus:
> maybe the code should be changed so that grid properties cannot remove their "ad-hoc flag".

alternatively it could produce a Big Fat Warning...

andlaus:
ok, I've added a patch to print such a warning.

opm/parser/eclipse/EclipseState/EclipseState.cpp
```diff
@@ -93,21 +93,24 @@ namespace Opm {
...
                         throw std::logic_error("Do not have information for the PORV keyword - some defaulted values in PORO");
-                    {
+                    else {
+                        const auto& poroData = poro->getData();
                         for (size_t globalIndex = 0; globalIndex < porv->getCartesianSize(); globalIndex++) {
-                            if (std::isnan(porv->iget(globalIndex))) {
-                                double cell_poro = poro->iget(globalIndex);
+                            if (std::isnan(values[globalIndex])) {
+                                double cell_poro = poroData[globalIndex];
                                 double cell_ntg = ntg->iget(globalIndex);
```
andlaus: @joakim-hove: isn't MULTPV required here as well?

andlaus:
I rebased this one on top of the current master (there were conflicts with #677) and squashed the cleanup patches into the main one. If travis is happy about it can it be merged?

joakim-hove:
I am sorry this has fallen out of my radar. I do not like the `adHoc...` prefix; I will merge it when you have done replaced `adHocxxxx` with e.g. `autoGenerated`.

andlaus:
> I do not like the adHoc... prefix; I will merge it when you have done replaced adHocxxxx with e.g. autoGenerated.

done. (For the sake of a "clean" history, I've squashed this into the main patch.)

joakim-hove:
Thanks - merging!

PR #691, opened 2016-02-18
Parser stores keywords as unique_ptr

jokva:
Removes shared_ptr< ParserKeyword > exchange in Parser interface and
replaces it with unique_ptr storage and raw pointers for return values.
This has some implications:

* addParserKeyword() no longer takes shared_ptr<>, but unique_ptr&&
  addParserKeyword has been modified to create unique_ptr instead of
  shared_ptr.

* generated-source/ParserKeyword* no longer use make_shared which had a
  -massive- impact on compile times, which are now more-or-less
  trivialised (on my machine: 7.5s -> 1s per file). This because the
  compiler no longer generates a bunch of forwarding make_shared
  instances of subclasses that are immediately thrown away, but rather
  uses an inline make_unique that instantiates the -parent- class
  unique_ptr.

### DIFF
--- a/opm/parser/eclipse/Applications/opm-eclkw.cpp
+++ b/opm/parser/eclipse/Applications/opm-eclkw.cpp
@@ -26,7 +26,7 @@
 #include <opm/parser/eclipse/Parser/ParserKeyword.hpp>
 
 
-static void printKeyword(Opm::ParserKeywordConstPtr keyword)
+static void printKeyword( const Opm::ParserKeyword* keyword)
 {
     std::string indent = " ";
     std::cout << keyword->getName() << std::endl;
@@ -65,7 +65,7 @@ static void printItem(Opm::ParserItemConstPtr item, std::string indent)
     std::cout << std::endl;
 }
 
-static void printItems(Opm::ParserKeywordConstPtr keyword)
+static void printItems( const Opm::ParserKeyword* keyword)
 {
     std::string indent = "  ";
     std::cout << std::endl;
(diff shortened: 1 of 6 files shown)


atgeirr:
The large change in compilation time is welcome, but surprising!

I think some functions that now take naked pointers should take references instead, since they do not support passing a null pointer (such as printKeyword). Also, functions that return (const) pointers could return (const) references if returning null is not an option (many of the function throw if they get invalid requests).

jokva:
It's not really that surprising (it was initially to me at first, but I ran through the compilation process in my head and it makes perfect sense really); here's why:

`template addKeyword` uses `make_shared< T >` to construct some T passed in. Now, instantiating addKeyword itself is a fairly lightweight thing to do, and measurements proved it only slightly heavier (we have approx 350 instances) than using `addParserKeyword` with raw new. However, every instance now also instantiates a unique `make_shared< T >` which is non-trivial (takes ~7x as long), but the value is upcasted to `shared_ptr< ParserKeyword >`, so all the work done in ParserKeywordsN.cpp is immediately discarded. Now we only make *one* instance of `make_unique< ParserKeyword >`, but N separate instances of addKeyword.

--

> I think some functions that now take naked pointers should take references instead, since they do not support passing a null pointer (such as printKeyword). Also, functions that return (const) pointers could return (const) references if returning null is not an option (many of the function throw if they get invalid requests).

I agree, but that should be a separate patch as that will require several changes throughout the codebase. Going from return-shared to return-raw-pointer is light as they're accessed the same way - references are not.

joakim-hove:
Very weøcome,  I am grateful for the effekt in curing opm-parser of shared_ptr disease.

PR #695, opened 2016-02-26
Refactor grid property intialisers

jokva:
A pretty substantial refactor for the grid property initialisers. Notable changes:

* The build hack using template parameters to avoid a circular include with EclipseState has been resolved. Because of this, the initialiser functions are all implemented in source files, not in headers.
* Related, this cleans up the actual headers a lot - now only contains definitions and no unnecessary includes.
* Now uses a custom function object instead of a virtual base interface. This makes all initialiser functions actual functions, and they behave accordingly (no more arbitrary side effects).
* Exploits C++11 init lists for a slightly more compact registration of supported keywords.
* The initialisation code has been heavily rewritten, with emphasis on every function being short and single-purpose. A nice effect from this was that we now only compute exactly the data we need, instead of computing ~20 vectors of data when only a couple were in use.


andlaus:
> - Exploits C++11 init lists for a slightly more compact registration of supported keywords.

I might be wrong (the details are subtle), but the original implementation used initializer lists. It was changed to the conventional approach in #517 because some compilers seem to have troubles with it (which result in long compile times). I don't know if this is still relevant, though...

jokva:
It doesn't seem to be an abnormally slow compilation on 4.8.5.

jokva:
Regardless, no init list is that long. They're mostly init lists over 8 (?) strings, passed to the regular add call.

andlaus:
> It doesn't seem to be an abnormally slow compilation on 4.8.5. I haven't tried clang.

it could also be a MacOS issue. probably @atgeirr should test it out.

> Regardless, no init list is that long. They're mostly init lists over 8 (?) strings, passed to the regular add call.

ok. then there's probably no problem here.

jokva:
This will be rebased once https://github.com/OPM/opm-parser/pull/696 is merged, but this patch depends on that commit.

opm/parser/eclipse/EclipseState/EclipseState.cpp
```diff
@@ -46,80 +47,55 @@ namespace Opm {
 
     namespace GridPropertyPostProcessor {
 
-        class DistributeTopLayer : public GridPropertyBasePostProcessor<double>
-        {
-        public:
-            DistributeTopLayer(const EclipseState& eclipseState) :
-                m_eclipseState( eclipseState )
-            { }
+        std::vector< double >& distTopLayer( std::vector<double>& values, const Deck& m_deck, const EclipseState& m_eclipseState ) {
```
joakim-hove: Is it necessary to return the values? As long as this function indeed mutates the input argument `values` I would prefer the function to be `void` - then everyone and his brother will understand that mutation of input argument take place.

joakim-hove:
This looks good - and I will merge it (modulo some return value fighting ...); but I would prefer to wait until the ultimate regression test: *Successfully run Norne* has been performed.

opm/parser/eclipse/EclipseState/EclipseState.cpp
jokva: It's not *necessary*, but comes down to a mix of taste and goal. Personally I don't think passing reference for mutation like this is a good thing at all (I'll happily take a look into changing that properly i.e. GridPropertyFunction doesn't mutate but rather return a new vector).

Other than that, I prefer returning the reference because it allows for `return subfunction(values)` where `subfunction` is some static function in a source file. Ignoring the return value effectively makes it void anyway.

jokva:
I'll run it on Norne and let you all know when this is done sucessfully.

jokva:
I ran Norne with and without this patch applied, and the results are identical. They both differ from the newly-added reference solution, but this might be because of patches to opm-core and output writing.

It's safe to assume that this patch doesn't introduce regressions (w.r.t. correctness).

joakim-hove:
> I ran Norne with and without this patch applied, and the results are identical. 

Good - merging.

PR #696, opened 2016-02-26
Introduce Utility/Functional

jokva:
Utility/Functional is a lightweight high level functional-oriented sub library that attempts to abstract some common uses and boilerplate around the parser code (and later maybe for other modules to use).

This patch introduce only three functions, but they have proven common enough to warrant some common implementation.

I plan for things here to be added on a by-need basis.

### DIFF
--- a/opm/parser/eclipse/Utility/Functional.hpp
+++ b/opm/parser/eclipse/Utility/Functional.hpp
@@ -1,0 +1,214 @@
+#include <algorithm>
+#include <iterator>
+#include <vector>
+
+namespace Opm {
+
+namespace fun {
+
+    /*
+     * The Utility/Functional library provides convenient high level
+     * functionality and higher order functions inspiried by functional
+     * languages (in particular Haskell) and modern C++. The goal is to provide
+     * lightweight features that reduce boilerplate and make code more
+     * declarative.
+     */
+
+    /*
+     * map :: (a -> b) -> [a] -> [b]
+     *
+     * maps the elements [a] of the passed container C to [b], by using the
+     * passed function f :: a -> b. Works like map in haskell, lisp, python etc.
+     *
+     * C can be any foreach-compatible container (that supports .begin,
+     * .end), but will always return a vector.
+     *
+     * F can be any Callable, that is both function pointer,
+     * operator()-providing class or std::function, including lambdas. F is
+     * typically passed by reference. F must be unary of type A (which must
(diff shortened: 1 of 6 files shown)


joakim-hove:
Gee - looking at your beautiful documentation even I can see the benefit of documentation.

opm/parser/eclipse/Utility/Functional.hpp
```diff
@@ -0,0 +1,214 @@
...
+     *
+     * int plus1( int x ) { return x + 1; }
+     * base_vec = { 0, 1, 2, 3, 4 };
+     * vec = fun::map( &plus1, base_vec );
+     *
+     * vec => { 1, 2, 3, 4, 5 }
+     *
+     * --
+     *
+     * int mul2 = []( int x ) { return x * 2; };
```
atgeirr: Minor comment glitch: `int` should be `auto`.

opm/parser/eclipse/Utility/Functional.hpp
```diff
@@ -0,0 +1,214 @@
...
+                []( std::size_t acc, const std::vector< A >& x ) {
+                    return acc + x.size();
+                    }
+                );
+
+        std::vector< A > dst;
+        dst.reserve( size );
+
+        for( auto& x : src )
+            std::move( x.begin(), x.end(), std::back_inserter( dst ) );
```
atgeirr: I cannot find a reference for this way to call move, is it really C++11?

jokva: Yes. http://en.cppreference.com/w/cpp/algorithm/move

opm/parser/eclipse/Utility/Functional.hpp
jokva: Absolutely so.

opm/parser/eclipse/Utility/Functional.hpp
atgeirr: Thanks, googling for std::move I only found the <utility> version and not the <algorithm> version, a bit strange that the two overloads are in different headers!

jokva: Nine times out of ten it's the std::move from utility you're looking for, as std::move(begin, end, output) is rather specific to the algorithms library. Confusing maybe, but makes sense in my opinion. It's a *very* useful feature however, as demonstrated here.

PR #698, opened 2016-02-29
Equil keyword lookup

jokva:
Makes https://github.com/OPM/opm-core/pull/960 obsolete.

Removes the old Utility/EquilWrapper in favour of an EclipseState managed, slightly richer implementation. Instead of depending on Deck, 'EQUIL' derived information can be read from the EclipseState.


opm/parser/eclipse/EclipseState/InitConfig/Equil.cpp
```diff
@@ -0,0 +1,79 @@
...
+namespace Opm {
+
+    EquilRecord::EquilRecord( const DeckRecord& record ) :
+            datum_depth( record.getItem( 0 ).getSIDouble( 0 ) ),
+            datum_depth_ps( record.getItem( 1 ).getSIDouble( 0 ) ),
+            water_oil_contact_depth( record.getItem( 2 ).getSIDouble( 0 ) ),
+            water_oil_contact_capillary_pressure( record.getItem( 3 ).getSIDouble( 0 ) ),
+            gas_oil_contact_depth( record.getItem( 4 ).getSIDouble( 0 ) ),
+            gas_oil_contact_capillary_pressure( record.getItem( 5 ).getSIDouble( 0 ) ),
+            live_oil_init_proc( record.getItem( 6 ).get< int >( 0 ) ),
```
joakim-hove: Observe that the three `int`values at the end of record are really used as enums:

```C++
if (live_oil_init_proc > 0)
   // Use init method1
else
   // Use init method2
```

What data is internalized is of course an implementation detail, but the public API should focus on the actually interesting data:

```C++ 
enum LiveOiilInitMethod {RSVD , CONSTANT_RS}

enum LiveOilInitMethod EquilRecord::liveOilInitProcedure() const {
     if (live_oil_init_proc > 0)
        return RSVD;
     else
        return CONSTANT_RS;
}
```

and similar for the two last items.

opm/parser/eclipse/EclipseState/InitConfig/Equil.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+    int EquilRecord::wetGasInitProcedure() const {
+        return this->wet_gas_init_proc;
+    }
+
+    int EquilRecord::initializationTargetAccuracy() const {
+        return this->init_target_accuracy;
+    }
+
+    Equil::Equil( const DeckKeyword& keyword ) :
```
joakim-hove: Elegant!

opm/parser/eclipse/EclipseState/InitConfig/Equil.cpp
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    int EquilRecord::wetGasInitProcedure() const {
+        return this->wet_gas_init_proc;
+    }
+
+    int EquilRecord::initializationTargetAccuracy() const {
+        return this->init_target_accuracy;
+    }
+
```
joakim-hove: Can I please have some indication in the code - e.g. `/*************** ..*/` that we are now actually changing class.

opm/parser/eclipse/EclipseState/InitConfig/Equil.hpp
```diff
@@ -0,0 +1,61 @@
...
+            int wet_gas_init_proc;
+            int init_target_accuracy;
+    };
+
+    class Equil {
+        public:
+            using const_iterator = std::vector< EquilRecord >::const_iterator;
+
+            Equil() = default;
+            Equil( const DeckKeyword& );
```
atgeirr: Explicit?

opm/parser/eclipse/EclipseState/InitConfig/Equil.cpp
jokva: This is coming.

opm/parser/eclipse/EclipseState/InitConfig/Equil.cpp
jokva: I can add this, sure.

opm/parser/eclipse/EclipseState/InitConfig/Equil.hpp
jokva: Do you mean the constructor? I have no strong opinion on it for cases like this, but it can be made explicit for sure.

atgeirr: Yes, it should be explicit to avoid implicit conversion (unless that is your specific intent).

jokva: There are *very* few cases where implicit conversion here would be a problem, but if there's opinion on having it explicit then I'll add it.

PR #701, opened 2016-03-02
More function opmlog

qilicun:
Add some more functions and members for OpmLog for further work,

### DIFF
--- a/opm/parser/eclipse/OpmLog/OpmLog.cpp
+++ b/opm/parser/eclipse/OpmLog/OpmLog.cpp
@@ -96,8 +96,8 @@ namespace Opm {
             std::string("\nWarnings          " + std::to_string(m_warning)) +
             std::string("\nProblems          " + std::to_string(m_problem)) +
             std::string("\nErrors            " + std::to_string(m_error)) + 
-            std::string("\nBugs              " + std::to_string(m_bug))+ "\n";
-            std::string("\nInfo              " + std::to_string(m_info))+ "\n";
+            std::string("\nBugs              " + std::to_string(m_bug)) +
+            std::string("\nInfo              " + std::to_string(m_info)) + "\n";
 
         m_streamLog->addMessage(Log::MessageType::Info, summary_msg);
     }
--- a/opm/parser/eclipse/OpmLog/OpmLog.cpp
+++ b/opm/parser/eclipse/OpmLog/OpmLog.cpp
@@ -19,7 +19,6 @@
 
 #include <opm/parser/eclipse/OpmLog/OpmLog.hpp>
 #include <opm/parser/eclipse/OpmLog/Logger.hpp>
-#include <opm/parser/eclipse/OpmLog/StreamLog.hpp>
 
 namespace Opm {
 
--- a/opm/parser/eclipse/OpmLog/OpmLog.cpp
+++ b/opm/parser/eclipse/OpmLog/OpmLog.cpp
@@ -73,6 +73,13 @@ namespace Opm {
     }
 
     
+    void OpmLog::debug(const std::string& message)
+    {
+        const std::string msg = Log::prefixMessage(Log::MessageType::Debug, message);
+        addMessage(Log::MessageType::Debug, msg);
+    }
+
+
     bool OpmLog::enabledMessageType( int64_t messageType ) {
         if (m_logger)
             return m_logger->enabledMessageType( messageType );
--- a/opm/parser/eclipse/OpmLog/OpmLog.cpp
+++ b/opm/parser/eclipse/OpmLog/OpmLog.cpp
@@ -19,6 +19,7 @@
 
 #include <opm/parser/eclipse/OpmLog/OpmLog.hpp>
 #include <opm/parser/eclipse/OpmLog/Logger.hpp>
+#include <opm/parser/eclipse/OpmLog/StreamLog.hpp>
 
 namespace Opm {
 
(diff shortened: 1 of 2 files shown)


qilicun:
Any comments here?
@joakim-hove @atgeirr

joakim-hove:
I Will look - am at holiday and it was a bit moren detail than I could
handle at a glance. Sorry for the delay.
6. mars 2016 22:01 skrev "Liu Ming" <notifications@github.com> følgende:

atgeirr:
While I agree with the intent (adding simple functions for making log messages) and think the interface is fine, I believe the implementation is flawed.

The StreamLog class is a logging backend, and should be added to the Logger, as is done in FlowMain::readDeckInput(). (I guess that code should be moved to its own function setupLogging() or something.)

I was confused, too: I think the logging backend classes should be renamed StreamLogBackend and CounterLogBackend to avoid future confusion.

qilicun:
@joakim-hove 
Enjoy your vacation.

@atgeirr 
I am not sure that I  understand the Logger Backend 100%. How to use it also confuse me. 
But my idea is quite simple, that is to make OpmLog more powerful, also I think if it is possible, StreamLog, CounterLog and TimerLog can be removed or merged into OpmLog, that will be much clear.

joakim-hove:
> I think the logging backend classes should be renamed StreamLogBackend and CounterLogBackend to avoid future confusion.

I agree.

joakim-hove:
This breaks the current implementation for logging quite hard, and can not be merged in it's current form. I see that your new code is quite heavily towards Eclipse .PRT file requirements - whereas the current logging functionality is more general. There is clearly some impedance mismatch here.

Logging is currently organized as follows:

1. There is a class `Logger.cpp` https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/OpmLog/Logger.cpp which can be instantiated.

2. The Logger class does not log anything itself, it just handles a collection of logging *backends* which can do something concrete with a log message. A *very* typical thing to do with a log message is of course to print it to a file - that is what the `StreamLog` does - but in principle a log backend can do *anything* with a log message - including posting it on facebook.

3. The Logger class has a method `addMessage()` which will go through all the backends in the logger and call the `addMessage()` of the individual backends.

In principle one could instantiate a Logger class in e.g. `flow` and then send a reference to that logger instance to all methods which might want to log something. Since that would require adding a `,Logger& logger)` to veeeery many methods - we have instead implemented the singleton `OpmLog` which serves as a global frontend to one shared `Logger` instance. In the current PR you are adding a `StreamLog`backend directly to this singleton; that is just wrong.

Brutal review aside - I am by no means blind to problems/limitations/deficits of the current logging implementation, but the current PR is to much 'bolted on' - without really fixing the problems. 

I do not really see a "perfect solution" here, but my suggestion is the following:

1. You can create a backend class `EclipsePRTStreamLog which is derived from `StreamLog`.
2. The `EclipsePRTStreamLog` class keeps track of the number of messagetype - based on the argument to `addMessage()`- this is essentially the content of the current `CounterLog` implementation. 
3. The `EclipsePRTStreamLog` destructor should add the summary of how many warnings/errors/...
4. The `EclipsePRTStreamLog` *could* even stop the whole simulation if the number of warnings were too many e.t.c. - although that is *quite ugly*.

I hope this makes some sense; the effort required to implement my suggestion should actually be quite limited.

joakim-hove:
Looking at this one more time; my main gripe with it is the adding of member `m_streamLog` and counters for number of `info`,  `error`+++ on the `OpmLog` singelton. 

The methods `OpmLog::info( ), OpmLog::error( )` and so on make sense - however they should invoke the normal `OpmLog::addMessage( )` functionality:

```C++
void OpmLog::info(const std::string& msg) {
    addMessage( Log::MessageType::Info , msg );
}
```

The reason such convenience methods were not added originally was that I wanted the set of possible message types to be runtime configurable; however that is probably too much flexibility anyway - and the existence of named methods as you have suggested certainly makes the functionality simpler to use. Feel free to contact me if you are uncertain on how to proceed.

qilicun:
@joakim-hove 
After reading your comments several times, I finally understand how OpmLog works and the backend meaning. 

For this PR, I admit ``StreamLog`` and the int members for counting break the current implementation,  they should be removed.  

>The methods OpmLog::info( ), OpmLog::error( ) and so on make sense - however they should invoke the normal OpmLog::addMessage( ) functionality:

Yes, true. Apart from this, do you think it's better to keep these methods or not?

atgeirr:
> Yes, true. Apart from this, do you think it's better to keep these methods or not?

I'd like to keep the methods, as I said earlier, the interface is fine.

qilicun:
>I'd like to keep the methods, as I said earlier, the interface is fine.

Good! I will update the PR according to the discussion.

joakim-hove:
> After reading your comments several times, I finally understand how OpmLog works and the backend meaning. Thanks!

Good - athough your comment points out that to my "self-documenting" code is not always up to it :smile:.

>@atgeirr: I'd like to keep the methods,

Sure - I am happy with that.

opm/parser/eclipse/OpmLog/OpmLog.cpp
```diff
@@ -19,6 +19,7 @@
 
 #include <opm/parser/eclipse/OpmLog/OpmLog.hpp>
 #include <opm/parser/eclipse/OpmLog/Logger.hpp>
+#include <opm/parser/eclipse/OpmLog/StreamLog.hpp>
```
joakim-hove: Please remove this include,

joakim-hove:
OK - in it's current form this PR is a beautification of the `OpmLog` interface - except the superfluous include in `OpmLog.cpp` I am happy to merge it as it is.

But: do you intend to add a new backend class combining output to file with a counter functionality - functionality wise quite similar to what you had already implemented? I would be happy to merge that as well.

qilicun:
>But: do you intend to add a new backend class combining output to file with a counter functionality - functionality wise quite similar to what you had already implemented? I would be happy to merge that as well.

I think so :). if not, in the user code like flow, the user need to call ``CounterLog`` to do the summary if the y like.

opm/parser/eclipse/OpmLog/OpmLog.hpp
```diff
@@ -24,6 +24,7 @@
 #include <cstdint>
 
 #include <opm/parser/eclipse/OpmLog/Logger.hpp>
+#include <opm/parser/eclipse/OpmLog/LogUtil.hpp>
```
joakim-hove: This include can also be removed.

joakim-hove:
> I think so :)

Good - I would encourage that; the current `CounterLog` implementation is mainly meant as an illustration. To use it to report the number of errors / warnings / ... in a file would require communication between the stream backend and the counter backend; such communication wasn't in the brochure.

opm/parser/eclipse/OpmLog/OpmLog.hpp
qilicun: When I remove this line, build failed.

```
/private/miliu/project/opm/opm-parser/opm/parser/eclipse/OpmLog/OpmLog.cpp:42:33: error: ?Log? has not been declared
         const std::string msg = Log::prefixMessage(Log::MessageType::Info, message);
                                 ^
/private/miliu/project/opm/opm-parser/opm/parser/eclipse/OpmLog/OpmLog.cpp:42:52: error: ?Log? has not been declared
         const std::string msg = Log::prefixMessage(Log::MessageType::Info, message);
                                                    ^
/private/miliu/project/opm/opm-parser/opm/parser/eclipse/OpmLog/OpmLog.cpp:43:20: error: ?Log? has not been declared
         addMessage(Log::MessageType::Info, msg);
```

But Logger.hpp includes LogUtil.hpp, why build does not work?

PR #703, opened 2016-03-07
Include guards for Functional

jokva:


jokva:
I'll perform a cheeky self merge her, as this is a trivial (but important) patch.

PR #706, opened 2016-03-08
Parser performance improvements

jokva:
Depends on https://github.com/OPM/opm-parser/pull/705

This patch set addresses some of the performance shortcomings and inefficiencies in opm-parsers, mostly tracked down using valgrind (callgrind). Each individual commit addresses some specific aspect. A short summary:

* string_view is a std::string interface compatible (not completely, but on a by-need basis) view into a (sub)string to avoid copying too many strings.
* boost algorithms have been replaced by faster hand-rolled (locale-ignoring) versions.
* some containers take size hints and preallocates memory.
* some algorithmic improvements and heuristics applied.

### DIFF
--- a/opm/parser/eclipse/CMakeLists.txt
+++ b/opm/parser/eclipse/CMakeLists.txt
@@ -138,6 +138,7 @@ EclipseState/IOConfig/IOConfig.cpp)
 
 set( utility_source
 Utility/Functional.cpp
+Utility/Stringview.cpp
 )
 
 set( HEADER_FILES
@@ -287,9 +288,11 @@ EclipseState/Tables/ColumnSchema.hpp
 EclipseState/Tables/TableEnums.hpp
 EclipseState/Tables/TableSchema.hpp
 EclipseState/Tables/TableIndex.hpp
-Utility/Functional.hpp)
+#
+Utility/Functional.hpp
+Utility/Stringview.hpp)
 
-add_library(buildParser ${rawdeck_source} ${build_parser_source} ${deck_source} ${unit_source} ${generator_source})
+add_library(buildParser ${rawdeck_source} ${build_parser_source} ${deck_source} ${unit_source} ${generator_source} ${utility_source})
 target_link_libraries(buildParser opmjson ${Boost_LIBRARIES}  ${ERT_LIBRARIES})
 
 #-----------------------------------------------------------------
--- a/opm/parser/eclipse/Deck/DeckRecord.cpp
+++ b/opm/parser/eclipse/Deck/DeckRecord.cpp
@@ -29,6 +29,10 @@
 namespace Opm {
 
 
+    DeckRecord::DeckRecord( size_t size ) {
+        this->m_items.reserve( size );
+    }
+
     size_t DeckRecord::size() const {
         return m_items.size();
     }
--- a/opm/parser/eclipse/RawDeck/RawRecord.hpp
+++ b/opm/parser/eclipse/RawDeck/RawRecord.hpp
@@ -20,9 +20,12 @@
 #ifndef RECORD_HPP
 #define RECORD_HPP
 
-#include <string>
 #include <deque>
 #include <memory>
+#include <string>
+#include <vector>
+
+#include <opm/parser/eclipse/Utility/Stringview.hpp>
 
 namespace Opm {
 
@@ -34,32 +37,27 @@ namespace Opm {
     public:
         RawRecord(const std::string& singleRecordString, const std::string& fileName = "", const std::string& keywordName = "");
 
-        std::string pop_front();
+        string_view pop_front();
         void push_front(std::string token);
         size_t size() const;
 
         const std::string& getRecordString() const;
-        const std::string& getItem(size_t index) const;
+        string_view getItem(size_t index) const;
         const std::string& getFileName() const;
         const std::string& getKeywordName() const;
 
         static bool isTerminatedRecordString(const std::string& candidateRecordString);
-        virtual ~RawRecord();
-        void dump() const;
+
+       void dump() const;
 
     private:
         std::string m_sanitizedRecordString;
-        std::deque<std::string> m_recordItems;
+        std::deque< string_view > m_recordItems;
+        std::vector< std::string > expanded_items;
         const std::string m_fileName;
         const std::string m_keywordName;
 
         void setRecordString(const std::string& singleRecordString);
-        void splitSingleRecordString();
-        void processSeparatorCharacter(std::string& currentToken, const char& currentChar, char& tokenStarter);
-        void processQuoteCharacters(std::string& currentToken, const char& currentChar, char& tokenStarter);
-        void processNonSpecialCharacters(std::string& currentToken, const char& currentChar);
-        bool charIsSeparator(char candidate);
-        static unsigned int findTerminatingSlash(const std::string& singleRecordString);
     };
     typedef std::shared_ptr<RawRecord> RawRecordPtr;
     typedef std::shared_ptr<const RawRecord> RawRecordConstPtr;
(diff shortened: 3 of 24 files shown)


andlaus:
that's pretty cool! What's the performance improvement you see for real decks, though? (I'm thinking about Norne and "Model 2". A ballpark measure is enough I suppose.)

jokva:
On my machine, Norne is parsed in ~5.3s with master. With this patch set applied it takes ~1.8s.

andlaus:
Oh, only a speedup of factor 2.5? that's definitely not worth the trouble. (just kidding ;)

jokva:
I just checked Model 2. Master takes ~11.92s, this patch ~3.3s.

joakim-hove:
Wow - that was impressive!

jokva:
I'll rebase this on top of master on Monday (which will resolve the merge conflict). It can obviously be reviewed now, but if you want to test it you might have to rebase locally.

atgeirr:
Great work! Just a simple question: why is string_view needed, would it not be possible to use const string references (or pointers)? Also: could not string_view::substr() return string_view instances?

jokva:
>Great work! Just a simple question: why is string_view needed, would it not be possible to use const string references (or pointers)?

string_view is needed because the RawRecord (used to build deck) stores an entire input set in one string (whereas previously it stored  them in a deque of smaller strings copied out of the larger input string which was a massive performance killer), meaning we'd need to store the string ref + offset + size anyway, which is what string_view does.

It's certainly possible, but then it's implementing string_view anyway. 

>Also: could not string_view::substr() return string_view instances? If it did I would see greater benefit of it.

It could, but I chose to mimic std::string's interface in order to have all other code using substr "just work"™. If there is need for a string_view substring to view feature then substrview or substr_view is ok.

>I guess the name of the class is chosen to align with possible future standardized features, but I fear this may be more of a problem than a benefit.

Partially - ideally this implementation goes away once we move on to a standardised version. I don't like the UpperCaseNames at all, and especially not since this has nothing to do with opm or reservoir logic, but is strictly a memory/implementation/optimisation thing.

jokva:
I've rebased this onto master.

jokva:
I have now rebased this onto master.

acelster:
Great work, Joergen! Also nice to see you are still contributing to OPM.
Glad I got you involved. ;-) Anne

opm/parser/eclipse/CMakeLists.txt
```diff
@@ -287,9 +288,12 @@ EclipseState/Tables/ColumnSchema.hpp
 EclipseState/Tables/TableEnums.hpp
 EclipseState/Tables/TableSchema.hpp
 EclipseState/Tables/TableIndex.hpp
-Utility/Functional.hpp)
+#
+Utility/Functional.hpp
+Utility/Stringview.hpp
+Utility/EquilWrapper.hpp)
```
joakim-hove: Not really part of this PR; but can you move `EquilWrapper` to `EclipseState/InitConfig`?

opm/parser/eclipse/Deck/DeckRecord.cpp
```diff
@@ -29,6 +29,10 @@
 namespace Opm {
 
 
+    DeckRecord::DeckRecord( size_t size ) {
```
joakim-hove: It certainly makes a lot of sense to reserve the size of each individual `DeckItem` - but reserving the number of `DeckItem` instances in a `DeckRecord` probably does not make much sense performance wise?

opm/parser/eclipse/CMakeLists.txt
jokva: This should be gone and is a rebase bug. i'll get rid of it.

opm/parser/eclipse/Deck/DeckRecord.cpp
jokva: It's pretty significant actually, I seem to recall 8-10% improvement. Mostly because reallocating here is expensive because it invokes clone.

joakim-hove:
Maybe the `Rawxxxx` classes should be renamed `InputXxxx`?

jokva:
It feels slightly more natural I suppose, I have no strong objections. Not sure if it's worth the effort as I'm not sure how much more this part will be touched, but it can surely be done. It really isn't that clear that 'raw' largely means the tokenising phase.

joakim-hove:
Good - then I think we should do 's/Raw/Input/'.
11. mars 2016 12:42 skrev "jokva" <notifications@github.com> følgende:

opm/parser/eclipse/RawDeck/RawRecord.hpp
```diff
@@ -34,32 +37,27 @@ namespace Opm {
...
-        virtual ~RawRecord();
-        void dump() const;
+
+       void dump() const;
 
     private:
         std::string m_sanitizedRecordString;
-        std::deque<std::string> m_recordItems;
+        std::deque< string_view > m_recordItems;
+        std::vector< std::string > expanded_items;
```
joakim-hove: I guess these strings are only to ensure that views which are `push_front()`'ed have something to view? In that case e.g. `string_storage` would be a better name.

opm/parser/eclipse/RawDeck/RawRecord.hpp
```diff
@@ -34,32 +37,27 @@ namespace Opm {
...
 
-        std::string pop_front();
+        string_view pop_front();
         void push_front(std::string token);
         size_t size() const;
 
         const std::string& getRecordString() const;
-        const std::string& getItem(size_t index) const;
+        string_view getItem(size_t index) const;
         const std::string& getFileName() const;
```
joakim-hove: If moving the too-many-items handling from `ParserRecord::parse()` to `ParserKeyword::parse()` we can get rid of the `filename` and `keyword` members of the `RawRecord` class. Maybe not much performance to gain - but slightly cleaner.

opm/parser/eclipse/RawDeck/RawRecord.hpp
jokva: Kind-of. The expanded_items stores the physical strings generated by *-expansion, which isn't stored in the sanitised string (it actually is, we could push back the same view to the star value multiple times). The vector is always empathy in the abscence of * in the input file.

joakim-hove:
OK  - this is very good; except the minor rename+++ points mentioned I am ready to merge this.

jokva:
I'll look at this first thing Monday.

joakim-hove:
Just one thought: Would it be possible to re*use* the `RawRecord/RawKeyword` instance(s) - so that it is not necessary to reallocate the `std::vector<string_view>` member of the `RawRecord` for every record?

joakim-hove:
Just a quick question: The fixes are all in the `input -> Deck` stage - no fixes in the `Deck -> EclipseState`. Is that because that is where you focused your effort, or is it because the `Deck -> EclipseState` phase was anyway fast?

jokva:
> Just one thought: Would it be possible to reuse the RawRecord/RawKeyword instance(s) - so that it is not necessary to reallocate the std::vector<string_view> member of the RawRecord for every record?

I suppose you *can*, but I sincerely doubt it's worth it. Primarily for code clarity reasons (as we need the Raw* to stick around until their corresponding Deck* has been constructed), but also because the 1-allocation-per isn't that big of a deal. Still something and it might be worth investigating surely.

The bigger costs came from repeated allocation or resizing. With a shared instance we'd probably let vector or deque reallocate itself until it reaches the longest record or item was reached, then reuse the memory. Since a lot of records are really small they don't really trigger any *re-allocations* so they can be considered cheap (no copying or moving of elements). I can think a bit about it, but my gut feeling tells me it's hardly any benefit in this.

> Just a quick question: The fixes are all in the input -> Deck stage - no fixes in the Deck -> EclipseState. Is that because that is where you focused your effort, or is it because the Deck -> EclipseState phase was anyway fast?

When I started, Input -> Deck accounted for approx. 90% of the run time, and I found the pain points where part-of-execution-time was too high (i.e. anything that isn't directly related to either AST management or string-to-int/double conversion. Now that number is closer to 80%, and while I'm sure there are plenty of inefficiencies in Deck -> EclipseState it simply hasn't been that much to gain from improving it yet.

PR #707, opened 2016-03-08
Add new keyword TLPMIXPA

totto82:
The TLPMIXPA keyword can be used for giving pressure depended Todd-
Longstaff parameters. The implementation follows the description of
PMISC.

### DIFF
--- a/opm/parser/eclipse/CMakeLists.txt
+++ b/opm/parser/eclipse/CMakeLists.txt
@@ -280,6 +280,7 @@ EclipseState/Tables/SorwmisTable.hpp
 EclipseState/Tables/SgcwmisTable.hpp
 EclipseState/Tables/MiscTable.hpp
 EclipseState/Tables/PmiscTable.hpp
+EclipseState/Tables/TlpmixpaTable.hpp
 EclipseState/Tables/MsfnTable.hpp
 EclipseState/Tables/TableColumn.hpp
 EclipseState/Tables/ColumnSchema.hpp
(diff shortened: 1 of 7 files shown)


joakim-hove:
This looks fine, technically I am happy to merge it.

But I would like a 'political' discussion, with this PR a 'optional' Keyword suddenly becomes a hard requirement to build flow, so at the the least I would like some other opinions from e.g. @alfbr  or @atgeirr.

As I see it we can just merge it, or we can devise a system for conditional compilation - I would prefer the latter.

joakim-hove:
Further suggestion: I have wanted the tables to be json specified - then this would be trivial, but that is of course a quite large task.

atgeirr:
> with this PR a 'optional' Keyword suddenly becomes a hard requirement to build flow

I don't see the problem, could you explain why this can be a bad thing?

alfbr:
Sorry for slow reply. I have thought about it, but no strong opinions either way. I see custom keywords as a slippery slope, so ideally we should be able to easily isolate the parser from them, such as disabling them at compile time. However, I do not see a particular use case for this in the intermediate future, so I am inclined to suggest that we cross that bridge when we get there. Right now, we would take on an implementation and maintenance burden to support compile time disablement. Hence, I suggest we go the easy route (just build support for all keywords for the time being), unless we see that it will be very expensive to reverse that decision later on. I do think the parser should produce a message to warn clients (whether it be simulators or viewers) of custom keyword usage though.

joakim-hove:
Thanks for opinions - I'll merge.

PR #709, opened 2016-03-12
EclipseState: remove deprecated has{Int,Double}GridProperty() methods

andlaus:
these methods have been renamed to hasDeck{Int,Double}GridProperty().
my reason why I kept the deprecated names initially was to avoid a
"hard" break for the downstream OPM modules. since all of these
modules have since been adapted to use the new names, there's IMO no
reason to keep the deprecated methods.

(Note that since we currently do *not* keep compatibility between
releases, the deprecated has*GridProperty() methods can be removed
immediately instead of waiting until after the next release.)

### DIFF
--- a/opm/parser/eclipse/EclipseState/EclipseState.hpp
+++ b/opm/parser/eclipse/EclipseState/EclipseState.hpp
@@ -81,11 +81,6 @@ namespace Opm {
         bool hasDeckIntGridProperty(const std::string& keyword) const;
         bool hasDeckDoubleGridProperty(const std::string& keyword) const;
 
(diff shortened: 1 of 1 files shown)


jokva:
The Properties on the EclipseState are receiving much love currently too, but I'm pretty confident this can be merged separately.

andlaus:
I don't mind if these methods are removed via a different PR. (or if they are not removed at all, actually.) Thus, go ahead as you prefer.

joakim-hove:
We certainly want these changes in - but as I know they are already part of ongoing churn I will close this PR now.

PR #711, opened 2016-03-14
Add EclipsePRTLog backend.

qilicun:
This PR intend to add a prototype for output eclipse PRT file, more usefule functions could be added later.

### DIFF
--- a/opm/parser/eclipse/OpmLog/EclipsePRTLog.cpp
+++ b/opm/parser/eclipse/OpmLog/EclipsePRTLog.cpp
@@ -68,12 +68,12 @@ namespace Opm {
 
         //output summary.
         const std::string summary_msg = "\n\nError summary:" + 
-            std::string("\nWarnings          " + warning) +
-            std::string("\nProblems          " + problem) +
-            std::string("\nErrors            " + error) + 
-            std::string("\nBugs              " + bug) + 
-            std::string("\nDebug             " + debug) +
-            std::string("\nProblems          " + problem) +"\n";
+            std::string("\nWarnings          " + std::to_string(warning)) +
+            std::string("\nProblems          " + std::to_string(problem)) +
+            std::string("\nErrors            " + std::to_string(error)) + 
+            std::string("\nBugs              " + std::to_string(bug)) + 
+            std::string("\nDebug             " + std::to_string(debug)) +
+            std::string("\nProblems          " + std::to_string(problem)) +"\n";
         addMessage(Log::MessageType::Info, summary_msg);
     }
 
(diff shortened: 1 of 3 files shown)


opm/parser/eclipse/OpmLog/EclipsePRTLog.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+    void EclipsePRTLog::clear()
+    {
+        m_count.clear();
+    }
+
+
+    EclipsePRTLog::~EclipsePRTLog()
+    {
+        const auto warning = numMessages(Log::MessageType::Warning);
```
joakim-hove: Not really necessary with these temporary variables?

qilicun: True, will remove them.

opm/parser/eclipse/OpmLog/EclipsePRTLog.cpp
```diff
@@ -0,0 +1,80 @@
...
+            if (iter == m_count.end())
+                return 0;
+            else
+                return (*iter).second;
+        } else
+            throw std::invalid_argument("The messageType ID must be 2^n");
+    }
+
+
+    void EclipsePRTLog::clear()
```
joakim-hove: This method seems like a *might be handy* - method. I have strongly come to the conclusion that such methods should be avoided - *implement it when you need it.*

opm/parser/eclipse/OpmLog/EclipsePRTLog.cpp
```diff
@@ -0,0 +1,80 @@
...
+*/
+
+#include <opm/parser/eclipse/OpmLog/EclipsePRTLog.hpp>
+#include <opm/parser/eclipse/OpmLog/LogUtil.hpp>
+
+namespace Opm {
+
+
+    EclipsePRTLog::EclipsePRTLog(const std::string& logFile, int64_t messageMask) : StreamLog(logFile, messageMask)
+    {}
```
joakim-hove: As long as these constructors only call the base constructor you do not need them?

qilicun: Does C++ will call the base constructor when you want to create an EclipsePRTLog object?

opm/parser/eclipse/OpmLog/EclipsePRTLog.cpp
qilicun: Ok.

opm/parser/eclipse/OpmLog/EclipsePRTLog.cpp
atgeirr: Inheriting constructors is a fairly new feature in C++ though, and you must explicitly declare them I think.

jokva: You can delegate to the parent constructor with

`class Foo : public Bar { public: using Bar::Bar; };`

qilicun: Good solution.

joakim-hove:
Thank you.

PR #712, opened 2016-03-14
simply replace throw with OpmLog functions.

qilicun:
This PR is the first attempt to  replace throw with OpmLog functions, starts from Deck folder.
In principle, all the ``throw`` need to be replaced, due to different error type, they will be replaced with different OpmLog functions, such as error and debug. There is no clear rule to distinguish the category, but the simple rule is if the messages caused by user (Input deck), that will be an error ,warning, if it caused by the developer(such as missing parameter, wrong function names), that will be debug. 

Any comments are welcome!

### DIFF
--- a/opm/parser/eclipse/Deck/Deck.cpp
+++ b/opm/parser/eclipse/Deck/Deck.cpp
@@ -24,6 +24,7 @@
 #include <opm/parser/eclipse/Deck/DeckKeyword.hpp>
 #include <opm/parser/eclipse/Deck/Section.hpp>
 #include <opm/parser/eclipse/Units/UnitSystem.hpp>
+#include <opm/parser/eclipse/OpmLog/OpmLog.hpp>
 
 namespace Opm {
 
@@ -44,21 +45,21 @@ namespace Opm {
 
     const DeckKeyword& DeckView::getKeyword( const std::string& keyword, size_t index ) const {
         if( !this->hasKeyword( keyword ) )
-            throw std::invalid_argument("Keyword " + keyword + " not in deck.");
+            OpmLog::error("Keyword " + keyword + " not in deck.");
 
         return this->getKeyword( this->offsets( keyword ).at( index ) );
     }
 
     const DeckKeyword& DeckView::getKeyword( const std::string& keyword ) const {
         if( !this->hasKeyword( keyword ) )
-            throw std::invalid_argument("Keyword " + keyword + " not in deck.");
+            OpmLog::error("Keyword " + keyword + " not in deck.");
 
         return this->getKeyword( this->offsets( keyword ).back() );
     }
 
     const DeckKeyword& DeckView::getKeyword( size_t index ) const {
         if( index >= this->size() )
-            throw std::out_of_range("Keyword index " + std::to_string( index ) + " is out of range.");
+            OpmLog::debug("Keyword index " + std::to_string( index ) + " is out of range.");
 
         return *( this->begin() + index );
     }
--- a/opm/parser/eclipse/Deck/DeckItem.cpp
+++ b/opm/parser/eclipse/Deck/DeckItem.cpp
@@ -19,6 +19,7 @@
 
 #include <opm/parser/eclipse/Deck/DeckItem.hpp>
 #include <opm/parser/eclipse/Units/Dimension.hpp>
+#include <opm/parser/eclipse/OpmLog/OpmLog.hpp>
 
 #include <boost/algorithm/string.hpp>
 
@@ -110,7 +111,7 @@ namespace Opm {
     template< typename T >
     void DeckTypeItem< T >::push_back( T x ) {
         if( this->dataPointDefaulted.size() != this->data.size() )
-            throw std::logic_error("To add a value to an item, no \"pseudo defaults\" can be added before");
+            OpmLog::debug("To add a value to an item, no \"pseudo defaults\" can be added before");
 
         this->data.push_back( x );
         this->dataPointDefaulted.push_back( false );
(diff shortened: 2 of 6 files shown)


jokva:
This doesn't work - get returns a *reference* and this way we risk returning a reference to the wrong item (or invoke undefined behaviour). We don't guarantee correctness trying to look up a keyword that doesn't exist in the Deck, so simply logging and error and moving on is not acceptable.

The throw is there to provide an escape hatch (in almost all cases simply crash the program) and cannot be replaced without either introducing an optional type or fall back onto using pointers.

joakim-hove:
No sorry - this will not be merged. 

I agree that throwing an exception is a brutal form of error handling; but the current PR will just let the application continue in an undefined/incorrect state - that is *much worse*. The current PR is focused on the `Deck` family of classes, and exception behavior in those classes. These classes are *semi internal* - the ambition is that calling scope like e.g. `flow` should only use the `EclipseState` class; hence when e.g. `Deck::getKeyword( )` method throws that is close to an internal bug in the opm-parser library, and the fix should be applied in e.g. the `EclipseState` construction of the opm-parser code. 

When calling scope has the statement:
```C++
auto keyword = deck.getKeyword("PERMX" , 0);
```
you expect the `keyword` variable to correspond to the first `PERMX` keyword of the deck. If the deck does not contain `PERMX` you will now have no idea what the `keyword` variable contains - the program will  at the very least give wrong and unexpected behavior - and most certainly it will die a horrible death at a later stage. If throw is not acceptable it is the responsability of calling scope to check before calling `getKeyword`:
```C++
if (deck.count("PERMX") > 0) {
   auto keyword = deck.getKeyword("PERMX" , 0);
   ...
}  else  
   // Missing keyword 'PERMX'
```

If you have concrete call sites where the aggressive throw behavior creates problems, please point them out and we should take it from there. But in general I assume the fix will mostly be of the type illustrated above - i.e. start by checking first.

qilicun:
@jokva  @joakim-hove 

I see your point and concern, I admit that simply replacing throw is somewhat unacceptable and also I agree that if some horrible problems happen, the program needs to stop, but if the error is data file related,  from the user point of view(user means the person who actually use flow to run simulation) I think it's no need to stop it immediately,  do you think it's ok to delay the throwing  manner after diagnosing the whole input data file when errors are detected?

>you expect the keyword variable to correspond to the first PERMX keyword of the deck. If the deck does not contain PERMX you will now have no idea what the keyword variable contains - the program will at the very least give wrong and unexpected behavior 

If the deck has no PERMX, this is a deck file related error, I think it needs to be recorded in the log file and marked as error, if errors are detected, the program will stop in a later stage(preferablly after diagnosing the whole input data file), which also means the unexpected behavior inside the source code does not matter at all.

Maybe it's not suitable to change all the throw(like this PR), but I think it's better at least for some of them.

alfbr:
Whether we want to throw at missing PERMX depends on usage. For flow we actually may want to throw, and catch it for the logger. However, this PR is a no-go for two reasons. One is that we do not want the logger in parser at all, it should be moved to opm-common. In the parser we should have a message container filled up with messages the logger can log from. The second reason is that we want to preserve strict reading by default. Throws will need to be handled with the parsemode implementation. Please have a look at the existing parsemode examples.

qilicun:
> One is that we do not want the logger in parser at all, it should be moved to opm-common. In the parser we should have a message container filled up with messages the logger can log from. 

Ok, I'm fine with this.
@joakim-hove I'm not sure if you have any plan to implement it, if not, then I can do it.  For the message container, I think it  should be a singleton just like OpmLog.

joakim-hove:
> I'm not I'm not sure if you have any plan to implement it, if not, then I can do it. For the message container, I think it should be a singleton just like OpmLog.

It is on our list of tasks; but we are not ready to start right away - so feel free to start on this task:

1. I really do *not* want a singleton - I felt quite dirty with the singleton implementation of `OpmLog` and don't want that to become a pattern in Opm.
2. Please implement with value semantics and not `std::shared_ptr` we are trying to cut down on the shared_ptr use.

I will send you a private email.

qilicun:
@joakim-hove 

Does that means that we need to let the message container be the one of the parameter for the functions and be of the the data member for the classes?

joakim-hove:
> Does that means that we need to let the message container be the one of the parameter for the functions

Yes - the uglyness of this is that the myriads of callsites must be updated to add a `MessageContainer&` argument. One possibility could be to:

1. Rename `ParseMode -> ParseContext`
2. Add the messages to the `ParseContext` class

Then the method signatures could stay the same, modulo the `const ParseMode& -> ParseContext` change.

 > and be of the the data member for the classes?

Well; as mentioned it could maybe be joined with `ParseMode` - alternatively for convenience a *reference* could be held by the `ParserState` struct: https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/Parser/Parser.cpp#L49 - but ultimately the calling scope - e.g. `flow` should be the owner of the message container:

```C++
ParseMode parseMode;
MessageContainer messages;

auto deck = parser->parseFile( deckfile , parseMode , messages );
auto state = EclipseState( deck , parseMode , messages );

// Go through the messages ....
```

atgeirr:
> One possibility could be to:
> 
> Rename ParseMode -> ParseContext
> Add the messages to the ParseContext class
> Then the method signatures could stay the same, modulo the const ParseMode& -> ParseContext change.

I like that idea, it seems less intrusive than the alternatives.

qilicun:
>I like that idea, it seems less intrusive than the alternatives.

Me too, this is a time saver.

Ok, if no one reject, I will start with the renaming.

qilicun:
@joakim-hove @atgeirr 

I have already finish the renaming, if you have no more suggestion on the current "ParseContext" implementation, I will make the PRs.

joakim-hove:
Go ahead; feel free to make PR with only the rename.

PR #719, opened 2016-03-17
Rename parsemode

qilicun:
As discussed in https://github.com/OPM/opm-parser/pull/712


qilicun:
Will fix this issue

qilicun:
Finally， travis feels happy!
But the strange thing is that my local build with the latest opm-common will never build opm-parser/python folder. Does anyone know why?

joakim-hove:
> But the strange thing is that my local build with the latest opm-common will never build opm-parser/python folder

The python code is not built by default; you must pass `-DENABLE_PYTHON=ON` during the configure step.

qilicun:
>you must pass -DENABLE_PYTHON=ON during the configure step.

Thanks for noticing.

joakim-hove:
OK - good; this was a massive search-replace effort. I'll run around pressing green buttons.

PR #715, opened 2016-03-14
Refactor satfunc init

pgdr:
Refactor, removed use of deck in SatfuncInitializers, moved to TableManager

PR #720, opened 2016-03-17
Query the name of (FIP) regions in State

jokva:
Enable support for querying what fluid-in-point regions are defined in
the deck. For now the only interesting regions are the ones specified by
the FIPNUM keyword in the REGIONS section.

The implementation is somewhat inefficient (n log n over the number of
cells in the grid), but since this is likely to only be called when
default-expanding SUMMARY section keywords this isn't too bad, since the
implementation is so dead simple. 

--

This still need some good test cases before I'm comfortable merging it. I'm also not sure if this belongs directly onto the EclipseState or if it is a grid property. Opinions very welcome.

### DIFF
--- a/opm/parser/eclipse/EclipseState/EclipseState.cpp
+++ b/opm/parser/eclipse/EclipseState/EclipseState.cpp
@@ -17,6 +17,8 @@
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <set>
+
 #include <boost/algorithm/string/join.hpp>
 
 #include <opm/common/OpmLog/OpmLog.hpp>
@@ -242,6 +244,18 @@ namespace Opm {
 
         m_transMult->setMultregtScanner( scanner );
     }
+
+    std::vector< int > EclipseState::getRegions( const std::string& kw ) const {
+        if( !this->get3DProperties().hasDeckIntGridProperty( kw ) ) return {};
+
+        const auto& property = this->get3DProperties().getIntGridProperty( kw );
+
+        std::set< int > regions( property.getData().begin(),
+                                 property.getData().end() );
+
+        return { regions.begin(), regions.end() };
+    }
+
 
     void EclipseState::complainAboutAmbiguousKeyword(DeckConstPtr deck, const std::string& keywordName) const {
         OpmLog::addMessage(Log::MessageType::Error, "The " + keywordName + " keyword must be unique in the deck. Ignoring all!");
--- a/opm/parser/eclipse/EclipseState/EclipseState.hpp
+++ b/opm/parser/eclipse/EclipseState/EclipseState.hpp
@@ -20,9 +20,10 @@
 #ifndef OPM_ECLIPSE_STATE_HPP
 #define OPM_ECLIPSE_STATE_HPP
 
-#include <utility>
 #include <memory>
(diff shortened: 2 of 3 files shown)


joakim-hove:
Some comments:

1. We know actually have to quite significant GridProperty changesets in flight: https://github.com/OPM/opm-parser/pull/715 and https://github.com/OPM/opm-parser/pull/695 - I would like to get them merged before merging this.

2. In my opinion it should be part of the GridProperties container.

3. We can live with the `n log n`Implementation; but maybe add a comment so we do not risk a situation like:
```C++
while (??) {
    auto regions = getRegions( );
    ...
}
in calling scope.

File: opm/parser/eclipse/EclipseState/EclipseState.cpp [hunk 1]
```diff
@@ -412,8 +414,16 @@ namespace Opm {
         }
     }
 
+    std::vector< int > EclipseState::getRegions() const {
```
joakim-hove: This should in my opinion take  a `const std::string& regionName` input. The `Equil` initiialisation in opm-core probably already has a "similar" implementation based on `EQLNUM`

jokva:
> We know actually have to quite significant GridProperty changesets in flight: #715 and #695 - I would like to get them merged before merging this.

Another reason for this being a WIP.

> In my opinion it should be part of the GridProperties container.
I'm ok with this.

>  We can live with the n log nImplementation; but maybe add a comment so we do not risk a situation like

Can do. We can cache the result too so that it's O(1) for every subsequent call, but that introduces complexity in the sense that we must maintain another vector in properties/state.

opm/parser/eclipse/EclipseState/EclipseState.cpp
jokva: Sure. I don't know if region identifiers are pure integers or not, but Norne only uses ints so I figured it was a reasonable assumption. If it's not then I'm happy to change it.

joakim-hove:
@jokva: got an e-mail address for the day?

opm/parser/eclipse/EclipseState/EclipseState.cpp
joakim-hove: > for wells it may not be so hard -- is it more than well names that are required?

The identifiers are pure integers.

akva2:

joakim-hove:
Now that the large gridproperties patch has been merged I would be happy to merge this; but right now it is not green.

jokva:
On it.

opm/parser/eclipse/EclipseState/EclipseState.hpp
```diff
@@ -88,9 +89,11 @@ namespace Opm {
 
         const TableManager& getTableManager() const;
 
-        // the unit system used by the deck. note that it is rarely needed to
-        // convert units because internally to opm-parser everything is
-        // represented by SI units...
+        std::vector< int > getRegions( const std::string& kw = "FIPNUM" ) const;
```
joakim-hove: Please leave out the default keyword - there is already a context aware `defaultRegion()` in the `EclipseState`.

joakim-hove:
OK - I'll merge when there is a test!

jokva:
A simple test is up & running.

PR #727, opened 2016-03-22
Added 3 keywords

kjellkongsvik:
AQUDIMS
MEMORY
PARALLEL


opm/parser/share/keywords/000_Eclipse100/M/MEMORY
```diff
@@ -0,0 +1,5 @@
+{"name" : "MEMORY" , "sections" : ["SOLUTION"],
```
joakim-hove: I would guess this also should be in the `RUNSPEC`section?

kjellkongsvik: No, it should only be in RUNSPEC, updated PR. Thanks.

PR #728, opened 2016-03-23
Various improvements

jokva:
A handful of minor performance enhancements and small refactors. Doesn't affect anything externally but runtime.

### DIFF
--- a/opm/parser/eclipse/Deck/DeckItem.cpp
+++ b/opm/parser/eclipse/Deck/DeckItem.cpp
@@ -82,8 +82,37 @@ namespace Opm {
             friend class DeckItem;
     };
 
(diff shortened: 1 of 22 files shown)


atgeirr:
This PR broke many tests for me in opm-parser. Since some of the tests produce changing garbage, I suspect some iterator or string_view is invalid. It seems to trigger upon reading the first record containing doubles, though that may be misleading (since it may not be the only place it fails).

atgeirr:
To follow up on this: we have identified the bug, and @jokva is working on a fix.

jokva:
These issues are fixed in https://github.com/OPM/opm-parser/pull/737

PR #729, opened 2016-03-24
Add message container

qilicun:
This message container is used to replace OpmLog inside opm-parser in the near future.

Feel free to make comments.


opm/parser/eclipse/Parser/MessageContainer.hpp
```diff
@@ -0,0 +1,68 @@
...
+        MessageType mtype;
+        std::string message;
+        std::unique_ptr<Location> location;
+    };
+
+
+    ///Message container is used to replace OpmLog functionalities.
+    class MessageContainer {
+    public:
+        void error(const std::string& msg, const std::string& filename, const int lineno);
```
joakim-hove: Please add overloads like:
```C++
void error(const std::string& msg);
```
i.e. *without* the `filename` and `lineno` arguments - the whole point of the funny `struct Location` and `std::unique_ptr<Location>`was to facilitate storing messages with no associated line number - this will typically arise from the `EclipseState` construction phase where we have left the input files behind.

opm/parser/eclipse/Parser/MessageType.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+#ifndef MESSAGETYPE_H
+#define MESSAGETYPE_H
+
+#include <cstdint>
+#include <string>
+
+namespace Opm {
+
+    enum MessageType {
```
joakim-hove: Please add an extra namespace for the `MessageType` enum; as the code stands now `Debug` will be a valid symbol in the `Opm` namespace. See e.g. https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Schedule/ScheduleEnums.hpp

opm/parser/eclipse/Parser/MessageType.hpp
```diff
@@ -0,0 +1,41 @@
...
+        Debug     = 1,
+        Info      = 2,
+        Warning   = 4,
+        Error     = 8,
+        Problem   = 16,
+        Bug       = 32
+    };
+    const int64_t DefaultMessageTypes = MessageType::Debug + MessageType::Info + MessageType::Warning + MessageType::Error + MessageType::Problem + MessageType::Bug;
```
joakim-hove: This is used as a mask in the `OpmLog` implementation  - I do not think it is needed here?

opm/parser/eclipse/Parser/MessageType.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Opm {
+
+    enum MessageType {
+        Debug     = 1,
+        Info      = 2,
+        Warning   = 4,
+        Error     = 8,
+        Problem   = 16,
+        Bug       = 32
```
joakim-hove: I feel quite uneasy about these numerical values. An important point of the whole `MessageContainer` implementation is to make certain that the parser can be built and deployed without using the C++ code in `opm-common` - that implies that the message types defined here must be independent of the message types defined in `opm-common` - and the translation must be implemented in the calling scope e.g. `flow`.

To really emphasize that these message types *are* independent of those in `OpmLog` I would suggest:

1. Using *different* numerical values.
2. A comment describing the fact.

opm/parser/eclipse/Parser/MessageContainer.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+        void info(const std::string& msg, const std::string& filename, const int lineno);
+
+        void debug(const std::string& msg, const std::string& filename, const int lineno);
+
+        void problem(const std::string& msg, const std::string& filename, const int lineno);
+
+
+    private:
+        std::vector<Message> m_message;
```
joakim-hove: I guess this could be `<const Message>`?

opm/parser/eclipse/Parser/MessageContainer.hpp
```diff
@@ -0,0 +1,68 @@
...
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef MESSAGECONTAINER_H
+#define MESSAGECONTAINER_H
+
+#endif // OPM_MESSAGECONTAINER_HEADER_INCLUDED
+
+#include <opm/parser/eclipse/Parser/MessageType.hpp>
```
joakim-hove: I guess the `MessageType` definition could just be in this file?

opm/parser/eclipse/Parser/MessageContainer.hpp
```diff
@@ -0,0 +1,68 @@
...
+        void bug(const std::string& msg, const std::string& filename, const int lineno);
+
+        void warning(const std::string& msg, const std::string& filename, const int lineno);
+
+        void info(const std::string& msg, const std::string& filename, const int lineno);
+
+        void debug(const std::string& msg, const std::string& filename, const int lineno);
+
+        void problem(const std::string& msg, const std::string& filename, const int lineno);
+
```
joakim-hove: Please provide `begin()` and `end()` const iterators:
```
std::vector<const Message>::const_iterator MessageContainer::begin() const {
    return m_messages.begin()
}
```

opm/parser/eclipse/Parser/MessageContainer.hpp
joakim-hove: Variable should be named `m_messages` - i.e. plural.

joakim-hove:
Where is the test?

joakim-hove:
OK good - apart from the small(?) comments - and the missing test[1] I am ready to merge this.

[1]: NB: The test should demonstrate that range based for works!

opm/parser/eclipse/Parser/MessageContainer.hpp
jokva: No, you can't (reasonably) have a container of const elements - every slot in the container must be assignable, and constness disallows that.

joakim-hove: > No, you can't (reasonably) have a container of const elements

Good - thank you; I had a feeling it *might* be like that - now I know.

qilicun:
This PR was updated according to the comments.

opm/parser/eclipse/Parser/tests/MessageContainerTest.cpp
```diff
@@ -0,0 +1,41 @@
...
+    
+    MessageContainer msgContainer;
+    msgContainer.error("This is an error.");
+    msgContainer.bug("This is a bug.", "dummy.log", 20);
+    {       
+        BOOST_CHECK_EQUAL("This is an error.", msgContainer.begin()->message);
+        BOOST_CHECK_EQUAL("dummy.log", (msgContainer.end()-1)->location->filename);
+        BOOST_CHECK_EQUAL(20, (msgContainer.end()-1)->location->lineno);
+    }
+
```
joakim-hove: OK; now I just want a range based for demonstration:
```C++
std::vector<Message> msgList;
for (const auto& msg : msgContainer) {
    msgList.push_back( msg );
}

BOOST_CHECK_EQUAL_COLLECTIONS( msgList.begin() , msgList.end() , msgContainer.begin());
```

joakim-hove:
This is good; when the test has been extended with range based for I will merge it.

opm/parser/eclipse/Parser/tests/MessageContainerTest.cpp
qilicun: >BOOST_CHECK_EQUAL_COLLECTIONS( msgList.begin() , msgList.end() , msgContainer.begin());

I guess this needs the overload implementation of ``==`` and ``!=``, right?

joakim-hove: > I guess this needs the overload implementation of == and !=, right?

Well; my initial guess was that the default operator== would do; but I guess the `std::unique_ptr<>` creates problems? Come to think of it - that probably creates problems also for the `msgList.push_back( )` - the `std::shared_ptr<T>` does have *some* charm!

Anyway - the important point was that `for( const auto& msg : msgContainer)` should work, the `msgList` was just a (broken ...) way to test it. If you instead test it with:
```C++

std::vector<string> msgList; 
for (const auto& msg : msgContainer) {
    msgList.push_back( msg.message );
}

for (size_t index = 0; index msgList.size(); index++) {
     // requires operator[] - which is a good thing I guess + MessageContainer.size() ??
    const auto& msg = messageContainer[index]; 
    BOOST_CHECK_EQUAL( msgList[index] , msg.message );
}
```

I'm going offline now ;-)

qilicun: >Well; my initial guess was that the default operator== would do; but I guess the std::unique_ptr<> creates problems? Come to think of it - that probably creates problems also for the msgList.push_back( ) - the std::shared_ptr<T> does have some charm!

Haha, yes!

> I'm going offline now ;-)
￼
No problem, I will check this.

jokva: No, it most certainly does not require that. You don't have to support checking two elements for equality for range support. Just don't bother checking for equality (it's somewhat nonsensical for messages anyway).

shared_ptr wouldn't have made this better either, as it's not a smart pointer issue.

push_back is not a problem either, since the elements are being created properly by the error/info/debug [...] methods.

qilicun: I agree @jokva .
If you want to compare ``std::vector<Message> msgList`` with ``msgContainer``, I think the ``push_back(msgContainer[index].message)`` will not work, since they are totally different. And also in ``MessageContainer`` class there is no function like ``getMessage()``, that also means ``msgContainer[index]`` goes wrong.

jokva: random access in the message list isn't necessarily interesting though, is it? Typically you'd just iterate over all messages anyway, making the feature somewhat redundant.

joakim-hove: The stuff about `operator==` and `std::uique_ptr<>` versus `std::shared_ptr<>` was all abut the not-very-well-thought-out possible example of test code; not the `MessageContainer` itself.

Whatever - just make sure the range based for is exercised in a test; I agree random access is not necessary.

qilicun:
Ok, I add some more tests, @joakim-hove Does this look good?

qilicun:
@joakim-hove 
The reason that why I did not try to test rang based for loop is that I thought ``msgContainer`` is a class object, and the class itself does not have any function to get the ``std::vector<Message>``, so  ``for (const auto& msg : msgContainer)`` will lead to an error, but I works eventually, what a magic thing!

jokva:
@qilicun for-based loop looks for the methods begin() and end() and expects them to return iterators. It won't work unless those two are defined, but since you now have implemented them `for` becomes valid.

qilicun:
@jokva Good to know, thanks！

opm/parser/eclipse/Parser/MessageContainer.hpp
```diff
@@ -0,0 +1,85 @@
...
+#include <string>
+#include <vector>
+#include <memory>
+
+namespace Opm {
+
+    namespace MessageType {
+
+        enum MessageTypeEnum {
+            DEBUG     = 1,
```
bska: This particular name (`DEBUG`) conflicts with the [macro](https://github.com/OPM/opm-parser/blob/master/CMakeLists.txt#L82) that is defined in the top-level `CMakeLists.txt` in debug name.  The proper resolution is **probably** to not define the `DEBUG` macro in the build system.

As of right now, module opm-parser does not build in debug mode due to this conflict.

atgeirr: > The proper resolution is probably to not define the DEBUG macro in the build system.

I agree, but I think we should *also* avoid the DEBUG name in code. I think using Debug, Info etc. is more in line with naming practice elsewhere in OPM.

bska: Update: We also define `DEBUG` in opm-common's [`UseOptimization.cmake`](https://github.com/OPM/opm-common/blob/master/cmake/Modules/UseOptimization.cmake#L44) (in the context of debug builds).  We probably should not do that, but it certainly makes the name `DEBUG` unusable as an identifier in our C++ (and C) code at the moment.

In other words, no OPM modules (except opm-verteq) are currently buildable in debug mode.

bska: Update: Fixed in Opm/opm-parser#733

PR #730, opened 2016-03-27
First output step

joakim-hove:
Interernalize the first output step where RFT / Restart output is requested. In particular for RFT this information will be used to speed up the RFT output.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/DynamicState.hpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/DynamicState.hpp
@@ -143,6 +143,23 @@ namespace Opm {
         }
 
 
+        /// Will return the index of the first occurence of @value, or
+        /// -1 if @value is not found.
+        ssize_t find(const T& value) {
+            auto iter = std::find( m_data.begin() , m_data.end() , value);
+            if (iter != m_data.end()) {
+                // Found normally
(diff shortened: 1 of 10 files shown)


opm/parser/eclipse/EclipseState/Schedule/DynamicState.hpp
```diff
@@ -143,6 +143,22 @@ namespace Opm {
...
+                    if (value == m_currentValue)
+                        return 0;
+                    else
+                        return -1;
+                } else
+                    return -1;
+            } else
+                return std::distance( m_data.begin() , iter );
+        }
+
```
jokva: I suppose it is a matter of preference, but I personally feel this style makes it much harder to follow the control flow. Instead, something along the lines of:

```
auto iter = std::find( m_data.begin(), m_data.end(), value );
if( iter != m_data.end() ) return std::distance( m_data.begin(), iter );

/* now the common case is handled and we deal with the edge cases */
if( m_data.size() == 0 && value == m_currentValue ) return 0;

return -1;

```

joakim-hove: > I suppose it is a matter of preference

Yes - and this time we agree; I have updated the branch according to your suggestion.

PR #734, opened 2016-03-29
New summary keywords

jokva:
Internalise and generate `smspec_node` for the Region and Completion family of keywords.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Summary/Summary.cpp
+++ b/opm/parser/eclipse/EclipseState/Summary/Summary.cpp
@@ -91,12 +91,12 @@ namespace Opm {
     }
 
     static inline std::array< int, 3 > getijk( const DeckRecord& record,
-                                               int init = 0 )
+                                               int offset = 0 )
     {
         return {
-            record.getItem( init + 0 ).get< int >( 0 ) - 1,
-            record.getItem( init + 1 ).get< int >( 0 ) - 1,
-            record.getItem( init + 2 ).get< int >( 0 ) - 1
+            record.getItem( offset + 0 ).get< int >( 0 ) - 1,
+            record.getItem( offset + 1 ).get< int >( 0 ) - 1,
+            record.getItem( offset + 2 ).get< int >( 0 ) - 1
         };
     }
 
--- a/opm/parser/eclipse/EclipseState/Summary/Summary.cpp
+++ b/opm/parser/eclipse/EclipseState/Summary/Summary.cpp
@@ -148,9 +148,7 @@ namespace Opm {
 
        for( const auto& record : keyword ) {
 
-           const auto& wellname = record.getItem( 0 ).get< std::string >( 0 );
-
-           if( wellname == "*" ) {
+           if( record.getItem( 0 ).defaultApplied( 0 ) ) {
                for( const auto& well : schedule->getWells() ) {
 
                    const auto& name = wellName( well );
@@ -159,7 +157,7 @@ namespace Opm {
                        auto cijk = getijk( *completion );
 
                        /* well defaulted, block coordinates defaulted */
-                       if( record.size() != 4 ) {
+                       if( record.getItem( 1 ).defaultApplied( 0 ) ) {
                            nodes.emplace_back( keywordstring, name, dims.data(), cijk.data() );
                        }
                        /* well defaulted, block coordinates specified */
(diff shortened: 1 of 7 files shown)


jokva:
Depends on an ert PR as well as https://github.com/OPM/opm-parser/pull/720

akva2:

opm/parser/eclipse/EclipseState/Summary/Summary.cpp
```diff
@@ -79,30 +82,114 @@ namespace Opm {
...
 
+    static inline std::array< int, 3 > dimensions( const EclipseGrid& grid ) {
+        return {
+            int( grid.getNX() ),
+            int( grid.getNY() ),
+            int( grid.getNZ() )
+        };
+    }
+
+    static inline std::array< int, 3 > getijk( const DeckRecord& record,
```
joakim-hove: It looks like this overload always gets `int` argument - so I guess there is no reason to make it optional? Instead of `init` I would prefer `offset`.

opm/parser/eclipse/EclipseState/Summary/Summary.cpp
```diff
@@ -79,30 +82,114 @@ namespace Opm {
...
+       const auto& keywordstring = keyword.name();
+       const auto& schedule = es.getSchedule();
+       const auto last_timestep = schedule->getTimeMap()->last();
+       auto dims = dimensions( *es.getEclipseGrid() );
+
+       for( const auto& record : keyword ) {
+
+           const auto& wellname = record.getItem( 0 ).get< std::string >( 0 );
+
+           if( wellname == "*" ) {
```
joakim-hove: Instead of comparing with `*` I would strongly prefer:
```
const auto& well_item = record.getItem( 0 );

if (well_item.defaultApplied(0)) { 
   ...
}
```

opm/parser/eclipse/EclipseState/Summary/Summary.cpp
```diff
@@ -79,30 +82,114 @@ namespace Opm {
...
+               for( const auto& well : schedule->getWells() ) {
+
+                   const auto& name = wellName( well );
+
+                   for( const auto& completion : *well->getCompletions( last_timestep ) ) {
+                       auto cijk = getijk( *completion );
+
+                       /* well defaulted, block coordinates defaulted */
+                       if( record.size() != 4 ) {
```
joakim-hove: This test will never kick in - the `DeckRecord` will *always* have size 4, but it might consist of defaulted items. So the test should rather be based on checking the `defaultApplied()` of one of the coordinate items.

opm/parser/eclipse/EclipseState/Summary/Summary.cpp
jokva: Well, "*" is when something is requested for **all wells**, but only connections at the specified coordinates. Is the "*" then read verbatim or will it fall back to defaultApplied? It wasn't empty in the input.

opm/parser/eclipse/EclipseState/Summary/Summary.cpp
jokva: Will do.

opm/parser/eclipse/EclipseState/Summary/Summary.cpp
jokva: Line 114. Offset is fine though, I can change that.

opm/parser/eclipse/EclipseState/Summary/Summary.cpp
joakim-hove: We should have `well_item.defaultApplied( 0 ) == true` *both* for empty input and for a literal '*'; see e.g.: https://github.com/OPM/opm-parser/blob/e9e552141f102c7689e8c02e65a2ad86c86db7b6/opm/parser/eclipse/IntegrationTests/ParseDATAWithDefault.cpp#L88

> It wasn't empty in the input.


Empty input is actually not an important concept in this context - the question is really:
```C++
if item.defaultApplied( n ) ...
```
and that will not differentiate between empty input and literal `*` - on the other hand the `getValue( )` methods will not differentiate between values injected in as defaults and values entered explicitly in the deck.

opm/parser/eclipse/EclipseState/Summary/Summary.cpp
joakim-hove: OK - good!

opm/parser/eclipse/EclipseState/Summary/Summary.cpp
jokva: If "*"-as-a-string is considered `defaultApplied` then checking for that is *much* better, which makes me very happy. This will be updated accordingly, and with a comment to reflect said (maybe non-obvious) behaviour.

joakim-hove:
jenkins build this please

joakim-hove:
Tickle travis

joakim-hove:
Ahhh - that was merged before squash; my bad.

PR #737, opened 2016-03-30
RawKeyword iterator support and correct RawRecord management

jokva:
Removes some undefined behaviour introduced by iterator invalidation (string_view's) caused by reallocation of the `m_records` vector.

The quick solution has been to replace `std::vector` with `std::list`. Performance-wise this takes us back to square one, but we're no worse off than we were storing a vector of `std::shared_ptr`.

Additionally, iterators have been introduced on `RawKeyword`, since expected use of `RawKeyword` is to facilitate iteration over `RawRecords`.

### DIFF
--- a/opm/parser/eclipse/Parser/Parser.cpp
+++ b/opm/parser/eclipse/Parser/Parser.cpp
@@ -363,8 +363,7 @@ bool Parser::parseState(std::shared_ptr<ParserState> parserState) const {
                     break;
                 }
                 else if (parserState->rawKeyword->getKeywordName() == Opm::RawConsts::paths) {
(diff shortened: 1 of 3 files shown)


blattms:
It seems like either Dune::ArrayList or boost:stable_vector might work
instead of std::list if the problem is/was a growing vector. That way
we might recover some of the performance improvements.

jokva:
@blattms There are no penalties as opposed to the *original* code (which used a vector of std::shared_ptr). Additionally, I haven't observed any significant penalties (we're still within the same 1/100th of a second) compared to the (buggy) std::vector usage, so as of now this is sufficient.

Thanks for the input though, I'll surely keep those two in mind!

PR #738, opened 2016-03-31
Replace opmLog in ParseContext with MessageContainer.

qilicun:
This PR is the first step to replace OpmLog with MessageContainer in opm-parser. 
Due to the implementation of new ParseContext, lots of argument of ``const  ParseContext`` is replaced with ``ParseContext``.  If you have any new or effective way to avoid this transition, please feel free to make comments.

### DIFF
--- a/opm/parser/eclipse/EclipseState/EclipseState.cpp
+++ b/opm/parser/eclipse/EclipseState/EclipseState.cpp
@@ -105,7 +105,7 @@ namespace Opm {
     }
 
 
-    EclipseState::EclipseState(DeckConstPtr deck , const ParseContext& parseContext)
+    EclipseState::EclipseState(DeckConstPtr deck , ParseContext& parseContext)
         : m_deckUnitSystem( deck->getActiveUnitSystem() ),
           m_defaultRegion("FLUXNUM"),
           m_parseContext( parseContext )
--- a/opm/parser/eclipse/EclipseState/EclipseState.hpp
+++ b/opm/parser/eclipse/EclipseState/EclipseState.hpp
@@ -60,7 +60,7 @@ namespace Opm {
             AllProperties = IntProperties | DoubleProperties
         };
 
-        EclipseState(std::shared_ptr< const Deck > deck , const ParseContext& parseContext);
+        EclipseState(std::shared_ptr< const Deck > deck , ParseContext& parseContext);
 
         const ParseContext& getParseContext() const;
         std::shared_ptr< const Schedule > getSchedule() const;
@@ -168,7 +168,7 @@ namespace Opm {
         std::shared_ptr<FaultCollection> m_faults;
         std::shared_ptr<NNC> m_nnc;
         std::string m_defaultRegion;
-        const ParseContext& m_parseContext;
+        ParseContext& m_parseContext;
     };
 
     typedef std::shared_ptr<EclipseState> EclipseStatePtr;
--- a/opm/parser/eclipse/EclipseState/Grid/tests/ADDREGTests.cpp
+++ b/opm/parser/eclipse/EclipseState/Grid/tests/ADDREGTests.cpp
@@ -54,7 +54,8 @@ static Opm::DeckPtr createDeckInvalidArray() {
         "\n";
 
     Opm::ParserPtr parser(new Opm::Parser());
-    return parser->parseString(deckData, Opm::ParseContext()) ;
+    Opm::ParseContext parseContext;
+    return parser->parseString(deckData, parseContext) ;
 }
 
 
@@ -72,7 +73,8 @@ static Opm::DeckPtr createDeckInvalidRegion() {
         "\n";
 
     Opm::ParserPtr parser(new Opm::Parser());
-    return parser->parseString(deckData, Opm::ParseContext()) ;
+    Opm::ParseContext parseContext;
+    return parser->parseString(deckData, parseContext) ;
 }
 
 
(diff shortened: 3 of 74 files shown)


atgeirr:
You can avoid the constification by making the message container member `mutable`, but I do not think that is a good solution. It is probably better to pass the non-const ParseContext because that makes it obvious it can be changed (by adding messages).

jokva:
I rather suggest we change the interface altogether to use forwarding and keep the phyiscal context object internal to the parsing routines. If there is a *good* reason to keep ParseContext external to, say `Deck` or some parse function then please let me know, as I cannot think of one this early in the morning.

atgeirr:
I do not get what you mean by forwarding here, @jokva. Can you give a brief example?

There is a missing link that needs to be discussed as well: a function that takes all the messages produced by the parser and transfers them to the actual log system.

jokva:
Say you have the parser, `Parser p` which internally has a `ParseContext` instance. Now `ParseContext` obviously receives some configuration (what errors to accept really), so we can either:

* Pass a pre-constructed object to -copy- into the `Parser` instance and then use & modify the local copy
* Make `Parser` also accept `ParseContext`'s arguments and pass them to the internal `ParseContext`'s constructor.

bska:
> I rather suggest we change the interface altogether to use forwarding and keep the physical context object internal to the parsing routines.

Yes please!

atgeirr:
Keeping the object that gets changed (i.e. gets the messages) internal (by either of the two approaches you give) makes it hard to get at those messages from the outside after parsing is done, how do you propose to do that? Also, the internal API would still change to use non-const references (which are most of the changes of this PR), only the API used from outside would change.

jokva:
> Keeping the object that gets changed (i.e. gets the messages) internal (by either of the two approaches you give) makes it hard to get at those messages from the outside after parsing is done, how do you propose to do that? 

`Deck.readMessages()`

> Also, the internal API would still change to use non-const references (which are most of the changes of this PR), only the API used from outside would change.

Yes, but then they're internal.

However, the actually *better* approach would be to revert `ParseContext` to a configuration type value (a glorified bitset really) `ParseMode`, and then have `MessageContainer` as local instances in `Parser`, `Deck` and `EclipseState`. These `MessageContainers` would be configured by `ParseMode` as before, and be available for reading (or concatenation for a full list) once `Deck` and `EclipseState` are fully constructed.

This also solves the interface problem.

joakim-hove:
OK; 

@jokva and myself have discussed this briefly. Our suggestion is:

1. The `Deck` and the `EclipseState` classes get a `MessageContainer` member - so after you have constructed a `Deck`, or `EclipseState` instance - you can ask for the associated messages.

2. We remove the `MessageContainer` from the `ParseContext` - this way the `ParseContext` only considers error handling configuration - and should maybe be renamed `ParseConfig` ..... and can also be left as `const`.

Sorry this suggestion has come quite late.

bska:
> Our suggestion is [omitted]

Seconded.

atgeirr:
Sounds like that is a better way, I agree. I just have a few points:
 - Why the Deck and EclipseState, and not the Parser?
 - I think readMessages() is not a good name since read can be both present and past tense... simply getMessages() may be better.
 - The mythical function that calls this function and converts them to log messages should be created and used, otherwise I think we will no longer see *any* messages from opm-parser?

joakim-hove:
> There is a missing link that needs to be discussed as well: a function that takes all the messages produced by the parser and transfers them to the actual log system.

Yes - that function must be downstream - something like:

```C++
for (const auto& msg : msgContainer) {
    int64 log_type;
    switch (msg.type) {
         case MessageType::Debug:
              log_type = Log::MessageType::Debug;
              break;
         case ...
     }
     {
         const auto& location = msg.location;
         if (location) 
             OpmLog::addMessage( log_type , OpmLog::fileMessage( location->filename , location->lineno , msg.message);
         else 
             OpmLog::addMessage( log_type , msg.message );
     }
}
```

Yes - the translation of message types is extremely ugly; but as long as we insist that the message type definitions in `opm-common:OpmLog`and `opm-parser:MessageContainer` are independent I see no way around it. As a *very* weak alternative we could have a test in `opm-autodiff` which enforces that the numerical enum values in `OpmLog` and `MessageContainer` are equal. I guess this is a choice between inelegant alternatives either way.

jokva:
> Why the Deck and EclipseState, and not the Parser?

The parser too, I just wanted to keep the message short and sweet (much like myself). `Deck` and `State` can generate its own errors though, so they must also have containers.

> I think readMessages() is not a good name since read can be both present and past tense... simply getMessages() may be better.

Personally I think `get`-style names are atrocities and disgusting, but it's mostly a matter of taste. We'll get to the name once the solution is being written.

> The mythical function that calls this function and converts them to log messages should be created and used, otherwise I think we will no longer see any messages from opm-parser?

Yes, but that means messages will be "dumped" on request rather than as-it-happens.

We're considering some other changes to the parser to make phases more independent, and this type of message storage goes hand-in-hand with that.

joakim-hove:
> Why the Deck and EclipseState, and not the Parser?

We have managed to keep the parser 100 stateless; and I *really* would like to keep it that way - the parser is `const` during the parse process - something I value highly.

atgeirr:
> Yes, but that means messages will be "dumped" on request rather than as-it-happens.

That is unavoidable if opm-parser is not to use the log feature, is it not?

qilicun:
Sorry, push the wrong button.

jokva:
> That is unavoidable if opm-parser is not to use the log feature, is it not?

This is a feature indeed.

joakim-hove:
I am sorry - I am closing this now; this PR was according to the original indications - but opinions have changed.

Instead of adding the `MessageContainer` to the `ParserContext` I would like *two* `MessageContainer` instances - one as part of the `Deck` - and one as part of the `EclipseState`. Both `Deck` and `EclipseState` should have a method e.g. `messages()` which will return `MessageContainer&` - and can be used both the spin through the messages afterwards, and to add messages as `deck.messages().info("Loading PORO ...")`.

Again - sorry for the change of direction.

Issue #740, opened 2016-04-01
Exception caught: Expected double, got: '-0.000'


qilicun:
I have some data file, when I use relpermDiagnostics and flow, the program threw an exception: Expected double, got: '-0.000'.  The message comes from ``opm-parser/opm/parser/eclipseRawDeck/StarToken.cpp`` I guess this due to new changes in opm-parser recently, because at least in Feb, relperm diagnostics works very well with these data files.

joakim-hove:
Ok.- If the filer can be shared please post av a gitt,  otherwise a private email.

qilicun:
>If the filer can be shared please post av a gitt, otherwise a private email.
I'm not sure, the data is used to test relpermDiagnostics from Alf, I think @alfbr could comment on this.

joakim-hove:
OK; no need to send anything; we will fix it.

It looks like the `std::atof()` code in `readValueToken()` actually can not read 0.0000! I guess we should use `std::strtod()` instead.

qilicun:
Why not ``std::stod()``?

joakim-hove:
Hmmm - strod will not work with the string_view.  We will fix IT.

jokva:
>  It looks like the std::atof() code in readValueToken() actually can not read 0.0000! I guess we should use std::strtod() instead.

It can, but because it's a C function it also uses 0.0 to signal a possible broken value. This error is in the slow path check to verify that it wasn't a malformed number but rather actually zero (which doesn't account for preceding +/-.

jokva:
@qilicun Could you please apply and  test https://github.com/OPM/opm-parser/pull/741?

qilicun:
fixed by https://github.com/OPM/opm-parser/pull/741

PR #741, opened 2016-04-01
ReadValueToken correctly parses numbers

jokva:
Changes the implementation of numbers parsing from std::atoi/f to
std::strtod/l. These support setting the optional end-of-string pointer
which are used to determine if a parsing was successful or not. This has
the nice side effect of *greatly* simplifying the logic, at the expense
of some C-style details.

Tests added to verify that the different edge cases are handled
properly.

Fixes issue https://github.com/OPM/opm-parser/issues/740


qilicun:
It works,thanks.

PR #742, opened 2016-04-01
Parser support symlink

chflo:
Support that includes are always relative to the directory of the data file even when:

- datafile is a symlink - includes should be relative to actual data file
- include files themselves could be symlinks - should still be interpreted as relative to actual data file
- includes in includes could be symlinks and so on


joakim-hove:
Good - very well tested ;-)

PR #747, opened 2016-04-03
Rawkeyword iterator

joakim-hove:
Using range based for instead of `getRecord( )`when iterating over `RawRecords`.

The `std::next( )` based solution chocked when meeting a keyword with O(400000) records.

### DIFF
--- a/opm/parser/eclipse/Parser/Parser.cpp
+++ b/opm/parser/eclipse/Parser/Parser.cpp
@@ -375,7 +375,7 @@ bool Parser::parseState(std::shared_ptr<ParserState> parserState) const {
                     }
                 }
                 else if (parserState->rawKeyword->getKeywordName() == Opm::RawConsts::include) {
-                    auto& firstRecord = parserState->rawKeyword->getRecord(0);
+                    auto& firstRecord = parserState->rawKeyword->getFirstRecord( );
                     std::string includeFileAsString = readValueToken<std::string>(firstRecord.getItem(0));
                     boost::filesystem::path includeFile = getIncludeFilePath(*parserState, includeFileAsString);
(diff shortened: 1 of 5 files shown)


atgeirr:
> The std::next( ) based solution chocked when meeting a keyword with O(400000) records.

Not surprising given the number of records, but I am very curious about what kind of keyword that is!

joakim-hove:
> I am very curious about what kind of keyword that is!

Well - the keyword was `EDITNNC`; but the content was generated by a script used for alternative/advanced fault property modelling - "Friends don't let friends use scripts to create input files!".

akva2:

blattms:
Well, then we really should think about replacing std::list with something like boost::stable_vector where std::next should be O(1) and iterators should stay valid  if the container grows.

atgeirr:
> we really should think about replacing std::list with something like boost::stable_vector

I agree, if we need the getRecord(int) method. But is it used anywhere else than in the context considered here? If not, then simply keeping a list and deleting the getRecord(int) method is the best.

joakim-hove:
>I agree, if we need the getRecord(int) method. But is it used anywhere else than in the context considered here? If not, then simply keeping a list and deleting the getRecord(int) method is the best.

In this case the `getRecord( int )` method is not used anymore; I guess I should remove it.

joakim-hove:
jenkins build this please

akva2:
build ran (and succeeded) but you won't see the result reported back because branch was rewritten.

joakim-hove:
jenkins build this please

atgeirr:
> you won't see the result reported back because branch was rewritten

So rebasing is best done as the very last thing before merging?

akva2:
yes, if rebasing is necessary. the jenkins jobs runs with the pulled branches, i.e. master with the PR context on top.

PR #748, opened 2016-04-04
Remove MessageContainer from ParseContext + storage type change

jokva:
As a step towards (re)making `ParseContext` a read-only configuration object, remove `MessageContainer` from parse context. The patch set also changes Messages to auto-only storage, deprecating `unique_ptr`s.

### DIFF
--- a/opm/parser/eclipse/Parser/MessageContainer.cpp
+++ b/opm/parser/eclipse/Parser/MessageContainer.cpp
@@ -132,6 +132,15 @@ namespace Opm {
         );
     }
 
+    void MessageContainer::add( const Message& msg ) {
+        this->m_messages.push_back( msg );
+    }
+
+    void MessageContainer::add( Message&& msg ) {
+        this->m_messages.push_back( std::move( msg ) );
+    }
+
+
     std::vector<Message>::const_iterator MessageContainer::begin() const
     {
         return m_messages.begin();
--- a/opm/parser/eclipse/Parser/MessageContainer.cpp
+++ b/opm/parser/eclipse/Parser/MessageContainer.cpp
@@ -141,14 +141,11 @@ namespace Opm {
     }
 
 
-    std::vector<Message>::const_iterator MessageContainer::begin() const
-    {
+    MessageContainer::const_iterator MessageContainer::begin() const {
         return m_messages.begin();
     }
 
-
-    std::vector<Message>::const_iterator MessageContainer::end() const
-    {
+    MessageContainer::const_iterator MessageContainer::end() const {
         return m_messages.end();
     }
 
--- a/opm/parser/eclipse/Parser/MessageContainer.cpp
+++ b/opm/parser/eclipse/Parser/MessageContainer.cpp
@@ -24,10 +24,17 @@
 
 namespace Opm {
 
+    Location::Location( const std::string& fn, size_t ln ) :
+        filename( fn ), lineno( ln )
+    {
+        if( ln == 0 )
+            throw std::invalid_argument( "Invalid line number 0 for file '"
+                                         + fn + "'" );
+    }
 
     void MessageContainer::error( const std::string& msg,
                                   const std::string& filename,
-                                  const int lineno ) {
+                                  const size_t lineno ) {
         m_messages.emplace_back(
             Message { Message::Error, msg, Location { filename, lineno } }
         );
@@ -41,7 +48,7 @@ namespace Opm {
 
     void MessageContainer::bug( const std::string& msg,
                                 const std::string& filename,
-                                const int lineno ) {
+                                const size_t lineno ) {
         m_messages.emplace_back(
             Message { Message::Bug, msg, { filename, lineno } }
         );
@@ -55,7 +62,7 @@ namespace Opm {
 
     void MessageContainer::warning( const std::string& msg,
                                     const std::string& filename,
-                                    const int lineno ) {
+                                    const size_t lineno ) {
         m_messages.emplace_back( Message {
                 Message::Warning, msg, { filename, lineno }
             }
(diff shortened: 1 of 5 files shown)


akva2:

opm/parser/eclipse/Parser/MessageContainer.hpp
```diff
@@ -43,21 +41,31 @@ namespace Opm {
 
 
     struct Location {
+        Location() = default;
+        Location( const std::string& s, int l ) : filename( s ), lineno( l ) {}
```
atgeirr: `l` is a bad variable name, being too close to a `1`.

opm/parser/eclipse/Parser/MessageContainer.cpp
```diff
@@ -25,94 +25,127 @@
...
 
 
-    void MessageContainer::error(const std::string& msg, const std::string& filename, const int lineno)
-    {
-        std::unique_ptr<Location> loc(new Location{filename, lineno});
-        m_messages.emplace_back(Message{MessageType::MessageTypeEnum::Error, msg, std::move(loc)});
+    void MessageContainer::error( const std::string& msg,
+                                  const std::string& filename,
+                                  const int lineno ) {
+        m_messages.emplace_back( Message {
```
atgeirr: Is not one of the reasons to use `emplace_back` that you can drop the explicit construction and just pass the arguments?
```
m_messages.emplace_back(
    MessageType::MessageTypeEnum::Error,
    msg,
    { filename, lineno });
```

opm/parser/eclipse/Parser/tests/MessageContainerTest.cpp
```diff
@@ -34,8 +34,8 @@ BOOST_AUTO_TEST_CASE(TestIterator) {
     msgContainer.bug("This is a bug.", "dummy.log", 20);
     {       
         BOOST_CHECK_EQUAL("This is an error.", msgContainer.begin()->message);
-        BOOST_CHECK_EQUAL("dummy.log", (msgContainer.end()-1)->location->filename);
-        BOOST_CHECK_EQUAL(20, (msgContainer.end()-1)->location->lineno);
+        BOOST_CHECK_EQUAL("dummy.log", (msgContainer.end()-1)->location.filename);
+        BOOST_CHECK_EQUAL(20, (msgContainer.end()-1)->location.lineno);
     }
```
joakim-hove: Can we have a test demonstrating the `operator bool( )` functionality of the `Location` class.

opm/parser/eclipse/Parser/MessageContainer.cpp
jokva: The compiler is unable to deduce it correctly without the hint from `Message {}` (or constructor definitions). I use `emplace_back` just to emphasise the move going on.

An alternative is surely to define two constructors for `Message`, in which case the `Message {}` can go away.

opm/parser/eclipse/Parser/tests/MessageContainerTest.cpp
jokva: Yes.

opm/parser/eclipse/Parser/MessageContainer.hpp
jokva: Actually a slight oversight on my part, this was mostly thrown in for testing the emplace_back syntax in https://github.com/OPM/opm-parser/pull/748#discussion_r58354804

I

joakim-hove:
jenkins build this please

opm/parser/eclipse/Parser/MessageContainer.hpp
```diff
@@ -20,17 +20,27 @@
...
 #include <vector>
 #include <memory>
 
 namespace Opm {
 
-    namespace MessageType {
+    struct Location {
+        Location() = default;
+        Location( const std::string& fn, int ln ) :
+            filename( fn ), lineno( ln ) {}
```
joakim-hove: This should throw for `ln <  0` - it should be freaking impossible to explicitly create a `Location` instance which evaluates to `false`.

joakim-hove:
I will merge this when we have a test demonstrating the `bool operator()`behavior of the `class Location` - something like:
```C++
Location loc_true("filename" , 100);
Location loc_false;
BOOST_CHECK( loc_true );
BOOST_CHECK( !loc_false);
BOOST_CHECK_THROW( Location("filename" , -1) , std::invalid_argument);
```

jokva:
@joakim-hove pushed, but without throw check. Simply because location.filename should be considered undefined if lineno is nonsensical.

Obviously this makes  container.warning( msg, filename, 0 ) which I'm happy to disallow.

joakim-hove:
Well - my point was to disallow construction of  an invalid location by checking and throwing in the cobstructor,  and documenting that with a test.

opm/parser/eclipse/Parser/MessageContainer.hpp
blattms: On Tue, Apr 05, 2016 at 01:45:05PM -0700, Joakim Hove wrote:
> >  #include <string>
> >  #include <vector>
> >  #include <memory>
> >  
> >  namespace Opm {
> >  
> > -    namespace MessageType {
> > +    struct Location {
> > +        Location() = default;
> > +        Location( const std::string& fn, int ln ) :
> > +            filename( fn ), lineno( ln ) {}
> 
> This should throw for `ln <  0` - it should be freaking impossible to explicitly create a `Location` instance which evaluates to `false`.

or instead change the type from int to std::size_t or another
unsgigned integer as negative values make no sense at all

joakim-hove:
OK - thank you for the test update.

PR #739, opened 2016-03-31
Introduce Eclipse3DProperties

pgdr:
Introduce Eclipse3DProperties and moved GridProperties stuff to new class.

Clean up initialization process of EclipseState.  Moved some accessors from EclipseState to Eclipse3DProperties that now need to be accessed via EclipseState.getEclipseProperties.

Made some members of state/properties references instead of shared_ptr.

PR #751, opened 2016-04-08
Introduce Eclipse3DProperties

pgdr:
This pull request replaces #739.

Introduce Eclipse3DProperties and moved GridProperties stuff to new class.

Clean up initialization process of EclipseState. Moved some accessors from EclipseState to Eclipse3DProperties that now need to be accessed via EclipseState.getEclipseProperties.

Made some members of state/properties references instead of shared_ptr.

Downstream: 
- OPM/opm-material#137
- OPM/opm-core#994
- OPM/opm-simulators#644
- OPM/opm-output#14


joakim-hove:
OK; this is ready to be merged. Will merge as soon as the downstream modules: https://github.com/OPM/opm-material/pull/137 +++ are marked as ready.

opm/parser/eclipse/EclipseState/EclipseState.hpp
```diff
@@ -70,33 +72,15 @@ namespace Opm {
...
-        void loadGridPropertyFromDeckKeyword(std::shared_ptr<const Box> inputBox,
-                                             const DeckKeyword& deckKeyword,
-                                             int enabledTypes = AllProperties);
 
         std::shared_ptr<const FaultCollection> getFaults() const;
         std::shared_ptr<const TransMult> getTransMult() const;
         std::shared_ptr<const NNC> getNNC() const;
         bool hasNNC() const;
 
+        const Eclipse3DProperties& getEclipseProperties() const;
```
joakim-hove: Sorry: this should be `getEclipse3DProperties()` - the room for confusion regarding names consisting of *Eclipse* and *properties* is **endless**.

jokva: Isn't eclipse already well-implied by the fact that it's on EclipseState? I suppose even get3DProperties would be sufficient (and an improvement).

pgdr: Which is it then?  `getEclipse3DProperties` or `get3DProperties` or `getProperties`?

andlaus: > Sorry: this should be getEclipse3DProperties() - the room for confusion regarding names consisting of Eclipse and properties is endless.

actually I have to concure: I'm already confused. Aren't grids always three dimensional and thus all properties?

PR #752, opened 2016-04-11
Add MessageContainer to EclipseState and Deck.

qilicun:


opm/parser/eclipse/EclipseState/EclipseState.cpp
```diff
@@ -125,6 +126,7 @@ namespace Opm {
         initFaults(deck);
         initMULTREGT(deck);
         initNNC(deck);
+        initMessageContainer();
```
pgdr: Why not put m_messageContainer initialization in initializer list?

opm/parser/eclipse/Deck/Deck.hpp
```diff
@@ -118,6 +120,8 @@ namespace Opm {
             void initUnitSystem() const;
 
             std::vector< DeckKeyword > keywordList;
+            std::shared_ptr<MessageContainer> m_messageContainer;
+
```
pgdr: Does it have to be a shared_ptr?  Couldn't it just be `MessageContainer m_messageContainer;` ?

joakim-hove:
We are trying to cut down on the use of `std::shared_ptr<T>` in the opm-parser API. In this case it should be OK to just add value members `MessageContainer m_messageContainer` to both the `Deck` and `EclipseState`. The `getMessages()` method can then just return a normal reference - should probably have both const and non const overloads.

qilicun:
>MessageContainer m_messageContainer to both the Deck and EclipseState

Will this store the messages in two different objects?

atgeirr:
> Will this store the messages in two different objects?

Yes, but so will the PR as it stands: you are making two MessageContainer objects, each in their own shared_ptr, but they do not point to the same instance. With shared_ptr it is possible to have the Deck and EclipseState use the same MessageContainer (by having the accessor return the shared_ptr and having the EclipseState copy it from the Deck), if that is what we want. I am not sure if it is?

joakim-hove:
> Will this store the messages in two different objects?

Yes it will - and that is fine.

opm/parser/eclipse/Deck/Deck.hpp
jokva: It can and it should. Additionally, if it is just `MessageContainer` it doesn't have to be explicitly created in the constructor.

joakim-hove:
Good - merging; looking forward to see this in use.

PR #757, opened 2016-04-13
Add appendMessages() function.

qilicun:


opm/parser/eclipse/Parser/MessageContainer.cpp
```diff
@@ -128,6 +128,15 @@ namespace Opm {
     }
 
 
+    void MessageContainer::appendMessages(const MessageContainer& msgContainer)
+    {
+        for(auto it = msgContainer.begin(); it != msgContainer.end(); it++) {
```
joakim-hove: When possible prefer to use the range based for to naked iterators:

```C++
for( const auto& msg : msgContainer )
    this->add( msg );
```

joakim-hove:
Please add a test to `MessageContainerTest.cpp`.

opm/parser/eclipse/Parser/MessageContainer.cpp
```diff
@@ -128,6 +128,15 @@ namespace Opm {
     }
 
 
+    void MessageContainer::appendMessages(const MessageContainer& msgContainer)
```
joakim-hove: I would prefer if the input argument was called `other` or something like that; because internally in this method we actually have *two* message containers in scope.

opm/parser/eclipse/Parser/tests/MessageContainerTest.cpp
```diff
@@ -62,3 +62,14 @@ BOOST_AUTO_TEST_CASE(LocationImplicitConversion) {
...
 }
+
+
+BOOST_AUTO_TEST_CASE(appendMessages) {
+    MessageContainer msgContainer;
+    MessageContainer msgList;
+    msgContainer.error("Error: msgContainer.");
+    msgList.warning("Warning: msgList.");
+    msgContainer.appendMessages(msgList);
+    BOOST_CHECK_EQUAL("Error: msgContainer.", msgContainer.begin()->message);
```
joakim-hove: Well, this does actually not  test the addMessages very much, please verify that the updated container has length 2 and ends with the right message.

joakim-hove:
Thank you I will merge now; but I am quite certain you are getting a signed / unsigned warning when compiling the test?

qilicun:
>but I am quite certain you are getting a signed / unsigned warning when compiling the test?

I did not see that...

opm/parser/eclipse/Parser/MessageContainer.cpp
```diff
@@ -128,6 +128,14 @@ namespace Opm {
     }
 
 
+    void MessageContainer::appendMessages(const MessageContainer& other)
```
jokva: Why wouldn't you just use vector.insert? That way you allocate more efficiently.

joakim-hove:
> I did not see that...

Maybe it was not there in the first place - we will fix if needed.

PR #758, opened 2016-04-13
Refactoring of internal parsing

jokva:
A somewhat large patch set of smaller incremental updates (for ease of review) designed to simplify the structure of the internal parsing, remove some redundant code and enable new optimisations. I've tried to keep every commit tiny and single-purpose. A few highlights:

* `string_view` uses char* internally instead of `std::string::const_iterator`
* streams removed in favour of reading the whole file at once and work with it in memory
* `RawRecord`s no longer copy memory, but stores offsets into the buffered file
* Simplified control flow of several helper functions
* The input file is immediately stripped for comments and other things to ignore - this reduces special casing and simplifies actual parsing
* Various performance benefits (I measure ~15%)

Except for speedups, no side effects of this effort should be visible for other projects.

### DIFF
--- a/opm/parser/eclipse/Parser/Parser.cpp
+++ b/opm/parser/eclipse/Parser/Parser.cpp
@@ -614,14 +614,14 @@ bool Parser::parseState(std::shared_ptr<ParserState> parserState) const {
 
 
     void Parser::applyUnitsToDeck(Deck& deck) const {
-        for (size_t index=0; index < deck.size(); ++index) {
-            auto& deckKeyword = deck.getKeyword( index );
-            if (isRecognizedKeyword( deckKeyword.name())) {
-                const auto* parserKeyword = getParserKeywordFromDeckName( deckKeyword.name() );
-                if (parserKeyword->hasDimension()) {
-                    parserKeyword->applyUnitsToDeck(deck , deckKeyword);
-                }
-            }
+        for( auto& deckKeyword : deck ) {
+
+            if( !isRecognizedKeyword( deckKeyword.name() ) ) continue;
+
+            const auto* parserKeyword = getParserKeywordFromDeckName( deckKeyword.name() );
+            if( !parserKeyword->hasDimension() ) continue;
+
+            parserKeyword->applyUnitsToDeck(deck , deckKeyword);
         }
     }
 
--- a/opm/parser/eclipse/Parser/Parser.cpp
+++ b/opm/parser/eclipse/Parser/Parser.cpp
@@ -487,16 +487,28 @@ bool Parser::parseState(std::shared_ptr<ParserState> parserState) const {
         return line;
     }
 
-    static inline std::string& trim_right( std::string& str ) {
-        /* because cctype functions can be macros, manually find the
-         * position to erase from. Written using isspace for performance.
-         */
+    template< typename Itr >
+    static inline Itr trim_left( Itr begin, Itr end ) {
 
-        size_t i = str.size();
-        for( ; i >= 1; --i )
-            if( !isspace( str[ i - 1 ] ) ) break;
+        const auto ws = []( char ch ) { return std::isspace( ch ); };
+        return std::find_if_not( begin, end, ws );
+    }
 
-        return str.erase( i );
+    template< typename Itr >
+    static inline Itr trim_right( Itr begin, Itr end ) {
+
+        const auto ws = []( char ch ) { return std::isspace( ch ); };
+
+        std::reverse_iterator< Itr > rbegin( end );
+        std::reverse_iterator< Itr > rend( begin );
+
+        return std::find_if_not( rbegin, rend, ws ).base();
+    }
+
+    static inline std::string& trim( std::string& str ) {
+        auto fst = trim_left( str.begin(), str.end() );
+        auto lst = trim_right( fst, str.end() );
+        return str.assign( fst, lst );
     }
 
     bool Parser::tryParseKeyword(std::shared_ptr<ParserState> parserState) const {
@@ -510,11 +522,9 @@ bool Parser::parseState(std::shared_ptr<ParserState> parserState) const {
 
         std::string line;
         while (std::getline(*parserState->inputstream, line)) {
-            if (line.find("--") != std::string::npos)
-                line = strip_comments( line );
-
-            line = trim_right( line ); // Removing garbage (eg. \r)
-            line = doSpecialHandlingForTitleKeyword(line, parserState);
+            line = strip_comments( line );
+            line = trim( line ); // Removing garbage (eg. \r)
+            doSpecialHandlingForTitleKeyword(line, parserState);
             std::string keywordString;
             parserState->lineNR++;
 
--- a/opm/parser/eclipse/Parser/Parser.cpp
+++ b/opm/parser/eclipse/Parser/Parser.cpp
@@ -647,7 +647,7 @@ bool Parser::parseState(std::shared_ptr<ParserState> parserState) const {
                         return true;
                     }
                 }
-                parserState->rawKeyword->addRawRecordString(line.string());
+                parserState->rawKeyword->addRawRecordString(line);
             }
 
             if (parserState->rawKeyword
--- a/opm/parser/eclipse/Parser/Parser.cpp
+++ b/opm/parser/eclipse/Parser/Parser.cpp
@@ -158,10 +158,10 @@ namespace Opm {
           of the data section of any keyword.
         */
 
-        void handleRandomText(const std::string& keywordString ) const {
+        void handleRandomText(const string_view& keywordString ) const {
             std::string errorKey;
             std::stringstream msg;
-            std::string trimmedCopy = boost::algorithm::trim_copy( keywordString );
+            std::string trimmedCopy = keywordString.string();
 
             if (trimmedCopy == "/") {
                 errorKey = ParseContext::PARSE_RANDOM_SLASH;
@@ -250,9 +250,8 @@ namespace Opm {
         return strip_comments( qend + 1, end );
     }
 
-    static inline std::string& strip_comments( std::string& str ) {
-        str.erase( strip_comments( str.begin(), str.end() ), str.end() );
-        return str;
+    static inline string_view strip_comments( const string_view& str ) {
+        return { str.begin(), strip_comments( str.begin(), str.end() ) };
     }
 
     /* stripComments only exists so that the unit tests can verify it.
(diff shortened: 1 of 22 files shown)


atgeirr:
I'd like to comment on a few items here:

> string_view uses char* internally instead of std::string::const_iterator

Would not that make it more difficult to detect bugs that a debug iterator implementation possibly could find?

> streams removed in favour of reading the whole file at once and work with it in memory

I do not know how large a "big" deck can get, but the Norne data set (< 50 k cells) is ~30 MB on disk. Assuming size scales with number of cells a 5 M cell case should be 3 GB or more (how much of this would end up in memory I don't know). I assume this is considered acceptable?

> RawRecords no longer copy memory, but stores offsets into the buffered file

This sounds like it could be more robust (w.r.t. string_view and copied strings etc.), since the buffered file remains for the duration of parsing. But there was some special treatment, I think it was of the "5*100.0" syntax (5 times 100.0) or perhaps related to defaults, that used to modify the strings (remember the case of the extra space?), has that been changed, too?

jokva:
> Would not that make it more difficult to detect bugs that a debug iterator implementation possibly could find?

Not really; `char*` is somewhat easier to inspect, but more string_view was rather stripped of debugging information anyway. Regardless, the *benefit* of using `char*` outweighs that anyway.

> I do not know how large a "big" deck can get, but the Norne data set (< 50 k cells) is ~30 MB on disk. Assuming size scales with number of cells a 5 M cell case should be 3 GB or more (how much of this would end up in memory I don't know). I assume this is considered acceptable?

Until proven otherwise, yes, in which case it's possible to re-adopt streaming behaviour and accept the necessary penalties.

This memory only lives during the parsing process, and is thrown immediately after. Additionally, we use significantly less heavy objects this way, so I assume memory overhead is lightened (although I haven't measured.). This memory requirement is also present for the finished deck+state too, which obviously uses binary representation for the same data which may-or-may not be larger (and which also keeps it all around).

I don't parse-phase memory consumption should become an issue, but if someone has a huge model lying around I'd love for them to test the patches.

> This sounds like it could be more robust (w.r.t. string_view and copied strings etc.), since the buffered file remains for the duration of parsing. But there was some special treatment, I think it was of the "5*100.0" syntax (5 times 100.0) or perhaps related to defaults, that used to modify the strings (remember the case of the extra space?), has that been changed, too?

That's the idea - we can copy strings around for free since we know the referenced memory is immutable. The repeated-string is currently special cased and the generated string stored separately (i.e. it hasn't been modified w.r.t. the old implementation) and still works fine. I have a plan for changing it into pushing a repeated `string_view` onto the `deque`, but I haven't quite gotten there yet.

jokva:
I would also like to point out that even with streams in use, we still effectively kept the entire input file around (although the memory was spread over multiple `std::string` instances.

atgeirr:
> we still effectively kept the entire input file around (although the memory was spread over multiple std::string instances.

I see, so this should actually reduce memory usage, not increase it!

jokva:
Yes. Although without measuring I can't say that for sure. It also depends on how good gcc/libstdc++/libc++ are at optimising string memory usage.

jokva:
I  take that back, this *does* increase the memory usage compared to the streaming solution. Not really because of stream usage, but because the old solution would drop the parsed input strings once a DeckKeyword was completed, and with this patch this is done once all input has been read.

It *is* possible to return to old behaviour since we already pseudo-stream if include files are used. Quick summary so we're on the same page:

1. The parser reads the full input file into memory and strips noise, comments etc.
2. It starts to parse left-to-right using a custom getline (much like with streams).
3. It creates DeckKeywords (the value it wants to produce) and memory preceding this is for all intents and purposes garbage. This is where the previous solution would free, but we keep it around
4. It meets an include and pushes this file to the top of its parsing stack and proceeds to parse from step 1. It parses it through and then resumes from the point in the previous file it found the include.
5. Continues until there is nothing more to parse.

At step 4, once the included file is fully parsed, we can actually release it. We don't right now, but it's not that much more work to do. It's actually unrolled recursion (master currently uses this). Of course, the buffers will be much larger than with `iostream`, and if the entire input deck is in one file then we *will* keep it around for the lifetime of the parse call, but we do get some nice benefits for it. Grids are often in separate files anyway, so it is definitely solvable if it becomes a problem.

tl;dr: this patch set introduces higher memory consumption, but it's easy to make it "streaming" again.

andlaus:
Is opm-parser memory usage actually a problem in practice? I mean a couple dozen megabytes don't really matter anymore nowadays? (My point is: It's probably wiser to keep this simple instead of spending a lot of effort on something which is only slightly more efficient.)

atgeirr:
> Is opm-parser memory usage actually a problem in practice

Probably not, and I agree that reducing this should not be a priority. I do not mind the approach taken by this PR, but thought it was worth discussing.

opm/parser/eclipse/Parser/Parser.cpp
```diff
@@ -42,60 +42,194 @@
...
+        auto qend = std::find( qbegin + 1, end, *qbegin );
+
+        // Quotes are not balanced - probably an error?!
+        if( qend == end ) return end;
+
+        return strip_after( qend + 1, end, terminator );
+    }
+
+    /**
+       This function will remove return a copy of the input string
```
joakim-hove: ... *remove return* - a typo?

opm/parser/eclipse/Parser/Parser.cpp
```diff
@@ -42,60 +42,194 @@
...
+            return this->input_stack.back()->path;
         }
 
-        ParserState(const ParseContext& __parseContext)
-            : parseContext( __parseContext ),
-              deck( new Deck() ),
-              pathMap( std::make_shared< std::map< std::string, std::string > >() ),
-              lineNR( 0 )
-        {}
+        size_t& line() {
```
joakim-hove: The possibility to modify line number does not seem like a *must have*?

opm/parser/eclipse/Parser/Parser.cpp
```diff
@@ -105,17 +239,34 @@ namespace Opm {
...
+            /*
+             * read the input file C-style. This is done for performance
+             * reasons, as streams are slow
+             */
+            auto* fp = ufp.get();
+            std::string buffer;
+            std::fseek( fp, 0, SEEK_END );
+            buffer.resize( std::ftell( fp ) );
+            std::rewind( fp );
+            std::fread( &buffer[ 0 ], 1, buffer.size(), fp );
```
joakim-hove: I think you can use `buffer.data()`

jokva: No, because it is not non-const until C++17: http://en.cppreference.com/w/cpp/string/basic_string/data

I **really** wish I could because this is sort-of hacky, but it's the only way (to my knowledge) to make it work.

opm/parser/eclipse/Parser/Parser.cpp
jokva: It is: https://github.com/jokva/opm-parser/blob/internal-parser-procedure-use-ranges/opm/parser/eclipse/Parser/Parser.cpp#L622

opm/parser/eclipse/Parser/Parser.cpp
jokva: Yes.

opm/parser/eclipse/Parser/Parser.cpp
```diff
@@ -42,60 +42,194 @@
...
 
+        ParserState(const ParseContext& __parseContext)
+            : parseContext( __parseContext ),
+              deck( new Deck() )
+        {}
 
-        void openStream(std::shared_ptr<std::istream> istream) {
-            dataFile = "";
-            inputstream = istream;
+        void openString(const std::string& input ) {
```
joakim-hove: Since this is not streaming anymore - can we have `loadString` and also `loadFile`?

opm/parser/eclipse/Parser/Parser.cpp
```diff
@@ -42,60 +42,194 @@
...
+        return std::find_if_not( rbegin, rend, RawConsts::is_separator ).base();
+    }
+
+    static inline string_view trim( string_view str ) {
+        auto fst = trim_left( str.begin(), str.end() );
+        auto lst = trim_right( fst, str.end() );
+        return { fst, lst };
+    }
+
+    static inline string_view strip_slash( string_view view ) {
```
joakim-hove: This is really `strip_after_slash( )`?

opm/parser/eclipse/Parser/Parser.cpp
```diff
@@ -42,60 +42,194 @@
...
+
+        auto itr = std::find( begin, end, '-' );
+        for( ; itr != end; itr = std::find( itr + 1, end, '-' ) )
+            if( (itr + 1) != end &&  *( itr + 1 ) == '-' ) return itr;
+
+        return end;
+    }
+
+    template< typename Itr, typename Term >
+    static inline Itr strip_after( Itr begin, Itr end, Term terminator ) {
```
joakim-hove: This is really `quoted_find()`?

opm/parser/eclipse/Parser/Parser.cpp
jokva: Yes.

opm/parser/eclipse/Parser/Parser.cpp
jokva: It is, I'm happy to rename it. It was named this way to be consistent with strip_comments.

opm/parser/eclipse/Parser/Parser.cpp
```diff
@@ -42,60 +42,194 @@
...
-         * resolution, i.e. PATHS specified in an included file can be used
-         * from the file that included it.
-         */
-        std::shared_ptr< std::map<std::string, std::string> > pathMap;
-        size_t lineNR;
-        std::shared_ptr<std::istream> inputstream;
+        std::map< std::string, std::string > pathMap;
+        std::list< std::string > input_strings;
+        std::list< file > input_files;
+        std::vector< file* > input_stack;
```
joakim-hove: I have two wishes for the members `input_strings`, `input_files` and `input_stack`:

1. They should be collected in a small container - e.g. `ContentStack` which has `push()`, `pop()` and `get()` (maybe even the two last can be combined?).
2. The sole purpose of the `input_strings` and `input_files` is to ensure that the `std::string` and `file` objects do not go out of scope - to communicate that I would like names like e.g. `string_storage` and `file_storage`.

opm/parser/eclipse/Parser/Parser.cpp
jokva: Not really. It looks for some terminator, either terminating slash *or* a comment (it is the shared implementation for both), but respects that both constructs can be present in quotes. It also properly returns the interesting range. It's really a `takeWhile` that respects quotes.

opm/parser/eclipse/Parser/Parser.cpp
jokva: This is the natural development for this (but I wanted to keep it out of this PR), especially since it allows us to re-enable streaming behaviour at the flip of a switch. I kept it out to keep the actual reading changes as small as possible, but I'm happy adding a new commit that fixes this.

Making the input stack a richer object is a good idea and was certainly on the horizon.

joakim-hove:
I was going to methe this,  but some rebasing is required.

PR #761, opened 2016-04-14
Message container sub class

qilicun:
Add message container to sub directories in EclipseState folder.
Each sub-class handle the messages and append it to  EclipseState finally.


qilicun:
There are some cases that can't handle. In ``EclipseGrid.cpp`` and ``Eclipse3DProperties.cpp``, some member functions are **static**, thus can't call the message container member to log these messages. The worse thing is that ``OpmLog`` is used in thoes functions...

opm/parser/eclipse/EclipseState/EclipseState.cpp
```diff
@@ -99,6 +97,9 @@ namespace Opm {
 
     
     MessageContainer& EclipseState::getMessageContainer() {
+        m_messageContainer.appendMessages(m_tables.getMessageContainer());
```
jokva: This interface is broken. 

If you call `getMessageContainer` (which is named and formed as a read-as-is) multiple times you will likely just collect the set of sub container messages as many times.

joakim-hove: Yes; should really think twice (at least!) before you let a `get()` method change the state of the object.

My suggestion: 
 
1. The `EclipseState` has *only*  a const variety of the `getMessageContainer()`
2. The assembly of messages from the various sub components is done at the end of the `EclipseState` constructor.

That way the division into sub message containers is (close to) an implementation detail - and downstream can choose to only care about the collection of messages coming from `EclipseState`

joakim-hove:
> There are some cases that can't handle

Will come with suggestion later today - unless someone else beats me to it.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -18,7 +18,7 @@
 
 
 #include <boost/algorithm/string/join.hpp>
-
+#include <opm/common/OpmLog/OpmLog.hpp>
```
joakim-hove: OK - why is this? If the `Eclipse3DProperties`has calls to `OpmLog` those as well should be replaced with adding messages to a container; and the `Eclipse3DProperties` class should also get a `MessageContainer` instance.

joakim-hove:
> There are some cases that can't handle

This was not obviously clear to me - good if you can point me in the right direction. But briefly I guess the solution is to make the offending static methods into ordinary bound methods.

jokva:
No, they're static for a reason. If something doesn't have to be on an instance it shouldn't be. And if they're static it shouldn't be necessary for them to log something (they're just computations after all) - the logging should rather be handled by whoever it is invoking the function.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
qilicun: Because of the ``static`` function. I didn't touch this file.

qilicun:
So the simplest way here to solve the static function problem seems like to delete the OpmLog directly?

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
joakim-hove: > I didn't touch this file.

"This file" - meaning exactly - `Eclipse3DProperties`?

qilicun: Yes, ``OpmLog.hpp`` was removed by mistake from``Eclipse3DProperties.cpp`` , when I found there are some static functions which use OpmLog, I decided recovery OpmLog header.

joakim-hove:
> So the simplest way here to solve the static function problem seems like to delete the OpmLog directly?

Now I have looked at the offending call site: https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp#L898. The problem caught here is quite severe; this should either be logged (to the `MessageContainer`) - or an exception should be raised. The problem identified at this call site is quite clearly documented as forbidden in the Eclipse documentation, and we have (as far as I know) never encountered it in the wild. I.e. this is probably not a "Works in Eclipse" situation which we must handle - I would therefor very carefully go for an exception based solution.

If an exception is not acceptable the method must add a message to a messagecontainer using one of two:

1. The `Eclipse3DProperties` gets a `MessageContainer` member just like the other large ojects - and the method in question becomes non static. 
2. Pass a `MessageContainer& ` argument to the methods in the `Eclipse3DProperties` class.

I would go for 1.

qilicun:
I agree with the first suggestion, If no one has strong opinion against 1, I will follow that way. Also apply it to ``EclipseGrid`` class.

jokva:
I disagree; it still doesn't depend at *all* on the internal state of `3DProperties` and shouldn't need to now. Make it return a tuple of success indicator+message or give it wherever it should log. I'm partial to the former.

joakim-hove:
> it still doesn't depend at all on the internal state of 3DProperties and shouldn't need to now. 

In principle that is easy to agree - but we **really** should conclude this now; so although not giving a carte blanche, I will more or less merge anything which works.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -854,7 +853,17 @@ namespace Opm {
         }
     }
 
-    // static member function
+    
+    const MessageContainer& Eclipse3DProperties::getMessageContainer() const {
```
joakim-hove: Oh my oh my - these containers have multiplied! This implementation seems wrong; when the `MessageContainer`of the full `Eclipse3DProperties` is bound to the `intGridProperties` member messages added by `GridProperties<double>` will be lost space.

I suggest you add the `MessageContainer` member to the `Eclipse3DProperties` instance itself, and then pass a reference to that instance to the `GridProperty<T>` methods - instead of letting them have a private `MessageContainer`.

qilicun: >messages added by GridProperties<double> will be lost space.

I agree that's a fault. need more consideration.

> and then pass a reference to that instance to the GridProperty<T>

I'm not quite clear how to do this, should I add one more argument to the methods?
```
 GridProperty<int>&      getIntGridProperty()
 GridProperty<double>&    getDoubleGridProperty()

```

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -895,10 +904,7 @@ namespace Opm {
         } else if (setCount != 0) {
             std::string msg = "BOX modifiers on keywords must be either "
                 "specified completely or not at all. Ignoring.";
-            OpmLog::addMessage(Log::MessageType::Error,
-                               Log::fileMessage(deckKeyword.getFileName(),
-                                                deckKeyword.getLineNumber(),
-                                                msg));
+            m_intGridProperties.getMessageContainer().error(deckKeyword.getFileName() + std::to_string(deckKeyword.getLineNumber()) + msg);
```
pgdr: Should `+` be `,` ?

qilicun: No, because the ``error()`` function accept a ``std::string`` argument, so  it should be ``+``, not ``,``

jokva: If you have file information, at least use the `error(msg, filename, lineno)` overload.

qilicun: That's possible. will change.

pgdr: But [a different MessageContainer.error](https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/Parser/MessageContainer.hpp#L70) function has the following signature
> `void error(const std::string& msg, const std::string& filename, const size_t lineno);`

qilicun: Sorry for forgetting that function...

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
qilicun: If ``GridPropertie`` does not have ``MessageContainer`` itself, some methods both in ``GridProperties`` and ``Eclipse3DProperties`` need to be modified to accept the ``MessageContainer`` reference from ``Eclipse3DProperties``, like ``initPORV``, ``addKeyword``, ``getOrCreateProperty``, etc.  A quick ``grep`` shows some of these funcitons are not used in other opm module, that shows we don't need to modify the other calling scope, but I'm not sure about this.

pgdr: If the question is whether `getIntGridProperty` and `getDoubleGridProperty` should take `messageContainer` as a parameter, I think the clear answer is _no_.  If you want to add a new input argument to `initPORV` you also have to add a new input argument to _every postprocessor and initializer_.  I don't think that is a good idea.

qilicun: >I don't think that is a good idea.

I don't think either.

If I understood @joakim-hove correct, the problem comes again: if ``GridProperties`` doesn't have the ``MessageContainer`` member itself, and don't pass a ``MessageContainer`` reference to the methods, how does the message transfer from ``intGridProperties`` and ``doubleGridProperties`` to ``Eclipse3DProperties`` without losing and message?  What do you think? @pgdr

pgdr: I don't have a good solution for this ... perhaps have a `MessageContainerManager` (`LogManager`) where you can ask for a `MessageContainer` whenever you need one?

joakim-hove:
@qilicun : I have fallen a bit out of sync here; good if you could summarize the current situation. As I have said previously I *really* think we should finalize this now; I do not think we will find a 'perfect' implementation in this round - so I will merge this when it works&trade;.

I previously suggested:

>I suggest you add the MessageContainer member to the Eclipse3DProperties instance itself, and then pass a reference to that instance to the GridProperty<T> methods - instead of letting them have a private MessageContainer.

Seems from the next messages that this solution was not very practical; then my next suggestion would be to let the `gridProperties<T>` class have a `MessageContainer` instance, and then the implement the `Eclipse3DProperties::getMessages()` as:

```C++
MessageContainer Eclipse3DProperteis::getMessages() {
    MessageContainer messages;
    messages.add( m_intProperties.gteMessages() );
    messages.add( m_doubleProperties.getMessages() );
}
```

The `MessageContainer` seems like a prime candidate for `operator+`.


> If the question is whether getIntGridProperty and getDoubleGridProperty should take messageContainer as a parameter, I think the clear answer is no. 

Here I agree strongly with @pgdr - we should **not** start on a route which will require changes initializer  / postprocessor signature.

qilicun:
>  I really think we should finalize this now.

I agree,  the PR is update, if travis feel happy, it is ready to merge.

Apart from this PR, there are some OpmLog left in Parser class, will do it the same way in Deck and EclipseState.

joakim-hove:
> I agree, the PR is update, if travis feel happy, it is ready to merge.

OK - I will merge this soon.

> there are some OpmLog left in Parser class, will do it the same way in Deck and EclipseState.

That I did not really understand?!

qilicun:
>That I did not really understand?!

I mean in Parser folder,  ``Parser`` class use ``OpmLog`` functionality to logging, it will be replaced with ``MessageContainer`` as the same way as ``Deck`` and ``EclipseState``, that is add a ``MessageContainer`` member to ``Parser`` class, do you think it's acceptable?

joakim-hove:
> that is add a MessageContainer member to Parser class, do you think it's acceptable?

Well - I have really tried hard to keep the parser *stateless* - that would break that. I though the messages generated during parsing would go to the `MessageContainer` in the `Deck`?

qilicun:
> would go to the MessageContainer in the Deck?

That's right! Will do it that way.

joakim-hove:
> That's right! Will do it that way.

Good - will you update this PR; or should I merge this and wait for another PR?

qilicun:
>hould I merge this and wait for another PR?

Please merge this first, I will create another PR.

PR #762, opened 2016-04-14
Silence missing-braces warning.

atgeirr:
Can switch back when we move to C++14 uniformly.


jokva:
Thanks! I'll merge this once we've fixed a broken merge.

PR #768, opened 2016-04-19
Fixes support for BOXed ACTNUM

pgdr:
* Added ACTNUM as keyword to Eclipse3DProperties
* Removed certain checks from EclipseGrid that are no longer of use
* Added tests to EclipseGridTests that test boxed ACTNUM and activeCells
* Made EclipseGrid non-const in EclipseState since we must run resetACTNUM
* Fixed a minor typo in an error message in EclipseGrid
* Fixed typo in GridProperty: Deckeyword -> DeckKeyword
* Fixed two errors in tests recently introduced relating to Ecl3DProps
* Removed test in EclipseGridTests; We now accept too small ACTNUM vectors
* Added ACTNUM test for GridProperties

### DIFF
--- a/opm/parser/eclipse/EclipseState/SimulationConfig/tests/ThresholdPressureTest.cpp
+++ b/opm/parser/eclipse/EclipseState/SimulationConfig/tests/ThresholdPressureTest.cpp
@@ -173,140 +173,102 @@ const std::string& inputStrMissingPressure = "RUNSPEC\n"
                                          "/\n"
                                          "\n";
 
-
-
-
-static DeckPtr createDeck(const ParseContext& parseContext , const std::string& input) {
+static DeckPtr createDeck(const ParseContext& parseContext, const std::string& input) {
     Opm::Parser parser;
-    return parser.parseString(input , parseContext);
+    return parser.parseString(input, parseContext);
 }
 
-static Eclipse3DProperties getEclipseProperties(const Deck& deck)
+/// Setup fixture
+struct Setup
 {
-    TableManager* tm = new TableManager(deck);
-    EclipseGrid* eg = new EclipseGrid(3, 3, 3);
+    ParseContext parseContext;
+    DeckPtr deck;
+    TableManager tablemanager;
+    EclipseGrid grid;
+    Eclipse3DProperties props;
+    ThresholdPressure threshPres;
 
-    // We are intentionally leaking memory here.  To save typing three lines
-    // every time we need an Eclipse3DProperties object, we use this helper
-    // function.  We need pointers so that tm and eg do not go out of scope.
-    return Eclipse3DProperties(deck, *tm, *eg);
-}
+    explicit Setup(const std::string& input) :
+            parseContext(),
+            deck(createDeck(parseContext, input)),
+            tablemanager(*deck),
+            grid(10, 3, 4),
+            props(*deck, tablemanager, grid),
+            threshPres(parseContext, *deck, props)
+    {
+    }
+    explicit Setup(const std::string& input, ParseContext& parseContext) :
+            deck(createDeck(parseContext, input)),
+            tablemanager(*deck),
+            grid(10, 3, 4),
+            props(*deck, tablemanager, grid),
+            threshPres(parseContext, *deck, props)
+    {
+    }
 
+};
 
 BOOST_AUTO_TEST_CASE(ThresholdPressureDeckHasEqlnum) {
-    ParseContext parseContext;
-    DeckPtr deck = createDeck(parseContext , inputStrWithEqlNum);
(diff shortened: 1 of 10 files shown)


joakim-hove:
jenkins build this please

opm/parser/eclipse/EclipseState/SimulationConfig/tests/ThresholdPressureTest.cpp
```diff
@@ -181,13 +181,10 @@ static DeckPtr createDeck(const ParseContext& parseContext , const std::string&
     return parser.parseString(input , parseContext);
 }
 
-static Eclipse3DProperties getEclipseProperties(const Deck& deck)
-{
-    static const EclipseGrid eclipseGrid( 3, 3, 3 );
-    return Eclipse3DProperties( deck, TableManager( deck ), eclipseGrid );
+static Eclipse3DProperties getEclipseProperties(const Deck& deck) {
+    return Eclipse3DProperties(deck, *new TableManager(deck), *new EclipseGrid(3, 3, 3));
```
joakim-hove: Why the `new` here?

atgeirr: I agree, also: this is guaranteed to leak memory, and must be avoided.

jokva: It's in a test snippet, so the leak doesn't matter much. Anyway, the reason for the new is because the `static getEclipseProperties` is a shorter initialisation of a very common configuration used in the tests, but `3DProperties` relies on `TableManager` and `EclipseGrid` being alive. Instead of mandating 3 objects being constructed every time, just constructing-and-leaking here was deemed acceptable (by me).

atgeirr: Now the link (in the email) to this comment failed to work due to rebasing. Just saying...

I think this is unacceptable even if its "just a test". Also, tests are used as guidelines and examples for people who want to use a feature, and we run the risk of teaching very bad habits.

pgdr: @atgeirr I didn't rebase, but I did amend the change into the previous commit.

When it comes to the memory leak, I've added a comment specifying that it leaks.  It's a tradeoff between leaking one place, and needing something like 10--16 new references in one of the tests, and a couple of new references in the other tests.  If we prefer the latter, I can do it, but it really clutters up the tests, and make them much less readable, thus also less useful as guidelines and examples.

jokva: The other test code could really use some love if they're to be good for educational purposes.

Patches welcome.

atgeirr: I stil think it is not good practice to leak (anywhere), and it would not clutter up the test, but make them more useful (in the sense of telling the user how to use the class). I made a quick mockup here:
```
struct Setup
{
    ParseContext parseContext;
    DeckPtr deck;
    TableManager tablemanager;
    EclipseGrid grid;
    Eclipse3DProperties props;
    SimulationConfig simulationConfig;

    explicit Setup(const std::string& input)
        : parseContext(),
          deck(createDeck(parseContext, input)),
          tablemanager(*deck),
          grid(10, 3, 4),
          props(*deck, tablemanager, grid),
          simulationConfig(parseContext, *deck, props)
    {
    }
};
```
With this Setup class (it is close to what is usually called a fixture, but not quite since it will be different from case to case) the tests are a little simpler, too:
```
BOOST_AUTO_TEST_CASE(SimulationConfigGetThresholdPressureTableTest) {
    Setup s(inputStr);
    SimulationConfigConstPtr simulationConfigPtr;
    BOOST_CHECK_NO_THROW(simulationConfigPtr = std::make_shared
            <const SimulationConfig>(s.parseContext, *s.deck, s.props));
}


BOOST_AUTO_TEST_CASE(SimulationConfigNOTHPRES) {
    Setup s(inputStr_noTHPRES);
    BOOST_CHECK( !s.simulationConfig.hasThresholdPressure() );
}
```

I do not think this is a messier than leaking.

pgdr: @atgeirr Removed leaks, replaced static function with fixture.  Comments?

joakim-hove:
> Now the link (in the email) to this comment failed to work due to rebasing. Just saying...

Ahhh - I get it right sometimes!

PR #769, opened 2016-04-19
Renamed getEclipseGrid to getInputGrid and changed EclipseGrid constructor

pgdr:
This commit introduces two changes:
* EclipseState.getEclipseGrid has been changed to getInputGrid
* EclipseGrid---ignoring ACTNUM in Deck as it now does---has a new constructor, one that explicitly (but default `nullptr`) takes an `const int* actnum`.

Relevant PR's from other repos:
* OPM/opm-output#18
* OPM/opm-material#141
* OPM/opm-core#1002
* OPM/opm-simulators#657
* OPM/ewoms#55

### DIFF
--- a/opm/parser/eclipse/EclipseState/EclipseState.cpp
+++ b/opm/parser/eclipse/EclipseState/EclipseState.cpp
@@ -54,12 +54,12 @@ namespace Opm {
         m_deckUnitSystem(    deck->getActiveUnitSystem() ),
         m_parseContext(      parseContext ),
         m_tables(            *deck ),
-        m_eclipseGrid(       std::make_shared<EclipseGrid>(deck)),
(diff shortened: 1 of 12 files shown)


atgeirr:
I think there should be an update for opm-grid as well. Consider dune/grid/cpgrid/readEclipseFormat.cpp line 111:
```
        Opm::EclipseGridConstPtr ecl_grid(new Opm::EclipseGrid(deck));
```
This will no longer work correctly (with actnum), right?

pgdr:
@atgeirr one could argue that it never did (hence the fix), but you are correct, it now _completely_ ignores `ACTNUM`.

atgeirr:
Then that must be adressed before this sequence can be merged. I think that is the last issue from me on this.

pgdr:
PR here: OPM/opm-grid#213

PR #770, opened 2016-04-20
Parser::parse functions and ParseContext is member of state, allowing for default constructor

pgdr:
ParseContext is member of state, allowing for default constructor

* ParseContext is now a true member of EclipseState
* ParseContext is copied into EclipseState so we no longer share reference
* Constructor now has default ParseContext argument
* Updated tests to reflect this, removed a test that checked ref equality
* Added static methods for Parser for constructing grid and state
* Added new ParseContext MISSING_SECTIONS with which we cannot construct state
* parse functions take ref over shared_ptr.  removed duplicate comment

### DIFF
--- a/opm/parser/eclipse/EclipseState/EclipseState.cpp
+++ b/opm/parser/eclipse/EclipseState/EclipseState.cpp
@@ -40,6 +40,7 @@
 #include <opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp>
 #include <opm/parser/eclipse/EclipseState/SimulationConfig/SimulationConfig.hpp>
 #include <opm/parser/eclipse/EclipseState/Tables/TableManager.hpp>
+#include <opm/parser/eclipse/Parser/ParseContext.hpp>
 #include <opm/parser/eclipse/Parser/ParserKeywords/M.hpp>
 #include <opm/parser/eclipse/Units/Dimension.hpp>
 #include <opm/parser/eclipse/Units/UnitSystem.hpp>
@@ -49,11 +50,11 @@
 namespace Opm {
 
 
-    EclipseState::EclipseState(std::shared_ptr<const Deck> deckptr, const ParseContext& parseContext) :
+    EclipseState::EclipseState(std::shared_ptr<const Deck> deckptr, const ParseContext parseContext) :
         EclipseState(*deckptr, parseContext)
     {}
 
-    EclipseState::EclipseState(const Deck& deck, const ParseContext& parseContext) :
+    EclipseState::EclipseState(const Deck& deck, const ParseContext parseContext) :
         m_deckUnitSystem(    deck.getActiveUnitSystem() ),
         m_parseContext(      parseContext ),
         m_tables(            deck ),
--- a/opm/parser/eclipse/EclipseState/EclipseState.cpp
+++ b/opm/parser/eclipse/EclipseState/EclipseState.cpp
@@ -50,11 +50,11 @@
 namespace Opm {
 
 
-    EclipseState::EclipseState(std::shared_ptr<const Deck> deckptr, const ParseContext parseContext) :
+    EclipseState::EclipseState(std::shared_ptr<const Deck> deckptr, ParseContext parseContext) :
         EclipseState(*deckptr, parseContext)
     {}
 
-    EclipseState::EclipseState(const Deck& deck, const ParseContext parseContext) :
+    EclipseState::EclipseState(const Deck& deck, ParseContext parseContext) :
         m_deckUnitSystem(    deck.getActiveUnitSystem() ),
         m_parseContext(      parseContext ),
         m_tables(            deck ),
--- a/opm/parser/eclipse/EclipseState/EclipseState.hpp
+++ b/opm/parser/eclipse/EclipseState/EclipseState.hpp
@@ -33,6 +33,7 @@
 #include <opm/parser/eclipse/EclipseState/Schedule/ScheduleEnums.hpp>
 #include <opm/parser/eclipse/EclipseState/Tables/TableManager.hpp>
 #include <opm/parser/eclipse/Parser/MessageContainer.hpp>
+#include <opm/parser/eclipse/Parser/ParseContext.hpp>
 
 namespace Opm {
 
@@ -68,10 +69,10 @@ namespace Opm {
             AllProperties = IntProperties | DoubleProperties
         };
 
-        EclipseState(const Deck& deck , const ParseContext& parseContext);
+        EclipseState(const Deck& deck , const ParseContext parseContext = ParseContext());
 
         /// [deprecated]
-        EclipseState(std::shared_ptr< const Deck > deck , const ParseContext& parseContext);
+        EclipseState(std::shared_ptr< const Deck > deck , const ParseContext parseContext = ParseContext());
 
         const ParseContext& getParseContext() const;
         std::shared_ptr< const Schedule > getSchedule() const;
@@ -129,7 +130,7 @@ namespace Opm {
 
 
         const UnitSystem& m_deckUnitSystem;
-        const ParseContext& m_parseContext;
+        const ParseContext m_parseContext;
         const TableManager m_tables;
         std::shared_ptr<EclipseGrid> m_inputGrid;
         Eclipse3DProperties m_eclipseProperties;
--- a/opm/parser/eclipse/EclipseState/EclipseState.hpp
+++ b/opm/parser/eclipse/EclipseState/EclipseState.hpp
@@ -69,10 +69,10 @@ namespace Opm {
             AllProperties = IntProperties | DoubleProperties
         };
 
-        EclipseState(const Deck& deck , const ParseContext parseContext = ParseContext());
+        EclipseState(const Deck& deck , ParseContext parseContext = ParseContext());
 
         /// [deprecated]
-        EclipseState(std::shared_ptr< const Deck > deck , const ParseContext parseContext = ParseContext());
+        EclipseState(std::shared_ptr< const Deck > deck , ParseContext parseContext = ParseContext());
 
         const ParseContext& getParseContext() const;
         std::shared_ptr< const Schedule > getSchedule() const;
--- a/opm/parser/eclipse/IntegrationTests/IOConfigIntegrationTest.cpp
+++ b/opm/parser/eclipse/IntegrationTests/IOConfigIntegrationTest.cpp
@@ -299,11 +299,7 @@ BOOST_AUTO_TEST_CASE( NorneResttartConfig ) {
     rptConfig.push_back( std::make_tuple(241 , true , boost::gregorian::date( 2006,10,10)) );
 
 
-    ParseContext parseContext;
-    ParserPtr parser(new Parser());
-    DeckConstPtr deck = parser->parseFile("testdata/integration_tests/IOConfig/RPTRST_DECK.DATA", parseContext);
-    EclipseState state( deck , parseContext );
-
+    auto state = Parser::parse("testdata/integration_tests/IOConfig/RPTRST_DECK.DATA");
     verifyRestartConfig(state.getIOConfigConst(), rptConfig);
 }
 
--- a/opm/parser/eclipse/Parser/Parser.hpp
+++ b/opm/parser/eclipse/Parser/Parser.hpp
@@ -27,6 +27,7 @@
 #include <vector>
 #include <boost/filesystem.hpp>
 
+#include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
 #include <opm/parser/eclipse/Parser/ParserKeyword.hpp>
 #include <opm/parser/eclipse/Utility/Stringview.hpp>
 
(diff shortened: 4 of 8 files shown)


joakim-hove:
Thank you - this was interesting. I think it is difficult to find the right balance between explicit (and typically verbose) and simple to use. Without really thinking it through my instinctive reaction is to go for the explicit and verbose solution; and come on - we are paid to do this! Seriously I guess the important question is to what extent it is simple/possible to misunderstand the API and use it incorrectly. If the API can easily be used wrong[1] I think explicit is good.

For the current PR I do not think the API is easily misunderstood; so in this case I guess simplicity should win. But I will let the PR linger here for some time, and invite other developers to voice an opinion on the general matter. 

To take it one step further I think this PR should be updated with the same default behavior for the `Parser::parseFile( )` and `Parser::parseString( )` methods.

[1]: Of course if possible the best thing would be to fix the API.

joakim-hove:
Inspired by @jokva,  and in the same line of simplification - we could create static methods which instantiate and subsequently discard a parser:

```C++
Deck Parser::createDeck( const std::string& inputFile , ParseContext parseContext = ParseContext()) {
    Parser parser;
    return parser.parseFile( inputFile , parseContext );
}

EclipseState Parser::createState(const std::string& inputFile , ParseContext parseContext = ParseContext()) {
    auto deck = Parser::createDeck( inputFile , parseContext );
    return EclipseState( deck );
}
```

joakim-hove:
>but the motivation was Atgeirr's comment that to get a grid you need to (1) create deck, (2) create ParseContext, (3) create EclipseState, (4) state.getInputGrid(). 

Actually it starts with a (0) create parser.

opm/parser/eclipse/EclipseState/EclipseState.cpp
```diff
@@ -43,16 +43,17 @@
...
+#include <opm/parser/eclipse/Parser/ParseContext.hpp>
 #include <opm/parser/eclipse/Units/Dimension.hpp>
 #include <opm/parser/eclipse/Units/UnitSystem.hpp>
 #include <opm/parser/eclipse/Parser/MessageContainer.hpp>
 
 
 namespace Opm {
 
-    EclipseState::EclipseState(std::shared_ptr<const Deck> deck, const ParseContext& parseContext) :
+    EclipseState::EclipseState(std::shared_ptr<const Deck> deck, const ParseContext parseContext) :
```
jokva: It's pretty pointless to pass something as a const copy (because the copy is local anyway, and modifications won't bubble outwards).

opm/parser/eclipse/EclipseState/EclipseState.cpp
```diff
@@ -43,16 +43,17 @@
...
         m_deckUnitSystem(    deck->getActiveUnitSystem() ),
-        m_parseContext(      parseContext ),
+        m_parseContext(      parseContext ), // copying the reference
```
jokva: It's not quite a reference :> Regardless, I think it's nice that when you intend to copy something like this that the argument itself is a value (as it is now), and that sufficiently communicates whats going on.

opm/parser/eclipse/EclipseState/EclipseState.hpp
```diff
@@ -66,7 +67,7 @@ namespace Opm {
             AllProperties = IntProperties | DoubleProperties
         };
 
-        EclipseState(std::shared_ptr< const Deck > deck , const ParseContext& parseContext);
+        EclipseState(std::shared_ptr< const Deck > deck , const ParseContext parseContext = ParseContext());
```
jokva: The default argument is redundant when ParseContext is a member - it would be default constructed anyway. In this case it's better to offer two constructors.

jokva:
>Inspired by @jokva, and in the same line of simplification - we could create static methods which instantiate and subsequently discard a parser:

Free function - they only use the public interface of `Parser` anyway and its meaningless for it to be a part of parser (for other than namespacing reasons). If that's the intention then it's all fine obviously.

opm/parser/eclipse/EclipseState/EclipseState.hpp
joakim-hove: > The default argument is redundant when ParseContext is a member

Good point

joakim-hove:
> for other than namespacing reasons

Well - the suggestion to use a static method was all about namespacing.

atgeirr:
(This is to the discussion about simple free or static methods for simplifying opm-parser usage.)

One idea could be to have a facade class for all things the parser creates. Essentially you'd give it the deck filename and ParseContext (with default) as a constructor argument and it creates the Parser, Deck and Eclipsestate. These objects (and the EclipseGrid) are then available by simple accessor methods. It would make for a simple one-stop solution while simultaneously being flexible enough for any application. What do you think?

jokva:
It doesn't need `Parser` (which is rather uninteresting once you've produced your `Deck`. `Deck`-use too is discouraged for most applications, so a short `EclipseState parse( ... )` should do the trick.

joakim-hove:
> @atgeirr: It would make for a simple one-stop solution while simultaneously being flexible enough for any application. What do you think?

> @jokva: so a short EclipseState parse( ... ) should do the trick.

I agree with @jokva here - when "the harm is done" we don't really neither the `Parser` nor the `ParseContext`; the only products from the parsing process are the `Deck` and `EclipseState` - and we are working towards a shiny future were only the `EclipseState` is needed.

> One idea could be to have a facade class for all things the parser creates.

 Actually the `EclipseState` is in my opinion quite close to such a facade.

atgeirr:
So an EclipseState constructor taking filename and optional ParseContext? That sounds quite nice actually. There are parts of the code that need direct Deck access and could not use this, although the need is reduced for every little part that gets into the EclipseState.

joakim-hove:
> So an EclipseState constructor taking filename and optional ParseContext? 

In principle yes; whether it is a static method on the Parser (my favorite), an `EclipseState` constructor or a free function is an open question though.

opm/parser/eclipse/IntegrationTests/IOConfigIntegrationTest.cpp
```diff
@@ -298,12 +298,7 @@ BOOST_AUTO_TEST_CASE( NorneResttartConfig ) {
     rptConfig.push_back( std::make_tuple(240 , true , boost::gregorian::date( 2006,10,1)) );
     rptConfig.push_back( std::make_tuple(241 , true , boost::gregorian::date( 2006,10,10)) );
 
-
-    ParseContext parseContext;
-    ParserPtr parser(new Parser());
-    DeckConstPtr deck = parser->parseFile("testdata/integration_tests/IOConfig/RPTRST_DECK.DATA", parseContext);
-    EclipseState state( deck , parseContext );
-
+    auto state = Parser::parse("testdata/integration_tests/IOConfig/RPTRST_DECK.DATA");
```
pgdr: @joakim-hove This is perhaps the most typical example where such a parser function could be nice to have.

pgdr:
I just want to stress again that this is _not ready to be merged_ and (for the moment) intended for educational purposes.

joakim-hove:
> I just want to stress again that this is not ready to be merged and (for the moment) intended for educational purposes.

Sure. PS: you can update the PR description at the very top!

opm/parser/eclipse/EclipseState/EclipseState.hpp
```diff
@@ -66,7 +67,8 @@ namespace Opm {
             AllProperties = IntProperties | DoubleProperties
         };
 
-        EclipseState(std::shared_ptr< const Deck > deck , const ParseContext& parseContext);
+        EclipseState(std::shared_ptr< const Deck > );
```
jokva: Since this constructor is new, can we use `Deck&` over `std::shared_ptr< Deck >`?

opm/parser/eclipse/Parser/Parser.hpp
```diff
@@ -101,6 +102,29 @@ namespace Opm {
             addParserKeyword( std::unique_ptr< ParserKeyword >( new T ) );
         }
 
+        // below are some helper functions
+
+        static EclipseState parse(std::shared_ptr<const Deck> deck, const ParseContext& context = ParseContext());
```
jokva: Unsure about this - with the `Deck` the input is already parsed, and his is *very* similar to the `EclipseState` constructor. Does it have to exist?

pgdr: I am not sure, but I can try.  It would imply that EclipseGrid can be constructed from a reference.  I don't think it is impossible, but I don't know how far that would propagate.  Investigating now.

opm/parser/eclipse/EclipseState/EclipseState.hpp
pgdr: I am not sure, but I can try.  It would imply that EclipseGrid can be constructed from a reference.  I don't think it is impossible, but I don't know how far that would propagate.  Investigating now.

opm/parser/eclipse/Parser/Parser.hpp
pgdr: It doesn't have to exist, but I added it primarily to have a single line of attack;  If you want to create an EclipseState, no matter if you have deck(ptr), context, file or string, you would resort to `Parser::parse`.

I'm open for removing it.

opm/parser/eclipse/EclipseState/EclipseState.hpp
pgdr: EclipseGrid and EclipseState now have constructors accepting references to Deck instead of shared_ptr's.

It took some refactoring;  I had to tweak the constructors of NNC, Schedule, and InitConfit to accept Deck& instead of DeckPtr as well.

But all in all, I think this is a good change that we should cherry-pick if we don't want this entire PR.

andlaus: > Since this constructor is new, can we use `Deck&` over `std::shared_ptr< Deck >`?

don't forget to mark it `explicit`. (that's also a good idea in its current form.)

opm/parser/eclipse/EclipseState/EclipseState.hpp
```diff
@@ -66,7 +67,8 @@ namespace Opm {
...
+        EclipseState(std::shared_ptr< const Deck > , ParseContext );
```
andlaus: why does ParseContext needs to be copied here instead of passing a const reference?

opm/parser/eclipse/EclipseState/EclipseState.hpp
```diff
@@ -123,7 +125,7 @@ namespace Opm {
 
 
         const UnitSystem& m_deckUnitSystem;
-        const ParseContext& m_parseContext;
+        const ParseContext m_parseContext;
```
andlaus: even if this object becomes an instance of `ParseContext`, the constructor can still take a const reference.

jokva: It can, but it's mostly moot because:
* Copying it is rather cheap
* It indicates to both users and compilers that this will in fact be copied and stored
* The copy is only done once because of moves.

opm/parser/eclipse/IntegrationTests/IOConfigIntegrationTest.cpp
andlaus: I don't like the name of the method (`parse()`): IMO this name applies to parsing in-memory data as equally as to parsing files, network resources and god knows what. what's wrong with `parseFile()`, `parseString()`, etc?

pgdr: Nothing wrong, but the main reason I chose `parse` was because there are already functions with the names you mention, with the same input parameters, but different return value.  One possibility would be to hijack these names, but I think they are very widely used; `DeckPtr parseFile` and `DeckPtr parseString` etc.

opm/parser/eclipse/EclipseState/EclipseState.hpp
andlaus: I agree that this is certainly not gonna affect performance in any measurable way. I think it is a bad habit which should not creep in unless there is a specific reason because sometimes it _does_ matter...

> - The copy is only done once because of moves.

that's actually only true if the `ParseContext` contains a lot of data allocated on the stack (i.e., copying it is probably a relatively expensive enterprise contradicting you first statement). If the compiler could inline the constructor, it *might* be able to do some copy-elision tricks, but with the current design (split sources and headers) that's very unlikely IMO unless LTO is explicitly enabled. (it is currently not enabled by default because it caused some obscure problems in the past.)

jokva: But since we're targeting a copy *anyway* it's most certainly not bad, and should allow some more optimisations because neither translation unit will have to care about aliasing.

Regarding your second point, the constructor now knows that its argument is a copy and will go out of scope with itself, so it should realise there's a move opportunity here, ensuring the actual copy is only performed once.

andlaus: > Regarding your second point, the constructor now knows that its argument is a copy and will go out of scope with itself, so it should realise there's a move opportunity here, ensuring the actual copy is only performed once.

If I didn't understand this completely incorrectly, moves work by "capturing" the internal pointers of the source object in C++ (if the RHS is a rvalue reference). I.e., they're not "magical" and in particular they won't work if the object to be moved does not feature any internal pointers...

jokva: That's true, moves are simply pointer-copies that are "marked" for non-delete when the old host object goes out of scope. But if an object doesn't point to any larger blobs of memory it's typically tiny anyway, meaning copies aren't expensive because most objects only contain a handful of tiny things (typically numbers). Using a reference-then-copy doesn't change that, and also doesn't enable copy elision.

And at the end of the day we're going to copy here sooner or later, and if that is done in the constructor argument some opportunities are enabled, but in the worse case we should have no difference compared to reference-then-copy.

andlaus: > moves are simply pointer-copies that are "marked" for non-delete 

the pointers of the source object are not marked anything, they are simply "ripped out" of the source object, i.e. the pointers are usually set to NULL (which works with delete, but don't try to use the source object after this ;)).

>  But if an object doesn't point to any larger blobs of memory it's typically tiny anyway, meaning copies aren't expensive because most objects only contain a handful of tiny things (typically numbers).

true. in the case at hand I don't think it would affect performance even if `ParseContext` was an extremely complex beast like `Deck`. On the other hand, if that happens for objects that are used in tight loops, you *will* have a performance impact. (which is what I mean with "bad habits".)

> And at the end of the day we're going to copy here sooner or later, and if that is done in the constructor argument some opportunities are enabled, but in the worse case we should have no difference compared to reference-then-copy.

I agree that this object needs to be independent of the scope of the objects passed. I still think that using a const reference argument for the constructor is slightly nicer, though...

jokva: > i.e. the pointers are usually set to NULL

A way of marking them for non-deletion. It's 100% implementation defined and can vary between objects though.

> true. in the case at hand I don't think it would affect performance even if ParseContext was an extremely complex beast like Deck. On the other hand, if that happens for objects that are used in tight loops, you will have a performance impact. (which is what I mean with "bad habits".)

But if they're to be copied anyway you can't avoid this cost, regardless of reference passing. If a copy isn't needed then the argument should obviously be a reference.

andlaus: > But if they're to be copied anyway you can't avoid this cost, regardless of reference passing. If a copy isn't needed then the argument should obviously be a reference.

In my personal opinion, objects of non-primitive types should always passed as const references and if they're required after the argument object has potentialy gone out of scope, the called function is responsible for copying it. some people (@atgeirr!?) are even more strict here and require even primitive by-value objects to be marked const.

Anyway, if you and @joakim-hove prefer the by-value call here, I'm certainly not going to object...

joakim-hove: > Anyway, if you and @joakim-hove prefer the by-value call here, I'm certainly not going to object...

I'm just a passenger ;-)

PR #772, opened 2016-04-22
Remove OpmLog usages in Parser folder.

qilicun:


joakim-hove:
OK; is the conversion process complete in Opm-Parser with this PR?

pgdr:
Perhaps it is better to make `MessageContainer` a `mutable` member of Deck rather than remove `const` everywhere?

jokva:
Making `Deck` mutable here is not acceptable.

joakim-hove:
> Perhaps it is better to make MessageContainer a mutable member of Deck rather than remove const everywhere?

Yes - please do this instead.

qilicun:
>is the conversion process complete in Opm-Parser with this PR?

Apart from the python folder, this is ture.

joakim-hove:
> Apart from the python folder, this is ture.

OK - good; leave that for now.

joakim-hove:
> OK - good; leave that for now.

Ehhh; just remove everything `OpmLog` related from the Python folder.

qilicun:
The process is finished by this PR, all the OpmLog functionalities are replaced with MessageContainer in opm-parser.

opm/parser/eclipse/Deck/Section.hpp
```diff
@@ -47,7 +47,7 @@ class Section : public DeckView {
 
         // returns whether the deck has all mandatory sections and if all sections are in
         // the right order
-        static bool checkSectionTopology(const Deck& deck,
```
pgdr: Why is `Deck` not const here anymore?  Did you make `MessageContainer` mutable in `Deck`?

qilicun: I tried, but mutable does not work here, I don't know the reason, did you try that?

jokva: The member variable must be marked mutable, not the argument. http://en.cppreference.com/w/cpp/keyword/mutable

qilicun: >The member variable must be marked mutable,

Yes, that in ``Deck.hpp`` ``mutable MessageContainer m_messages`` is made,  the error happened in ``checkDeck.hpp``.

``
diff --git a/opm/parser/eclipse/Deck/Deck.hpp b/opm/parser/eclipse/Deck/Deck.hpp
index a4e9815..99b0fb3 100644
--- a/opm/parser/eclipse/Deck/Deck.hpp
+++ b/opm/parser/eclipse/Deck/Deck.hpp
@@ -121,7 +121,7 @@ namespace Opm {
             void initUnitSystem() const;
 
-            MessageContainer m_messageContainer;
+            mutable MessageContainer m_messageContainer;
 
             mutable std::unique_ptr< UnitSystem > activeUnits;
             mutable std::unique_ptr< UnitSystem > defaultUnits;
diff --git a/opm/parser/eclipse/Deck/Section.hpp b/opm/parser/eclipse/Deck/Section.hpp
index 0e85143..a21f4b5 100644
--- a/opm/parser/eclipse/Deck/Section.hpp
+++ b/opm/parser/eclipse/Deck/Section.hpp
@@ -47,7 +47,7 @@ class Section : public DeckView {
 
         // returns whether the deck has all mandatory sections and if all sections are in
         // the right order
-        static bool checkSectionTopology(Deck& deck,
+        static bool checkSectionTopology(const Deck& deck,
                                          const Parser&,
                                          bool ensureKeywordSectionAffiliation = false);
``

Error:
``
In function ‘bool Opm::checkDeck(Opm::DeckConstPtr, Opm::ParserConstPtr, size_t)’:
EclipseState/checkDeck.cpp:36:43: error: passing ‘const Opm::Deck’ as ‘this’ argument of ‘Opm::MessageContainer& Opm::Deck::getMessageContainer()’ discards qualifiers [-fpermissive]
                 deck->getMessageContainer().warning(keyword.getFileName(), msg,keyword.getLineNumber());
                                           ^
``

atgeirr: The getMessageContainer() method must be const. The compiler should not complain about that if the message container member is mutable (if not, you would not be allowed to return the non-const reference in a const method).

jokva: But when the `MessageContainer&` is returned it is no longer qualified as `mutable`, is it?

atgeirr: What do you mean? If you do `MessageContainer& mc = deck->getMessageContainer()` mc is non-const i.e. mutable? Just tested a tiny little program to be sure:
```
#include <iostream>

class M
{
public:
    M(int f) : foo_(f) {}
    int& getFoo() const { return foo_; }
private:
    mutable int foo_;
};

int main()
{
    M m(7);
    const int foo1 = m.getFoo();
    m.getFoo() = 8;
    const int foo2 = m.getFoo();
    std::cout << foo1 << ' ' << foo2 << std::endl;
}
```
The program compiles and runs as expected.

jokva: Oh, you mean making the return type non-const, sure. Yeah, that (obviously) works.

qilicun:
Thanks for the discussion, that is really helpful. Now, it should works, the ``const`` comes back.

opm/parser/eclipse/Deck/Section.hpp
jokva: I would like to point out that doing this *is* contract breaking and violates reasonable assumptions w.r.t. Deck constness behaviour.

atgeirr: Yes, I agree and do not in any way insist on this solution, just trying to help using the C++ toolchest...

qilicun: I was confused, is ``mutable`` not accept here? The originally implementation is not using ``mutable``.

jokva: It breaks the contract established by `Deck`'s interface by performing externally visible modifications to a `const`'d access path.

qilicun: If mutable is not needed, I will go back to the previous commit.

PR #779, opened 2016-04-26
Adds Group::getWells method

jokva:
Add support for querying what wells are in a given group at a certain
time step.


jokva:
This is a required support feature for Group-related keywords in opm-output.

PR #780, opened 2016-04-26
Ensure that CMAKE_CXX_FLAGS are propagated.

joakim-hove:
Make sure CXX_FLAGS are passed on. 

Should I fetch one of my brown paper bags? Comments from e.g @akva2 or other build specialists welcome.


akva2:
seems fine. no other way to do it if we are changing the flags in the first place.

joakim-hove:
> seems fine. no other way to do it if we are changing the flags in the first place.

Good - thank you; merging.

PR #781, opened 2016-04-26
Emit compiler warning when including Deck.hpp

joakim-hove:
This commit is a first step towards reducing the use of the Deck api
outside of opm-parser. Downstream modules should preferably use the
EclipseState api which is richer in features, and also easier to use
correctly.

The current form of the code is temporary, in the future we will remove
the OPM_PARSER_DECK_API_WARNING compile guard.


joakim-hove:
OK - no comments received; merging.

Thou shall get warnings!

PR #783, opened 2016-04-27
Added method getIJK in EclipseGrid

pgdr:
This PR introduces one new method on `EclipseGrid`: essentially reimplementing (instead of exposing) `ecl_grid.h` function
```C
void ecl_grid_get_ijk1(const ecl_grid_type * , int global_index , int *, int * , int *);
```

### DIFF
--- a/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
+++ b/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
@@ -254,6 +254,17 @@ namespace Opm {
         return (i + j * getNX() + k * getNX() * getNY());
     }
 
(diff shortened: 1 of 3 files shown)


bska:
I personally prefer to do this calculation the other way around
```C++
i = g % Nx;  g /= Nx;
j = g % Ny;  g /= Ny;
k = g % Nz;
```

pgdr:
@bska That's much nicer, I agree.  Perhaps we could fix that later.  I now realize that I could've added one more thing to the PR;  `assertGlobalIndex`.

As it is now, it will gladly give you the i,j,k for arbitrarily high indices.  If we call `assertGlobalIndex` before computing i,j,k, we will throw `invalid_argument`.

PR #784, opened 2016-04-28
Clean eclipse writer

kjellkongsvik:


opm/parser/eclipse/Deck/Deck.hpp
```diff
@@ -113,17 +115,24 @@ namespace Opm {
...
 
+
             std::vector< DeckKeyword > keywordList;
             mutable MessageContainer m_messageContainer;
 
             mutable std::unique_ptr< UnitSystem > activeUnits;
             mutable std::unique_ptr< UnitSystem > defaultUnits;
+
+            std::string m_dataFile;
+            bool m_hasDataFile;
```
pgdr: 1. do we need bool and string, or can you have empty string meaning no data file?
2. what is the intention of `m_dataFile`?  Can you provide a comment, as in, if I set dataFile to be a different file than what the deck was constructed from, can I expect anything to happen except that dataFile is now a different path?
3.  What if I call setDataFile to a something which is not a valid POSIX path, or to somewhere I don't have read/write access?

opm/parser/eclipse/Parser/Parser.cpp
```diff
@@ -229,6 +229,7 @@ namespace Opm {
         parserState->openRootFile( dataFileName );
         parseState(parserState);
         applyUnitsToDeck(*parserState->deck);
+        parserState->deck->setDataFile(dataFileName);
```
joakim-hove: Here I would prefer:

1. The `parserState` instance internalizes the rootFilename.
2. The `Deck` constructor gets a `const std::string& inputFile`input argument.

Then we can manage without the `Deck::setDataFile( )` method.

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.hpp
```diff
@@ -158,6 +158,15 @@ namespace Opm {
 
         std::string getRestartFileName(const std::string& restart_base, int report_step, bool output) const;
 
+        std::string getDeckFileName();
```
joakim-hove: Is this required?

opm/parser/eclipse/Parser/Parser.cpp
jokva: This code gets touched by https://github.com/OPM/opm-parser/pull/758

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.hpp
kjellkongsvik: Yes, used by EclipseWriter to get baseName for output

joakim-hove: Well there is a `getBaseName()`method as well - in my opinion that should be the only required method.

opm/parser/eclipse/Deck/Deck.hpp
kjellkongsvik: Good input, simplified with new commit.
1. No data means deck is created from string, not file
2. Now only sets filename in deck after successfully opened by Parser. No check inside deck for validity, should we there as well?
3. This should not happen. see 2.

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.hpp
kjellkongsvik: correct, the test for .DATA suffix can be removed

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
```diff
@@ -46,8 +47,12 @@ namespace Opm {
...
         m_FMTIN(false),
         m_FMTOUT(false),
-        m_eclipse_input_path(input_path),
-        m_ignore_RPTSCHED_RESTART(false){
+        m_ignore_RPTSCHED_RESTART(false),
+        m_deck_filename(input_path),
+        m_output_enabled(true),
+        m_output_dir(".")
+    {
+        m_base_name = boost::filesystem::path(input_path).stem().string();
```
joakim-hove:
```C++
if (!input_path.empty()) {
    boost::filesystem::path path( input_path );
    m_base_name = path.stem().string();
    m_output_dir = path.parent_path().string();
}
```

PR #786, opened 2016-04-28
Add EclipseGrid.cartesianAdjacent and throw if out of bounds

pgdr:
* This PR introduces a method EclipseGrid.cartesianAdjacent that takes two global indices and check whether they are adjacent in the cartesian way, that is, _(i<sub>1</sub>-i<sub>2</sub>)<sup>2</sup> + (j<sub>1</sub>-j<sub>2</sub>)<sup>2</sup> + (k<sub>1</sub>-k<sub>2</sub>)<sup>2</sup> = 1_
* Adds tests for EclipseGrid.cartesianAdjacent
* Rewrote, as per @bska the previously implemented `getIJK`
* Added `assertIJK` and `assertGlobalIndex` in corresponding translate functions

The cartesianAdjacent function will be used in PR relating to output NNC.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
+++ b/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
@@ -249,19 +249,40 @@ namespace Opm {
         return m_minpvValue;
     }
 
+    /// static helper function checking that _one_ coordinate differs with 1
+    bool ijk_are_adj(int i1, int j1, int k1, int i2, int j2, int k2) {
+        return ((i1-i2)*(i1-i2)+
+                (j1-j2)*(j1-j2)+
+                (k1-k2)*(k1-k2)) == 1;
+    }
+
+    // checks cartesian adjacency of global indices g1 and g2
+    bool EclipseGrid::cartesianAdjacent(size_t g1, size_t g2) const {
+        assertGlobalIndex(g1); assertGlobalIndex(g2);
+
+        auto cell1 = getIJK(g1);
+        auto cell2 = getIJK(g2);
+        return ijk_are_adj(cell1[0], cell1[1], cell1[2],
+                           cell2[0], cell2[1], cell2[2]);
+    }
+
+
+
 
     size_t EclipseGrid::getGlobalIndex(size_t i, size_t j, size_t k) const {
+        assertIJK(i, j, k);
         return (i + j * getNX() + k * getNX() * getNY());
     }
 
-    std::array<int, 3> EclipseGrid::getIJK(size_t globalIndex) const {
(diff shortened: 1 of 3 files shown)


opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
```diff
@@ -249,19 +249,40 @@ namespace Opm {
         return m_minpvValue;
     }
 
+    /// static helper function checking that _one_ coordinate differs with 1
+    bool ijk_are_adj(int i1, int j1, int k1, int i2, int j2, int k2) {
```
joakim-hove: Comment says `static` - code says no?

opm/parser/eclipse/EclipseState/Grid/tests/EclipseGridTests.cpp
```diff
@@ -127,6 +127,27 @@ BOOST_AUTO_TEST_CASE(CheckGridIndex) {
...
 
     BOOST_CHECK_EQUAL(17 * 19 * 41, grid.getCartesianSize());
+
+    BOOST_CHECK_THROW(grid.getIJK(17 * 19 * 41), std::invalid_argument);
+
+    BOOST_CHECK_THROW(grid.getGlobalIndex(17, 0, 0), std::invalid_argument);
+    BOOST_CHECK_THROW(grid.getGlobalIndex(0, 19, 0), std::invalid_argument);
+    BOOST_CHECK_THROW(grid.getGlobalIndex(0, 0, 41), std::invalid_argument);
+
+    size_t c1 = grid.getGlobalIndex(7, 11, 13);
```
joakim-hove: Thank you for being so thorough with testing!

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
pgdr: I meant `free`, but it should probably also be a static function, or what?  I don't intend it to be used from anywhere else than `cartesianAdjacent`.

joakim-hove: > but it should probably also be a static function, or what?

I would say so!

jokva: Definitely yes. No reason to make it visible to the linker unless it has to be.

andlaus: static or not, so far the rest of opm-parser uses camelCase (instead_of_underscores) as its naming scheme for methods and classes. is the name `ijk_are_adj()' intentional?

joakim-hove:
OK - I will merge this when it is green again.

PR #788, opened 2016-04-29
Support explicit setting of BaseName

kjellkongsvik:
Needed when reading deck from string and trying to write to disk.
If this is not set there is no baseName to base the file name on.

A _very_ special case. Will only ever be used by tests? Should we actually support this?

### DIFF
--- a/opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
+++ b/opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
@@ -465,7 +465,7 @@ namespace Opm {
         m_output_dir = outputDir;
     }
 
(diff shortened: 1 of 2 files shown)


joakim-hove:
Yes - we should support this, it is not *that* special. I typically see that it could be interesting in a multi realization setup.

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
```diff
@@ -469,4 +469,8 @@ namespace Opm {
         return m_base_name;
     }
```
joakim-hove: const std::string&

PR #792, opened 2016-05-03
Bugfixes

jokva:
A couple of bugs unearthed by clang-asan


atgeirr:
> unearthed by clang-asan

It is good to have good tools! Since you have set it up -- will you look for bugs in other modules?

jokva:
It was first discovered by a (different) gcc, but reproduced locally with clang.

I haven't gotten it set up beyond parser due to current availability of clang-compatible dependencies. Maybe in the future.

Issue #790, opened 2016-05-03
Messages that handled by ParseContext


qilicun:
There are still some messages that handled by ParseContext that not logged in the ``MessageContainer``. The approaches that I can think are:

- Add ``MessageContext`` as argument in ``ParseContext::handleError()`` method.
- Add ``MessageContext`` as member in ``ParseContext`` class.

I think the second way is probably against the ''law'' in opm-parser, so I vote for the first way. 
What do you think?

PR #794, opened 2016-05-04
log messages before handleError.

qilicun:
Messages is handled by ``handleError()`` method  in ``ParseContext`` are not logged in the log file, this PR simply write the messages in the log file before ``handleError()`` is called,  but the best way is probably to modify the ``ParseContext`` class itself so that one function can do both thing,  here is a relevant issue  https://github.com/OPM/opm-parser/issues/790. Any comment is welcome.

### DIFF
--- a/opm/parser/eclipse/Parser/Parser.cpp
+++ b/opm/parser/eclipse/Parser/Parser.cpp
@@ -343,7 +343,7 @@ void ParserState::handleRandomText(const string_view& keywordString ) const {
             << this->current_path()
             << ":" << this->line();
     }
-
+    deck->getMessageContainer().warning(msg.str());
     parseContext.handleError( errorKey , msg.str() );
 }
 
@@ -386,6 +386,7 @@ std::shared_ptr< RawKeyword > createRawKeyword( const string_view& kw, ParserSta
     if( !parser.isRecognizedKeyword( keywordString ) ) {
         if( ParserKeyword::validDeckName( keywordString ) ) {
             std::string msg = "Keyword " + keywordString + " not recognized.";
+            parserState.deck->getMessageContainer().error(msg);
             parserState.parseContext.handleError( ParseContext::PARSE_UNKNOWN_KEYWORD, msg );
             return {};
         }
@@ -432,7 +433,7 @@ std::shared_ptr< RawKeyword > createRawKeyword( const string_view& kw, ParserSta
 
     std::string msg = "Expected the kewyord: " + sizeKeyword.first
                     + " to infer the number of records in: " + keywordString;
-
+    parserState.deck->getMessageContainer().error(msg);
     parserState.parseContext.handleError(ParseContext::PARSE_MISSING_DIMS_KEYWORD , msg );
 
     const auto* keyword = parser.getKeyword( sizeKeyword.first );
--- a/opm/parser/eclipse/Parser/Parser.cpp
+++ b/opm/parser/eclipse/Parser/Parser.cpp
@@ -343,8 +343,7 @@ void ParserState::handleRandomText(const string_view& keywordString ) const {
             << this->current_path()
             << ":" << this->line();
     }
-    deck->getMessageContainer().warning(msg.str());
-    parseContext.handleError( errorKey , msg.str() );
+    parseContext.handleError( errorKey , deck->getMessageContainer() , msg.str() );
 }
 
 void ParserState::openRootFile( const boost::filesystem::path& inputFile) {
@@ -386,8 +385,8 @@ std::shared_ptr< RawKeyword > createRawKeyword( const string_view& kw, ParserSta
     if( !parser.isRecognizedKeyword( keywordString ) ) {
         if( ParserKeyword::validDeckName( keywordString ) ) {
             std::string msg = "Keyword " + keywordString + " not recognized.";
-            parserState.deck->getMessageContainer().error(msg);
-            parserState.parseContext.handleError( ParseContext::PARSE_UNKNOWN_KEYWORD, msg );
+            auto& msgContainer = parserState.deck->getMessageContainer();
+            parserState.parseContext.handleError( ParseContext::PARSE_UNKNOWN_KEYWORD, msgContainer, msg );
             return {};
         }
 
@@ -433,8 +432,8 @@ std::shared_ptr< RawKeyword > createRawKeyword( const string_view& kw, ParserSta
 
     std::string msg = "Expected the kewyord: " + sizeKeyword.first
                     + " to infer the number of records in: " + keywordString;
-    parserState.deck->getMessageContainer().error(msg);
-    parserState.parseContext.handleError(ParseContext::PARSE_MISSING_DIMS_KEYWORD , msg );
+    auto& msgContainer = parserState.deck->getMessageContainer();
+    parserState.parseContext.handleError(ParseContext::PARSE_MISSING_DIMS_KEYWORD , msgContainer, msg );
 
     const auto* keyword = parser.getKeyword( sizeKeyword.first );
     const auto record = keyword->getRecord(0);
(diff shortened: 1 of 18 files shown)


alfbr:
Has anybody looked at this one?

opm/parser/eclipse/Parser/Parser.cpp
```diff
@@ -343,7 +343,7 @@ void ParserState::handleRandomText(const string_view& keywordString ) const {
             << this->current_path()
             << ":" << this->line();
     }
-
+    deck->getMessageContainer().warning(msg.str());
```
pgdr: Here you have access to filename and line number.  Why not add that to the message?

opm/parser/eclipse/Parser/Parser.cpp
```diff
@@ -432,7 +433,7 @@ std::shared_ptr< RawKeyword > createRawKeyword( const string_view& kw, ParserSta
 
     std::string msg = "Expected the kewyord: " + sizeKeyword.first
                     + " to infer the number of records in: " + keywordString;
-
+    parserState.deck->getMessageContainer().error(msg);
```
pgdr: Why is this an error and not a warning?

opm/parser/eclipse/Parser/Parser.cpp
qilicun: ``msg`` is ``std::stringstream``, it have already contain the filename and line number I guess?

opm/parser/eclipse/Parser/Parser.cpp
qilicun: Default setting for ``ParseContext::PARSE_UNKNOWN_KEYWORD`` is ``throw``, so I guess this is an error.

opm/parser/eclipse/Parser/Parser.cpp
jokva: Make it not a `stringstream` then.

opm/parser/eclipse/Parser/Parser.cpp
jokva: That's up to the `ParseContext` to decide.

qilicun: Yes, that's the reason I opened an issue, see https://github.com/OPM/opm-parser/issues/790

joakim-hove:
OK; I have looked briefly at this - and #790. The call which actually adds the message must be in `ParseContext` - otherwise we will not know if we should call `error()` or `warning()`. So my suggestion is that `ParseContext::handleError()` is modified to something like:

```C++
ParseContext::handleError( const std::string& errorKey, MessageContainer& messages, const std::string msg) {

    InputError::Action action = get( errorKey );

    if (action == InputError::WARN) {
         messages.warning( msg ); 
         return Message::Warning;
     } else if (action == InputError::THROW_EXCEPTION) {
         messages.error( msg ); 
         throw std::invalid_argument(errorKey + ": " + msg);
     }

}
```

qilicun:
>So my suggestion is that ParseContext::handleError() is modified to something like:

Ok, this is a proper way to handle messages, will update the PR.

qilicun:
Currently ``handleError()`` are called in the following scope.
```c
./Parser/ParseContext.cpp:    Message::type ParseContext::handleError(
./Parser/ParserRecord.cpp:            parseContext.handleError(ParseContext::PARSE_EXTRA_DATA , msg);
./Parser/Parser.cpp:    parseContext.handleError( errorKey , msg.str() );
./Parser/Parser.cpp:            parserState.parseContext.handleError( ParseContext::PARSE_UNKNOWN_KEYWORD, msg );
./Parser/Parser.cpp:    parserState.parseContext.handleError(ParseContext::PARSE_MISSING_DIMS_KEYWORD , msg );
./EclipseState/SimulationConfig/ThresholdPressure.cpp:                m_parseContext.handleError(ParseContext::INTERNAL_ERROR_UNINITIALIZED_THPRES, msg);
./EclipseState/Schedule/Schedule.cpp:                    parseContext.handleError( ParseContext::UNSUPPORTED_SCHEDULE_GEO_MODIFIER , msg );
./EclipseState/Schedule/Schedule.cpp:                parseContext.handleError( ParseContext::UNSUPPORTED_COMPORD_TYPE , msg );
```

For ``Schedule`` and ``Parser`` that is easy to add another argument for ``handleError()``, but for  ``ParserRecord``, especially for ``ThresholdPressure``  is not that easy. Maybe in the future, we need to log more messages through ``handleError()`` method, more API changes are needed.

joakim-hove:
> For Schedule and Parser that is easy to add another argument for handleError(), but for ParserRecord, especially for ThresholdPressure is not that easy. 

Ahhh - this gets so unwieldy. For the ThresholdPressure I suggest you start by removing the error handling completely - and just throw an exception. This is a quite special case - and we should not implement the API by chasing special cases all over the place. 

The error condition in `ParserRecord` is more common, and should be handled properly. I guess you can just pass a `MessageContainer` reference all the way from the parser scope?

> [....] more API changes are needed.

Yes - that is probably true; this certainly does not feel *Just Right*&trade; yet. For now we should get this through to a working state, and then we can sit back - take a deep breath and rethink things slightly.

jokva:
All these problems is a good indicator that our entire approach here is flawed.

alfbr:
I do not find that comment helpful, specific improvement suggestions are better. Personally I really appreciate Joakim's approach here. Let us get something that works, then we can perfect it.

pgdr:
I know that I'm a bit late to the dinner table now, but @joakim-hove wouldn't it be better to add `messageContainer` to `ParseContext` rather than insisting on throwing on `UNSUPPORTED_INITIAL_THPRES`?

@qilicun If we end up removing `ParseContext` from `ThresholdPressure` you need to remove `INTERNAL_ERROR_UNINITIALIZED_THPRES` from ParseContext, and maybe mention it in the throw.

jokva:
No. `ParseContext` is a **configuration** (or should be anyway) and this would break that.

qilicun:
>I know that I'm a bit late to the dinner table now, but @joakim-hove wouldn't it be better to add messageContainer to ParseContext rather than insisting on throwing on UNSUPPORTED_INITIAL_THPRES?

Discussion is coming back... see the reason https://github.com/OPM/opm-parser/pull/738

joakim-hove:
> I know that I'm a bit late to the dinner table now, but @joakim-hove wouldn't it be better to add messageContainer to ParseContext rather than insisting on throwing on UNSUPPORTED_INITIAL_THPRES?

As @qilicun indicated this has been turned down previously. I can only repeat my previous comment:

> For now we should get this through to a working state, and then we can sit back - take a deep breath and rethink things slightly.

**When** we are rethinking this - one possible idea could be:

1. The configuration part of the object current `ParseContext` object is renamed `ParseConfig`, and the `ParseContext` object gets an immutable `ParseConfig`member.
2. The `ParseContext` gets  a `MessageContainer` member - which is mutable.
3. The `handleError()` method is part of the new `ParseContext` implementation.

That way we would separate immutable configuration and mutable message storage/parsing context better.

alfbr:
Are we ready to merge this one?

pgdr:
I think I missed why `parseContext` is now removed from `ThresholdPressure`.  Can give a short comment on why?

Can you edit the PR comment (on the top) to contain a summary of the changes (and the reasons for them)?

jokva:
Please rebase and beautify the log a bit, too many noisy commits.

joakim-hove:
> When we are rethinking this - one possible idea could be:

> 1 The configuration part of the object current ParseContext object is renamed ParseConfig, [....]
2. The ParseContext gets a MessageContainer member - which is mutable.
3. The handleError() method is part of the new ParseContext implementation.

Or just f... it and use a global variable.

joakim-hove:
OK - I am ready to merge this now. The work from @qilicun has been according to the original plan (where I was one of the orginal offendors), but the chattering classes in Bergen have been a significant hurdle. As I have done several times already I insist we get some experience with this now; but personally i feel that this is not *Just Right&trade;* yet - so I assume we will revisit it.

I have always been *very* reluctant to use global variables, or their glorified cousin Singleton; but I guess that would have been a better solution here.

For some minor git technical reasons I am closing this and merging: https://github.com/OPM/opm-parser/pull/807 instead.

PR #796, opened 2016-05-08
Check deck for ACTNUM keyword.

joakim-hove:
Will check the deck for `ACTNUM` keyword when creating an `EclipseGrid` based on `Deck`. This is how the implementation until recently, and will work nicely in many cases. However - manipulations of `ACTNUM` beyond one global assignment - will be silently ignored.

See also comment in EclipseGrid.cpp

### DIFF
--- a/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
+++ b/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
@@ -115,6 +115,36 @@ namespace Opm {
                EclipseGrid(*deckptr, actnum)
     {}
 
+
+    /*
+      This is the main EclipseGrid constructor, it will inspect the
+      input Deck for grid keywords, either the corner point keywords
+      COORD and ZCORN, or the various rectangular keywords like DX,DY
+      and DZ.
+
+      Actnum is treated specially:
+
+        1. If an actnum pointer is passed in that should be a pointer
+           to 0 and 1 values which will be used as actnum mask.
+
+        2. If the actnum pointer is not present the constructor will
+           look in the deck for an actnum keyword, and use that if it
+           is found. This is a best effort which will work in many
+           cases, but if the ACTNUM keyword is manipulated in the deck
+           those manipulations will be silently lost; if the ACTNUM
+           keyword has size different from nx*ny*nz it will also be
+           silently ignored.
+
+      With a mutable EclipseGrid instance you can later call the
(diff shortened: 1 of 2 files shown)


opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
```diff
@@ -166,6 +196,13 @@ namespace Opm {
         }
         if (actnum != nullptr)
             resetACTNUM(actnum);
+        else {
+            if (deck.hasKeyword<ParserKeywords::ACTNUM>()) {
+                const auto& actnumData = deck.getKeyword<ParserKeywords::ACTNUM>().getIntData();
+                if (actnumData.size() == getCartesianSize())
+                    resetACTNUM( actnumData.data());
```
atgeirr: This would be a good place for a warning message I think...

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
```diff
@@ -115,6 +115,36 @@ namespace Opm {
...
+           keyword has size different from nx*ny*nz it will also be
+           silently ignored.
+
+      With a mutable EclipseGrid instance you can later call the
+      EclipseGrid::resetACTNUM() method when you have complete actnum
+      information. The EclipseState based construction of EclipseGrid
+      is a two-pass operation, which guarantees that actnum is handled
+      correctly.
+    */
+
```
atgeirr: Great documentation! Is there a reason it is in the cpp and not the hpp file (I usually expect docs in the hpp)?

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
joakim-hove: I agree - coming.

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
joakim-hove: > Great documentation!

Thank you - who would have thought that!

> Is there a reason it is in the cpp and not the hpp file

Not a strong reason; I guess I like the documentation to be close to the peculiarities of the implementation - but of course I realize the hpp file would be a better place for *api* documentation.

joakim-hove:

joakim-hove:
This is quite trivial - merging.

PR #797, opened 2016-05-09
Summary config from eclipsestate

jokva:


opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -449,6 +449,17 @@ namespace Opm {
         }
     }
 
+    std::vector< int > Eclipse3DProperties::getRegions( const std::string& kw ) const {
```
joakim-hove: Can this be renamed to e.g. `getRegionValues()` - to more clearly differentiate from the full region keyword itself.

jokva: They're not *values* though, they're names/identifiers represented as an integer.

opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp
```diff
@@ -189,27 +186,40 @@ namespace Opm {
...
+    SummaryConfig::SummaryConfig( const Deck& deck, const EclipseState& es )
+        : SummaryConfig( deck,
+                         *es.getSchedule(),
+                         es.get3DProperties(),
+                         dimensions( *es.getInputGrid() ) )
+    {}
+
+    SummaryConfig::SummaryConfig( const Deck& deck,
+                                  const Schedule& schedule,
+                                  const Eclipse3DProperties& props,
```
joakim-hove: We should augment the `Eclipse3DProperties` class with the `(nx,ny,nz)` information, that way we can shorten this signature slightly and it makes a certain sense to have there as well.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
joakim-hove: > they're names/identifiers represented as an integer.

I agree - maybe `getRegionID` - or just leave it.

joakim-hove:
Not really part of this patch - but in the `SummaryConfig` constructor there are several constructions of this type:

```C++
const auto regions = item.size() > 0 && item.hasValue( 0 )
            ? item.getData< int >()
            : es.getRegions( "FIPNUM" );
```
These keywords do not have a default specified, so the tests `item.size() > 0` and `itemt.hasValue( 0 )` must be both true - or alternatively both false.

joakim-hove:
Extremely minor comments - fix if you agree; and I will merge.

jokva:
> These keywords do not have a default specified, so the tests item.size() > 0 and itemt.hasValue( 0 ) must be both true - or alternatively both false.

Oh, ok. So by checking item.size() we can determine if we must look in the deck or not? I'll put it on the refactor list, that file will get some love soon.

PR #800, opened 2016-05-10
Ignore RPTONLY

jokva:


joakim-hove:
Hehe; 1, 2 general?

jokva:
For now, yes. This is mostly an exercise in catchup, until we figure out where and how to handle keywords that control output behaviour that's not mapping to a value.

PR #801, opened 2016-05-10
Throw now contains name() of deckKeyword

pgdr:
Replaces error message `"Not a data keyword ?"` with `"Not a data keyword "KEYWORD"?"`

### DIFF
--- a/opm/parser/eclipse/Deck/DeckKeyword.cpp
+++ b/opm/parser/eclipse/Deck/DeckKeyword.cpp
@@ -97,7 +97,7 @@ namespace Opm {
         if (m_recordList.size() == 1)
             return getRecord(0);
         else
-            throw std::range_error("Not a data keyword ?");
+            throw std::range_error("Not a data keyword \"" + name() + "\"?");
     }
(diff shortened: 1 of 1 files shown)


andlaus:
IMO that's a good start. Question: Is canonical location tracking planned?

(e.g. for keywords like "WCONHIST", just knowing the name is probably not too useful.)

jokva:
It's not *planned* in the sense that we have a list of throws or error messages to update, but if we do stumble across them with more information floating about we'll certainly add when it's convenient.

andlaus:
> but if we do stumble across them with more information floating about we'll certainly add when it's convenient.

To have any decent chance of this 'being convenient', I suppose the first step needs to be to add `getFileName()` and `getLineNumber()` methods to `Deck(Keyword|Record|Item)`. For the keyword and probably also for the record class I don't see a problem because there are typically not too many of those objects, but doing it naively may be an issue for the DeckItem class.

That said, I personally need to create decks only sporadically, so I currently assign a very low priority for this feature.

PR #804, opened 2016-05-12
Support for eclipse -> SI unit conversions

jokva:


andlaus:
the patch itself is fine IMO, but the approach is unfortunate, I think:

- the conversion factors need to be accessed using magic numbers (e.g. `metric2si[3]` for length conversions).
- I don't really get why these arrays are needed at all: `Opm::(Metric|Field)::*` already are the conversion factors to SI units!
- the approach is not really general: Access to arbitrary combined units is already provided by `Opm::UnitSystem`

andlaus:
I'd also recommend to not use l33d-speak for variable names, i.e. rename `metric2si` to `metricToSi`

jokva:
I'll shed some lights on the whys:

In conversion between output (eclipse units, both for writing and reading) and simulator (SI) there are a lot conversions needed that could be both metric and SI. The arrays, used informally as `conversion_table` are used as vtables since output knows what unit to convert to, but doesn't want to recalculate what **system** to use. The tables themselves are really only there for `EclipseWriter`, `Summary` etc. to grab; all logic use `from_si` or `to_si`. The functions are identical since they use the same set of identifiers to determine offset, but later we can add debug features here, i.e. asserts that checks if the table goes the right way.

The conversion factors aren't as such accessed by magic numbers, but rather the `dim` enum. The typical use scenario would be `from_si( conversion_table, dim::pressure, val )`. I don't want to deal with the multiplications, divisions etc on a per-case basis when converting between output and internal representation, hence a more general approach.

andlaus:
> there are a lot conversions needed that could be both metric and SI.

That's true, but only for data items.  since the conversion factor needs to be calculated only once for the whole array of these items (e.g., in `getSiDoubleData()`) it should not at all performance critical because a typical deck contains only a few couple of thousands items (e.g. the Norne deck only contains about 250 non-summary, non-schedule keywords less than 1500 overall)...

> The tables themselves are really only there for EclipseWriter, Summary etc. to grab;

this sounds an awful lot like an implementation detail of the output writers, so why don't you simply put this into opm-output?

> The conversion factors aren't as such accessed by magic numbers, but rather the dim enum.

I'd still count these as "magical constants", mainly because the order of the numbers matters when creating the `foo2bar` arrays. Also, I'd my first association with the name `dim` is a 3D axis (i.e. either X, Y or Z: I think it is even widely used for this purpose in opm-core and opm-simulators).

jokva:
> That's true, but only for data items. since the conversion factor needs to be calculated only once for the whole array of these items (e.g., in getSiDoubleData()) it should not at all performance critical because a typical deck contains only a few couple of thousands items (e.g. the Norne deck only contains about 250 non-summary, non-schedule keywords less than 1500 overall)...

The conversion is done once, yes, and only on contact between text files on disk. Still has to be done, and it's configurable how, and there are plenty of mappings that needs converting. I wrote it because it was too tedious to use `(Metric|Field)::`, and frankly very noisy.

> this sounds an awful lot like an implementation detail of the output writers, so why don't you simply put this into opm-output?

I considered it, but parser needs it too, and parser already owns the definitions. The first iteration was implemented in output, but it was then moved.

> I'd still count these as "magical constants", mainly because the order of the numbers matters when creating the foo2bar arrays. Also, I'd my first association with the name dim is a 3D axis (i.e. either X, Y or Z: I think it is even widely used for this purpose in opm-core and opm-simulators).

Sure, but moving this logic into something else and using a switch is also a magic number in that sense, so it doesn't change anything. As long as this is statically defined and maintained by parser, and accessed via the `from`/`to` functions this isn't a problem.

If you can come up with a better name I'm happy to change it, I've never liked it.

andlaus:
> The conversion is done once, yes, and only on contact between text files on disk. Still has to be done, and it's configurable how, and there are plenty of mappings that needs converting. I wrote it because it was too tedious to use (Metric|Field)::, and frankly very noisy.

Honestly, i can't see why `bar = Opm::field2si(foo, Opm::dim::pressure)` is less noisy than `bar = foo*Opm::Field::Pressure`. Anyway if you're hell-bound on doing it in the way of this PR, I'm not the one who will be trying (hard) to prevent you.

> >  this sounds an awful lot like an implementation detail of the output writers, so why don't you simply put this into opm-output?
>
> I considered it, but parser needs it too, and parser already owns the definitions. The first iteration was implemented in output, but it was then moved.

as I outlined above, the use cases of opm-parser already are better covered by `Opm::UnitSystem` IMO...

> If you can come up with a better name I'm happy to change it, I've never liked it.

I think it would be the best to use the infrastructure which is already in place. It might not be perfect, but in this universe what's perfect?

jokva:
> Honestly, i can't see why `bar = Opm::field2si(foo, Opm::dim::pressure)` is less noisy than `bar = foo*Opm::Field::Pressure`. Anyway if you're hell-bound on doing it in the way of this PR, I'm not the one who will be trying (hard) to prevent you.

It's not. It **is** less noisy than `bar = foo * (es.getDeckUnitSystem().getType() == UNIT_SYSTEM_METRIC ? Opm::Metric::Pressure : Opm::Field::Pressure)`

andlaus:
I think it should be used like this:

```c++
double conversionFactor = es.getDeckUnitSystem().parse("Pressure");
// [...]
bar = foo * conversionFactor
```

that's not more noisy, isn't it? (and it also automatically will support atrocities like LAB and PVT-M units.)

andlaus:
small correction: he snippet should be

```c++
auto pDim = es.getDeckUnitSystem().parse("Pressure");
// [...]
bar = pDim->convertRawToSi(foo);
```

the point is the same, though...

joakim-hove:
Enough girls - merging!

PR #805, opened 2016-05-13
Add SPE9_CP_GROUP as test case for the parser

totto82:
Is this sufficient for adding SPE9_CP_GROUP as a test case for the parser?
How can I run the test?


joakim-hove:
Thank you!

> Is this sufficient for adding SPE9_CP_GROUP as a test case for the parser?

Yes - that is sufficient; the test is of course not very comprehensive but it guarantees that we can parse the input and create a `EclipseState` instance. I am very happy to have these tests for the decks in opm-data.

> How can I run the test?

If you have told cmake where the opm-data repository is located; `cmake .. -DOPM_DATA_ROOT=/path/to/opm-data` the test should be run as part of the normal parser tests. 

Observe that same functionality is available for opm-simulators - if `OPM_DATA_ROOT` is set in the configure step one of the SPE cases is actually run as one of the tests.

PR #806, opened 2016-05-13
Renamed getNNC() to getInputNNC() avoids confusion

pgdr:
* The NNC in the eclipse state is and will always be the nnc from deck
* updated tests

Depends on
* OPM/opm-simulators#680 which in turn depends on
* OPM/opm-output#27


joakim-hove:
jenkins build this with downstreams opm-output=27 opm-simulators=680 please

Issue #802, opened 2016-05-11
PORV=0 and ACTNUM


totto82:
I am running the BRUGGE case. In this case there are many cells with zero porosity, but ACTNUM is not used to make this cells inactive. Cells with PORV=0 should be treated automaticly as non-active cells by the simulator. I have a fix for this in opm-core in the grid-processing, still I add this as an issue to opm-parser as I think this behaviour belongs in the parser.

pgdr:
This might be handled by a one-liner in [EclipseState.cpp#L66](https://github.com/pgdr/opm-parser/blob/master/opm/parser/eclipse/EclipseState/EclipseState.cpp#L66).

pgdr:
Do you have test data?  Here's a naïve patch (prototype) that I can test if you have some data I can test on.

By the way, should the converse also be done; if `actnum` is 0, set `poro` and `ntg` to 0?

    diff --git a/opm/parser/eclipse/EclipseState/EclipseState.cpp b/opm/parser/eclipse/EclipseState/EclipseState.cpp
    index bd5071e..fa52bed 100644
    --- a/opm/parser/eclipse/EclipseState/EclipseState.cpp
    +++ b/opm/parser/eclipse/EclipseState/EclipseState.cpp
    @@ -80,8 +80,17 @@ namespace Opm {
    
             // after Eclipse3DProperties has been constructed, we have complete ACTNUM info
    -        if (m_eclipseProperties.hasDeckIntGridProperty("ACTNUM"))
    -            m_inputGrid->resetACTNUM(m_eclipseProperties.getIntGridProperty("ACTNUM").getData().data());
    +        if (m_eclipseProperties.hasDeckIntGridProperty("ACTNUM")) {
    +            std::vector<int> actnum  = m_eclipseProperties.getIntGridProperty("ACTNUM").getData();
    +            std::vector<double> poro = m_eclipseProperties.getDoubleGridProperty("PORO").getData();
    +            std::vector<double> ntg  = m_eclipseProperties.getDoubleGridProperty("NTG").getData();
    +            size_t m = std::min(ntg.size(), poro.size());
    +            for (size_t i = 0; i < m; i++) {
    +                if (ntg[i] == 0 || poro[i] == 0)
    +                    actnum[i] = 0;
    +            }
    +            m_inputGrid->resetACTNUM(actnum.data());
    +        }

OPMUSER:
The correct logic is that if the PV is zero the block is inactive

[...]

PR #808, opened 2016-05-18
Set ACTNUM = 0 for cells with PORV == 0

joakim-hove:
This should fix: https://github.com/OPM/opm-parser/issues/802 

Unfortunately the solution is semi-ugly special case hack; the ordering requirements required to get it to work as post processor just made my head spin.


opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -412,21 +412,45 @@ namespace Opm {
...
-         invoking the getKeyword() method of the m_intGridProperties /
-         m_doubleGridProperties() directly and not through these methods.
-
-      2. Observe that this will autocreate a property if it has not been
-         explicitly added.
-    */
+
     const GridProperty<int>& Eclipse3DProperties::getIntGridProperty( const std::string& keyword ) const {
-        return m_intGridProperties.getKeyword( keyword );
+        auto& gridProperty = const_cast< Eclipse3DProperties* >( this )->m_intGridProperties.getKeyword( keyword );
```
pgdr: Can you flip the order of testing ACTNUM?  I.e., start the function with

    if (keyword != "ACTNUM")
        return m_intGridProperties.getKeyword(keyword);

And if not, _then_ do `const_cast`?

jokva:
Can't we make an ACTNUM post processor that would do this? Looks to me like it "only" requires PORV to run, and PORO (I think anyway) already works like that. Am I missing something?

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
joakim-hove: That was a very good idea - implemented in fixup.

joakim-hove:
> Can't we make an ACTNUM post processor that would do this? 

That was my original plan - and it might still be possible; but it was more difficult than I thought. The problem is the ordering of int properties and double properties.

jokva:
Because ACTNUM is an int property? It's not immediately clear to me why this should be a problem (being post processors and all).

It would be a saner and cleaner solution for sure, so I think it's worth the effort unless there is a fire to put out.

joakim-hove:
> It would be a saner and cleaner solution for sure, so I think it's worth the effort unless there is a fire to put out.

I agree - postprocessor based solution pushed.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -90,15 +87,56 @@ namespace Opm {
...
+             3. The code below will fail hard if it is called with a
+                grid which does not have full cell information.
+        */
+
+        void ACTNUMPostProcessor( std::vector<int>&       values,
+                                  const EclipseGrid*      eclipseGrid,
+                                  GridProperties<double>* doubleGridProperties)
+        {
+            bool hasPORV = doubleGridProperties->hasKeyword( "PORV" ) || doubleGridProperties->hasKeyword( "PORO");
+            if (!hasPORV)
```
jokva: Nitpick, but why store this? Why not just check `if( !hasKeyword( "PORV") && !hasKeyword( "PORO" ) )` return;

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -90,15 +87,56 @@ namespace Opm {
...
+                return;
+
+            {
+                auto& porv = doubleGridProperties->getOrCreateProperty("PORV");
+                porv.runPostProcessor();
+
+                {
+                    const auto& porvData = porv.getData();
+                    std::vector<bool> porvZeroMask(porvData.size() , false);
```
jokva: Is porvZeroMask even used?

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -412,23 +464,15 @@ namespace Opm {
...
-         functions in EclipseState should therefore ask for properties by
-         invoking the getKeyword() method of the m_intGridProperties /
-         m_doubleGridProperties() directly and not through these methods.
-
-      2. Observe that this will autocreate a property if it has not been
-         explicitly added.
-    */
     const GridProperty<int>& Eclipse3DProperties::getIntGridProperty( const std::string& keyword ) const {
-        return m_intGridProperties.getKeyword( keyword );
+        auto& gridProperty = const_cast< Eclipse3DProperties* >( this )->m_intGridProperties.getKeyword( keyword );
```
jokva: We shouldn't need a cast here. Why is this being introduced?

jokva:
A much better solution indeed.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
joakim-hove: Ehhh - no; thank you!

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
joakim-hove: Well - we need the `const_cast< >` to get a mutable `GridProperty`object - which is required to run the post processor. Previously none of the int properties had a post processor.

joakim-hove:

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
joakim-hove: *I* think it increases readability - ymmv though.

jokva: Then please make it `const`. I'm pretty paranoid when it comes to even local mutable vars simply because then I must always consider a code path where it might change and I need to scan for potential mutation. I consider it good practice.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
jokva: Wouldn't it be better to tie the postprocessor calling to `GridProperties::getKeyword` instead? That way we guarantee that the post processor is run when needed, right before it's returned properly. Because this way is possibly broken.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
joakim-hove: > Then please make it const

Sure - will do.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
joakim-hove: > Wouldn't it be better to tie the postprocessor calling to GridProperties::getKeyword instead?

That is probably a good idea; we must just think twice so we guarantee that can not create a recursive death of postprocessors calling themselves; that was certainly something to be feared previously - but it might be trivially ok now.

jokva: Splendid. Seems like a good task for tomorrow.

joakim-hove:

joakim-hove:
OK - this is ready for review.

joakim-hove:

opm/parser/eclipse/EclipseState/Grid/GridProperties.cpp
```diff
@@ -0,0 +1,127 @@
+/*
+  Copyright 2014 Statoil ASA.
```
pgdr: Close enough?

opm/parser/eclipse/EclipseState/Grid/GridProperties.hpp
```diff
@@ -82,33 +95,12 @@ namespace Opm {
...
-            else
-                throw std::invalid_argument("Invalid index");
-        }
-
-        GridProperty<T>& getKeyword(size_t index)  {
-            if (index < size())
-                return *m_property_list[index];
-            else
-                throw std::invalid_argument( "Invalid index" );
+            GridProperty<T>& property = (*m_properties.at( keyword ));//const_cast<const GridProperty<T>&> (*m_properties.at( keyword ));
```
pgdr: can you clarify the comment?

opm/parser/eclipse/EclipseState/Grid/GridProperties.hpp
```diff
@@ -192,8 +184,139 @@ namespace Opm {
             return getKeyword(name);
         }
 
+        /**
+           The fine print of the manual says the ADD keyword should support
+           some state dependent semantics regarding endpoint scaling arrays
+           in the PROPS section. That is not supported.
+        */
+
+        void handleADDRecord( const DeckRecord& record, BoxManager& boxManager) {
```
pgdr: why are these not in source file?

opm/parser/eclipse/EclipseState/Grid/GridProperties.hpp
jokva: Looks like a ghetto const cast to obtain a mutable reference from a const context.

pgdr:
Ready to merge after comments have been addressed and after rebase.

jokva:
There's no need to rebase since there aren't any conflicts.

opm/parser/eclipse/EclipseState/Grid/GridProperties.hpp
joakim-hove: > can you clarify the comment?

Ehhh no - I will remove it!

joakim-hove:
Code moved from `GridProperties.hpp` to `GridProperties.cpp` with explicit instantiation for `int` and `double`.

opm/parser/eclipse/EclipseState/Grid/GridProperties.cpp
```diff
@@ -0,0 +1,422 @@
...
+                                      I2Item.get< int >(0) - 1,
+                                      J1Item.get< int >(0) - 1,
+                                      J2Item.get< int >(0) - 1,
+                                      K1Item.get< int >(0) - 1,
+                                      K2Item.get< int >(0) - 1);
+        } else if (setCount != 0) {
+            std::string msg = "BOX modifiers on keywords must be either "
+                "specified completely or not at all. Ignoring.";
+            throw std::invalid_argument( msg );
+            //getMessageContainer().error(deckKeyword.getFileName() + std::to_string(deckKeyword.getLineNumber()) + msg);
```
pgdr: Is this future code or past code?

opm/parser/eclipse/EclipseState/Grid/GridProperties.cpp
```diff
@@ -0,0 +1,422 @@
...
+        }
+    }
+
+
+    const MessageContainer& GridProperties<T>::getMessageContainer() const {
+        return m_messages;
+    }
+
+    MessageContainer& GridProperties<T>::getMessageContainer() {
```
pgdr: If I'm not mistaking, these two functions lack `template< typename T >` or?

opm/parser/eclipse/EclipseState/Grid/GridProperties.cpp
joakim-hove: This is past code; I considered it not worth it to get access to the filename and lineno information; but still worth keeping as a comment. I'll just remove it completely.

PR #811, opened 2016-05-19
VS2015 : Adjustments and fixes for Windows

magnesj:


joakim-hove:
Well - I take you now get opm-parser to build on Windows; very good! Will look at your patches tomorrow.

opm/parser/eclipse/EclipseState/Schedule/MSW/SegmentSet.cpp
```diff
@@ -21,6 +21,11 @@
 #include <cmath>
 #include <map>
 
+#ifdef _WIN32
+#define _USE_MATH_DEFINES 
+#include <math.h>
```
joakim-hove: What is this needed for?

opm/parser/eclipse/Generator/KeywordLoader.cpp
```diff
@@ -109,7 +109,12 @@ namespace Opm {
         std::shared_ptr<ParserKeyword> parserKeyword = std::make_shared<ParserKeyword>(*jsonConfig);
         {
             boost::filesystem::path abs_path = boost::filesystem::absolute( path );
-            addKeyword( parserKeyword , abs_path.string() );
```
joakim-hove: Could we use `abs_path.generic_string()` unconditionally?

opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp
```diff
@@ -227,7 +227,19 @@ namespace Opm {
...
+        for (auto& x : section)
+        {
+            auto keywords = handler(x);
+            for (auto& keyword : keywords)
+            {
+                this->keywords.push_back(keyword);
+            }
+        }
+#else
         this->keywords = fun::concat( fun::map( handler, section ) );
```
joakim-hove: Why does this fail? If the MSVC compiler does not handle something in the `fun` namespace the workaround should be here: https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/Utility/Functional.hpp

opm/parser/eclipse/EclipseState/Schedule/MSW/SegmentSet.cpp
magnesj: M_PI is used in the file, and is defined in math.h on Windows

https://github.com/magnesj/opm-parser/blob/72539e28e1c786ba3d1d6b5a3c7192b68a140afb/opm/parser/eclipse/EclipseState/Schedule/MSW/SegmentSet.cpp#L174

opm/parser/eclipse/Generator/KeywordLoader.cpp
magnesj: Probably, but I have not tested it on Linux

jokva: Then this can't be merged. We should strive to avoid these kinds of \#ifdefs, and rather abstract them properly (i.e. this should either use `generic_string` unconditionally or `addKeyword` should take the boost path and figure this out for itself.

It won't be merged as-is.

opm/parser/eclipse/EclipseState/Schedule/MSW/SegmentSet.cpp
jokva: I don't like this solution. We already include cmath (so math.h is pointless), and this is just the define.

It should either be moved to be a build parameter, `_USE_MATH_DEFINES` should be above `#include <cmath>` or we should define pi ourselves as a `constexpr`.

andlaus: the

```c++
#define _USE_MATH_DEFINES 
```

looks suspicious. (since that macro starts with an underscore, it is assumed as an internal one by the C standard, so it is not allowed to be defined by user code. some compilers may warn about this...)

jokva: It's a Windows thing. It uses that to enable/disable parts of the `math.h` file. Regardless, there are better ways of solving this.

opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp
andlaus: why not just always use the new windows code path? it is not rocket science and using `fun::` makes it briefer but IMO not easier to understand...

opm/parser/eclipse/EclipseState/Tables/TableColumn.cpp
```diff
@@ -22,6 +22,8 @@
 #include <opm/parser/eclipse/EclipseState/Tables/ColumnSchema.hpp>
 #include <opm/parser/eclipse/EclipseState/Tables/TableColumn.hpp>
 
+#include <ert/util/ssize_t.h>
```
andlaus: what is that needed for? (in case it is `size_t`, plain c++ provides that via `#include <cstddef>`)

opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp
jokva: It's not rocket science, but it's slower, noisy and annoyingly manual and duplicated logic.

`map` and `concat` are scheduled to be replaced with lazy implementations too, which should be even more effective. `map` is seeing some more use throughout parser (and output) and abstractions are nice.

It's intent is also clearer (in my opinion anyway), and `map` and `concat` are pretty well established names. Maybe `concat` could've been aliased or renamed to `flatten`, but that's a different discussion.

Regardless, I'm curious as to what the problem is.

opm/parser/eclipse/EclipseState/Schedule/MSW/SegmentSet.cpp
magnesj: From the suggested solutions, here is my prioritized list

1. define pi as constxpr
2. move _USE_MATH_DEFINES above #include <cmath>
3. build parameter

opm/parser/eclipse/EclipseState/Tables/TableColumn.cpp
magnesj: ssise_t is used at several locations in opm-parser, and is able to store -1 in addition to size_t values.

https://github.com/magnesj/opm-parser/blob/72539e28e1c786ba3d1d6b5a3c7192b68a140afb/opm/parser/eclipse/EclipseState/Tables/TableColumn.cpp#L272

joakim-hove:
OK;

some cleanups are needed here:

1. As a general we do *not* want `#ifdef xxxx` around in the code; if it is absolutely needed it should be at a lowest possible level, and not in the "application code". 
2. This: `Probably, but I have not tested it on Linux` just  does not cut it.
3. The compiler flags changes look kind of ugly; but I have no better alternative. Maybe @akva2 has some best practice suggestions.

opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp
atgeirr: I have to admit I find the functional notation here a little hard to read, but not because of fun::map but because of std::bind and placeholders -- could it be changed to use a lambda instead? I guess this boils down to taste and habit (we're not going to have Haskell elegance anyway).

jokva: It was originally a lambda, I've been going back and forth on it. The entire file will get some love soon anyway, so it can be expected to be beautified a tad.

magnesj: Here is a dump of the compiler error in VS2015 Update 2 trying to compile 
` this->keywords = fun::concat( fun::map( handler, section ) );`


`1>D:\gitroot\ResInsight\ThirdParty\custom-opm-parser\opm-parser\opm\parser\eclipse\EclipseState\SummaryConfig\SummaryConfig.cpp(226): fatal error C1001: An internal error has occurred in the compiler.
1>  (compiler file 'f:\dd\vctools\compiler\cxxfe\sl\p1\c\symbols.c', line 5785)
1>   To work around this problem, try simplifying or changing the program near the locations listed above.
1>  Please choose the Technical Support command on the Visual C++
`

atgeirr: That's kinda hard to argue around... Too bad!

opm/parser/eclipse/EclipseState/Schedule/MSW/SegmentSet.cpp
magnesj: Here is an example of moving from M_PI to boost constants

https://github.com/ddemidov/vexcl/commit/696d74a44d406e979ebbe37dc4a1bd924e0d8358

CMakeLists.txt
```diff
@@ -173,4 +178,7 @@ if(NOT BUILD_SHARED_LIBS)
   list(APPEND opm-parser_LIBRARIES cjson)
 endif()
 set(opm-parser_TARGET opmparser)
-opm_cmake_config(opm-parser)
+
+if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "Windows")
+  opm_cmake_config(opm-parser)
```
akva2: config mode is just as valid on windows.

opm/parser/eclipse/EclipseState/Tables/TableColumn.cpp
akva2: unless it sits in ert interfaces, please use the one from sys/types.h

opm/parser/eclipse/EclipseState/Schedule/MSW/SegmentSet.cpp
akva2: this is 100% kosher windows code. it's how you are supposed to do it.

akva2:
apart from the minor comments i think this looks fine (barring that #ifdef but i think that point has been raised sky high by now).

opm/parser/eclipse/EclipseState/Schedule/MSW/SegmentSet.cpp
jokva: Regardless, it's used in the .cpp file and should be used there instead.

opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp
andlaus: > It's not rocket science, but it's slower, noisy and annoyingly manual and duplicated logic.

but it is still better than having both codepaths and some preprocessor-foo? I personally prefer some minor duplication of trivial code if it makes the code simpler in the sense that it reduces the mental load of stuff which you need to know in order to understand it. (remember the recent 'leftpad' incident?) That's my personal opinion, though.

> map and concat are scheduled to be replaced with lazy implementations too, which should be even more effective. map is seeing some more use throughout parser (and output) and abstractions are nice.

most places where these are used are in no way performance critical, I suppose?

> It's intent is also clearer (in my opinion anyway), and map and concat are pretty well established names. Maybe concat could've been aliased or renamed to flatten, but that's a different discussion.

if we were talking lisp or scheme, I'd agree. But in C++, these are not standard constructs, and people are not used to the functional programming paradigm...

jokva: > but it is still better than having both codepaths and some preprocessor-foo? I personally prefer some minor duplication of trivial code if it makes the code simpler in the sense that it reduces the mental load of stuff which you need to know in order to understand it. (remember the recent 'leftpad' incident?) That's my personal opinion, though.

The preprocessor solution is unacceptable, yes. But the solution should then go into `fun` or similar considerations, not abandoning all kind of abstraction. Leftpad has nothing to do with this and abstractions **reduce** the mental load. If it didn't we'd all still be using vacuum tubes, just to go on a tangent.

> most places where these are used are in no way performance critical, I suppose?

It's not just for efficiency, it's for generality and applicability. Want to populate a vector? Fine. Want to populate a list? Fine, same code. Want to populate a `std::map`? No problem. Want to not populate anything, but rather evalute-and-reduce? Got your back.

> if we were talking lisp or scheme, I'd agree. But in C++, these are not standard constructs, and people are not used to the functional programming paradigm...

This is just not true.

atgeirr: >> if we were talking lisp or scheme, I'd agree. But in C++, these are not standard constructs, and people are not used to the functional programming paradigm...

> This is just not true.

This statement is quite bold... While awareness of functional programming is on the rise among C++ programmers (and in general) I think it is fair to say that most don't know it very well. At least not well enough to make the use of fun::map and fun::concat easier to understand than the "manual" alternative.

I this particular case I guess I'd (somewhat regretfully) use the more primitive code, perhaps adding a comment that it works around a compiler bug.

jokva: I don't think that's an argument for using poor abstractions. I think we should rather improve upon and document better abstractions so they **are** know well. `map` and `concat` aren't conceptually hard, and what's going on and intent is communicated in an instant instead of having to mentally step through the program (loop in this case). The loop may have components that are easier to immediately recognise, which could be considered a case of Stockholm Syndrome I might add, but only in the sense that it's easier to see what they **do**, not what we want to **happen**. And I sincerely believe we should emphasise the latter.

If `map` and `concat` aren't sufficiently documented to be understood then that's a documentation problem and we'll fix that.

This problem should be addressed as a problem with the broken MSVC compiler and handled in `fun`. This `#ifdef` solution isn't happening regardless.

andlaus: > Leftpad has nothing to do with this

really? I agree that the NPM packaging issues have nothing to do with this, but the fact that this community felt like it was a good idea to introduce abstraction for two lines of code **has** quite a bit of relevance in the current context IMO. 

> abstractions reduce the mental load

IMO they first increase the mental load because you have to get familiar with them. They are only beneficial when they are so common that you don't need to think about them anymore...



>>   most places where these are used are in no way performance critical, I suppose?
>
> It's not just for efficiency, it's for generality and applicability. Want to populate a vector? Fine. Want to populate a list? Fine, same code. Want to populate a std::map? No problem. Want to not populate anything, but rather evalute-and-reduce? Got your back.

yeah, and in all cases I need to look up in the implementation what it means _exactly_ to e.g. concatenate a map and a vector.

>> if we were talking lisp or scheme, I'd agree. But in C++, these are not standard constructs, and people are not used to the functional programming paradigm...
>
> This is just not true.

let's compromise by saying that in C++ it is in the eye of the beholder and that at least the "old c++ dogs" are not very familiar with it ;)

jokva: > really? I agree that the NPM packaging issues have nothing to do with this, but the fact that this community felt like it was a good idea to introduce abstraction for two lines of code has quite a bit of relevance in the current context IMO. 

The for loop is an abstraction over "two lines of code". Why doesn't that apply?

> IMO they first increase the mental load because you have to get familiar with them. They are only beneficial when they are so common that you don't need to think about them anymore...

And there's only one way of getting there.

> yeah, and in all cases I need to look up in the implementation what it means exactly to e.g. concatenate a map and a vector.
```
auto m = fun::map( f, seq );
return std::vector< int > { m.begin(), m.end() }
```

> let's compromise by saying that in C++ it is in the eye of the beholder and that at least the "old c++ dogs" are not very familiar with it ;)

Sure. But an old dog could learn new tricks.

andlaus: > The for loop is an abstraction over "two lines of code". Why doesn't that apply?

it does, that's my point: the "plain old stupid" way with the for-loop would not require you to know anything about `fun::` and `boost::bind` and it would not even be more lines  ;)

> Sure. But an old dog could learn new tricks.

true, but it's only necessary if the new trick is sufficiently better than the old way, and that's not the case here, IMO. (N.B: over-abstraction is a common problem I see through the OPM codebase: abstraction can be a good thing, but it is a double-edged sword and IMO a cure worse than the disease if there are only a very limited number of users for an API.)

jokva: > it does, that's my point: the "plain old stupid" way with the for-loop would not require you to know anything about fun:: and boost::bind and it would not even be more lines ;)

It would most certainly be more lines, I don't know where you're getting that from. And it's not like you don't need to know the interfaces you use anyway, so that's not an argument.

> true, but it's only necessary if the new trick is sufficiently better than the old way, and that's not the case here, IMO. 

Your loss. You don't **have** to use it, but transforming-and-returning is a **common** operation, it will be used more in the future and a lot of operations today are essentially transform-and-return but are so convoluted in their manual approach that they don't appear that way anymore, and that removes the opportunity to generalise and simplify them. And fewer lines are typically easier to maintain than more lines.

Arguing that `map` is conceptually more difficult than declare container; iterate; append; is a *tough* sell, but feel free to give it a go.

andlaus: > It would most certainly be more lines, I don't know where you're getting that from. And it's not like you don't need to know the interfaces you use anyway, so that's not an argument.

```c++
for (auto& x : section)   {
    for (auto& keyword : handleKW(x, schedule, props, n_xyz))
       this->keywords.push_back(keyword);
}
```

vs 

```c++
        using namespace std::placeholders;
         const auto handler = std::bind( handleKW, _1, schedule, props, n_xyz );
         this->keywords = fun::concat( fun::map( handler, section ) );
```

I'd call this a draw. on the other hand the second version requires more implicit knowledge.

Let's stop the discussion here, I'm getting tired of it: If you want to use functional style programming, that's fine with me, even though for maintenance reasons I advice to avoid it if it does not come with clear benefits other than "it is more elegant".

jokva: Curious as to how "more elegant" isn't a clear benefit.

`auto f = [&]( const DeckKeyword& kw ) { return handleKW( kw, schedule, props, n_xyz ); }` for reference.

atgeirr: Much as I hate to short-circuit a fine discussion is the fact that it is not obvious why the code here does push keywords around at all, and the solution is an explanatory comment, irrespective of beautiful or primitive code...

joakim-hove: > it is not obvious why the code hete does ...

That is the beauty of functional programming: write incomprehensible code very compactly.

magnesj:
Thanks for feedback, closing this PR for now, and will create a new PR with updates ASAP.

PR #812, opened 2016-05-20
Introducing EclipseConfig -- a container class

pgdr:
This PR cleans up some dependencies and intricacies in the `EclipseState` constructor and how the members depend on each other.  We do this by introducing a container class `EclipseConfig`.

* The API exposed by Parser (`EclipseState`) is unchanged as of now
* This class contains the following State members with methods
  - `m_ioConfig` and `getIOConfig` as well as `getIOConfigConst`
  - `m_initConfig` and `getInitConfig`
  - `m_simulationConfig` and `getSimulationConfig`
  - `m_summaryConfig` and `getSummaryConfig`
* Got rid of awkward `initIOConfigPostSchedule` function (and corresponding call)
* For better decoupling
* For future, ease the shared_ptr-ref transition

### DIFF
--- a/opm/parser/eclipse/CMakeLists.txt
+++ b/opm/parser/eclipse/CMakeLists.txt
@@ -70,6 +70,7 @@ ${generator_source}
 
 set (state_source
 EclipseState/EclipseState.cpp
+EclipseState/EclipseConfig.cpp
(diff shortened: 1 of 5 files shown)


jokva:
Thumbs up, looks good. I'll merge today unless someone protests.

PR #814, opened 2016-05-23
Unit system improvements

jokva:
Removes the conversions namespace from ConversionFactors.hpp and implements to/from_si, a less clunky conversion interface, to UnitSystem.

Personally I'd go further with the alternative to 'getDeckUnitSystem' and just call it `units` (or if maybe `unit_conversion`), and this is obviously open for discussion.


joakim-hove:
> Personally I'd go further with the alternative to 'getDeckUnitSystem' and just call it units (or if maybe unit_conversion), and this is obviously open for discussion.

Well - the point is that these units are not any set of units, but precisely the units found in the deck - which will *always* be different from the SI units used internally. But I agree that the `getDeckUnitSystem` is a mouthfull.

opm/parser/eclipse/EclipseState/EclipseState.hpp
```diff
@@ -104,6 +104,7 @@ namespace Opm {
         // units because internally to opm-parser everything is represented by SI
         // units...
         const UnitSystem& getDeckUnitSystem() const;
+        const UnitSystem& getUnits() const;
```
joakim-hove: I do not think we should have two functions doing the same; what about `getDeckUnits()` or `getInputUnits()`?

jokva:
I feel that's already given away by the fact you're querying `EclipseState` about it, as well as the presence of to/from_si. YMMV I suppose.

And of course, what the set of units are shouldn't be of that much interest to clients, the important aspect is the conversion.

PR #816, opened 2016-05-25
VS2015 compiler fixes

magnesj:
`ssize_t` is included using `<ert/util/ssize_t.h>` On Linux, this will end up with include of `<sys/types.h>`

This PR will fix most of compiler issues on VS2015. The compiler issue related to
```C++ 
this->keywords = fun::concat( fun::map( handler, section ) );
```
will be handled in a separate PR.

See link below for closed (and not merged) PR and commit related to fun::concat issue
https://github.com/OPM/opm-parser/pull/811/commits/d6bd4485b0daef9d397e94f652c157db505c0778

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,13 +74,16 @@ if (USE_RUNPATH)
    endif()
 endif (USE_RUNPATH)
 
-
-if (NOT DEFINED CMAKE_CXX_FLAGS)
-  SET( CMAKE_CXX_FLAGS "-pipe -std=c++0x -pedantic -Wall -Wextra -Wformat-nonliteral -Wcast-align -Wpointer-arith -Wmissing-declarations -Wcast-qual -Wshadow -Wwrite-strings -Wchar-subscripts -Wredundant-decls")
-endif()
-
-SET( CMAKE_CXX_FLAGS_DEBUG   "${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_DEBUG}   -O0 -DDEBUG  -ggdb3")
-SET( CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_RELEASE} -O3 -DNDEBUG -mtune=native")
+if (MSVC)
+  add_definitions( "/W3 /D_CRT_SECURE_NO_WARNINGS /wd4996 /wd4244 /wd4267")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
+else()
+  if (NOT DEFINED CMAKE_CXX_FLAGS)
+    SET( CMAKE_CXX_FLAGS "-pipe -std=c++0x -pedantic -Wall -Wextra -Wformat-nonliteral -Wcast-align -Wpointer-arith -Wmissing-declarations -Wcast-qual -Wshadow -Wwrite-strings -Wchar-subscripts -Wredundant-decls")
+  endif()
+  SET( CMAKE_CXX_FLAGS_DEBUG   "${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_DEBUG}   -O0 -DDEBUG  -ggdb3")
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -74,13 +74,16 @@ if (USE_RUNPATH)
...
-
-if (NOT DEFINED CMAKE_CXX_FLAGS)
-  SET( CMAKE_CXX_FLAGS "-pipe -std=c++0x -pedantic -Wall -Wextra -Wformat-nonliteral -Wcast-align -Wpointer-arith -Wmissing-declarations -Wcast-qual -Wshadow -Wwrite-strings -Wchar-subscripts -Wredundant-decls")
-endif()
-
-SET( CMAKE_CXX_FLAGS_DEBUG   "${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_DEBUG}   -O0 -DDEBUG  -ggdb3")
-SET( CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_RELEASE} -O3 -DNDEBUG -mtune=native")
+if (MSVC)
+  add_definitions( "/W3 /D_CRT_SECURE_NO_WARNINGS /wd4996 /wd4244 /wd4267")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
```
joakim-hove: My CMake skills are quite rudimentary; but as I understand this code will not set `_DEBUG` and `_RELEASE` flags. Wouldn't this be more in line with the gnu solution:
```
if (MSVC)
    add_definitions( "/W3 /D_CRT_SECURE_NO_WARNINGS /wd4996 /wd4244 /wd4267")
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")

    set(CMAKE_CXX_FLAGS_DEBUG ${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_DEBUG} -extra -windows -debug -flags)
    set(CMAKE_CXX_FLAGS_RELEASE ${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_RELEASE} -extra -windows -release -flags) 
endif()
```

opm/parser/eclipse/EclipseState/Schedule/MSW/SegmentSet.cpp
```diff
@@ -21,6 +21,11 @@
 #include <cmath>
 #include <map>
 
+#ifdef _WIN32
```
joakim-hove: OK - I have looked a bit more on this; and it seems it is the Gnu compiler which is non standard here. My suggestion is:
1. Remove the whole `#ifdef` block.
2. Introduce `static const double PI = 3.1415926535897932384626;`
3. Replace the `M_PI` with `PI`.

joakim-hove:
OK - I will merge when the PI issue has been resolved; the CMake issue was a suggestion - your call.

opm/parser/eclipse/EclipseState/Schedule/MSW/SegmentSet.cpp
andlaus: *please* no OPM-specfic hacks, there are enough of these already. (e.g. @joakim-hove's proposal may lead to naming conflicts if no extra precautions are taken.) better use boost here:

https://stackoverflow.com/questions/1727881/how-to-use-the-pi-constant-in-c

and

http://www.boost.org/doc/libs/1_61_0/libs/math/doc/html/math_toolkit/constants.html

joakim-hove: No boost; this is not enough of a reason to increase boost usage. I thought the `static` should be sufficient to protect against possible naming conflicts, an alternative would be to put it in an anonymous namespace.

akva2: this is how you are supposed to do it with msvc. this is a separate compilation unit, the define (and the *PREPROCESSOR* definitions it enables) is only around while compiling this .cpp file so i don't see what the fuss is about.

joakim-hove: > so i don't see what the fuss is about.

OK - fair enough. I fold.

PR #817, opened 2016-05-25
EclipseGrid copy constructor.

joakim-hove:
Removed EclipseGrid::EclipseGrid( ecl_grid_type * )

Tiny steps towards fully functional C++ value semantics.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
+++ b/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
@@ -257,6 +257,23 @@ namespace Opm {
     }
 
 
+    size_t EclipseGrid::activeIndex(size_t i, size_t j, size_t k) const {
+        return activeIndex( getGlobalIndex( i,j,k ));
+    }
+
+
+
+    size_t EclipseGrid::activeIndex(size_t globalIndex) const {
+        assertCellInfo();
+        {
+            int active_index = ecl_grid_get_active_index1( m_grid.get() , globalIndex );
+            if (active_index < 0)
+                throw std::invalid_argument("Input argument does not correspond to an active cell");
+            return static_cast<size_t>( active_index );
+        }
+    }
+
+
 
     size_t EclipseGrid::getNX( ) const {
         return m_nx;
--- a/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
+++ b/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
@@ -65,18 +65,21 @@ namespace Opm {
         m_nz = static_cast<size_t>( ecl_grid_get_nz( c_ptr() ));
     }
 
-    EclipseGrid::EclipseGrid(const ecl_grid_type * src_ptr)
-        : m_minpvValue(0),
-          m_minpvMode(MinpvMode::ModeEnum::Inactive),
-          m_pinch("PINCH"),
-          m_pinchoutMode(PinchMode::ModeEnum::TOPBOT),
-          m_multzMode(PinchMode::ModeEnum::TOP)
+    /* Copy constructor */
+    EclipseGrid::EclipseGrid(const EclipseGrid& src)
+        : m_minpvValue( src.m_minpvValue ),
+          m_minpvMode( src.m_minpvMode ),
+          m_pinch( src.m_pinch ),
+          m_pinchoutMode( src.m_pinchoutMode ),
+          m_multzMode( src.m_multzMode ),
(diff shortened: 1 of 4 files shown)


opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
```diff
@@ -65,18 +65,20 @@ namespace Opm {
         m_nz = static_cast<size_t>( ecl_grid_get_nz( c_ptr() ));
     }
 
-    EclipseGrid::EclipseGrid(const ecl_grid_type * src_ptr)
+    /* Copy constructor */
+    EclipseGrid::EclipseGrid(const EclipseGrid& src)
         : m_minpvValue(0),
```
atgeirr: Should not all these member values be copied from `src`?

joakim-hove:

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
joakim-hove: Aiaiai - well spotted; thank you.

joakim-hove: > Should not all these member values be copied from src?

Is there any best practice in a situation like this; should I access the members of src directly - or through their accessors?

jokva: src directly.

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
```diff
@@ -65,18 +65,20 @@ namespace Opm {
...
-          m_pinchoutMode(PinchMode::ModeEnum::TOPBOT),
-          m_multzMode(PinchMode::ModeEnum::TOP)
+    /* Copy constructor */
+    EclipseGrid::EclipseGrid(const EclipseGrid& src)
+        : m_minpvValue( src.m_minpvValue ),
+          m_minpvMode( src.minPvMode ),
+          m_pinch( src.m_pinch ),
+          m_pinchoutMode( src.pincoutMode ),
+          m_multzMode( src.multMode ),
+          m_nx( m_nx ),
```
atgeirr: Missing `src.` here.

joakim-hove:

joakim-hove:
Any objections to merging this tomorrow?

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
```diff
@@ -65,18 +65,20 @@ namespace Opm {
...
+          m_pinch( src.m_pinch ),
+          m_pinchoutMode( src.m_pinchoutMode ),
+          m_multzMode( src.m_multzMode ),
+          m_nx( src.m_nx ),
+          m_ny( src.m_ny ),
+          m_nz( src.m_nz )
     {
-        m_grid.reset( ecl_grid_alloc_copy( src_ptr ) );
+        if (src.hasCellInfo())
+            m_grid.reset( ecl_grid_alloc_copy( src.c_ptr() ) );
```
pgdr: We don't want to copy `m_messages`?

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
```diff
@@ -254,6 +256,23 @@ namespace Opm {
...
 
 
+    size_t EclipseGrid::activeIndex(size_t i, size_t j, size_t k) const {
+        return activeIndex( getGlobalIndex( i,j,k ));
+    }
+
+
+
+    size_t EclipseGrid::activeIndex(size_t globalIndex) const {
+        assertCellInfo();
```
pgdr: Can we change this with something like

    if (!hasCellInfo())
        return globalIndex;

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
jokva: We do, otherwise it's not a copy.

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
joakim-hove: > Can we change this with something like

I would say no - because if you do not have cellinfo you basically have no clue; lack of cellinfo does *not* imply that all cells are active, which should be the case for `return globalIndex;` to be correct. 

When we have implemented a `GridDim` class for the lean grids we can stop this checking.

PR #818, opened 2016-05-25
VS2015 : Workaround for VS2015 compiler bug

magnesj:
I have tested several proposed solutions based on input from @jokva, but unfortunately I was not able to fix the compiler issue in any other way than suggested here. As I have mentioned earlier, I do not have much experience with newer c++, so it is likely there exists an easy solution I am not aware of.

To fix this issue in a different way than suggested here, I believe a developer with experience in this field working on VS2015 will be required.

### DIFF
--- a/opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp
+++ b/opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp
@@ -252,7 +252,18 @@ namespace Opm {
 
         using namespace std::placeholders;
         const auto handler = std::bind( handleKW, _1, schedule, props, n_xyz );
-        this->keywords = fun::concat( fun::map( handler, section ) );
+
+        /* This line of code does not compile on VS2015
+         *   this->keywords = fun::concat( fun::map( handler, section ) );
+         * The following code is a workaround for this compiler bug */
+        for (auto& x : section)
+        {
+            auto keywords = handler(x);
+            for (auto& keyword : keywords)
+            {
+                this->keywords.push_back(keyword);
+            }
+        }
     }
 
     SummaryConfig::const_iterator SummaryConfig::begin() const {


joakim-hove:
OK - thank you for the effort. As indicated from private communication this will not be *immediately* required, so I assume we have some days to look into this?

I am happy with fix itself, the problem is that the underlying VS problem introduces an uncertainty in all of the `fun::` code - what can safely be used there, and which combinations should be avoided.

joakim-hove:
OK - with https://github.com/OPM/opm-parser/pull/820 we have added two small `fun::` tests, can you please report back how those tests fare.

magnesj:
For some reason (probably related to compiler flags) - the line
`this->keywords = fun::concat( fun::map( handler, section ) );`
does not give compiler crash anymore on VS2015 in Debug using the head master branch on opm-parser (359293bcd91d0e9d46fd6da4f4cf9158f7bb7441). Release still causes compiler internal error as before.

The code does not compile in Debug, and I suspect the compiler issue is caused by the combination of `std::bind `and `fun::map` 

Here is the compiler Debug output :
`1>  SummaryConfig.cpp
1>D:\test-opm-parser\gitroot\opm-parser\opm\parser\eclipse\EclipseState\SummaryConfig\SummaryConfig.cpp(255): error C2672: 'Opm::fun::map': no matching overloaded function found
1>D:\test-opm-parser\gitroot\opm-parser\opm\parser\eclipse\EclipseState\SummaryConfig\SummaryConfig.cpp(255): error C2893: Failed to specialize function template 'std::vector<std::result_of<F(C::const_iterator::value_type &)>::type,std::allocator<_Ty>> Opm::fun::map(F,const C &)'
1>          with
1>          [
1>              _Ty=std::result_of<F(C::const_iterator::value_type &)>::type
1>          ]
1>  D:\test-opm-parser\gitroot\opm-parser\opm\parser\eclipse\EclipseState\SummaryConfig\SummaryConfig.cpp(255): note: With the following template arguments:
1>  D:\test-opm-parser\gitroot\opm-parser\opm\parser\eclipse\EclipseState\SummaryConfig\SummaryConfig.cpp(255): note: 'F=std::_Binder<std::_Unforced,std::vector<ERT::smspec_node,std::allocator<_Ty>> (__cdecl &)(const Opm::DeckKeyword &,const Opm::Schedule &,const Opm::Eclipse3DProperties &,std::array<int,3>),const std::_Ph<1> &,const Opm::Schedule &,const Opm::Eclipse3DProperties &,std::array<int,3> &>'
1>  D:\test-opm-parser\gitroot\opm-parser\opm\parser\eclipse\EclipseState\SummaryConfig\SummaryConfig.cpp(255): note: 'C=Opm::SUMMARYSection'
1>D:\test-opm-parser\gitroot\opm-parser\opm\parser\eclipse\EclipseState\SummaryConfig\SummaryConfig.cpp(255): error C2672: 'Opm::fun::concat': no matching overloaded function found
========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========
`

jokva:
Change bind to a lambda and see what happens.

magnesj:
I have never written a lambda, how do you formulate a lambda out to the bind?

jokva:
`auto handler = [&]( const DeckKeyword& kw ) { return handleKW( kw, section, props, n_xyz ); };`

akva2:
```
auto f =[](T x){ return g(x,b); };
```
where g is the two param function, b is the bound value and T the type of x.


-------- Opprinnelig melding --------
Fra: Magne Sjaastad <notifications@github.com>
Dato: 01.06.2016 15:20 (GMT+01:00)
Til: OPM/opm-parser <opm-parser@noreply.github.com>
Emne: Re: [OPM/opm-parser] VS2015 : Workaround for VS2015 compiler bug (#818)


I have never written a lambda, how do you formulate a lambda out to the bind?

magnesj:
`auto handler = [&](const DeckKeyword& kw)         { 
            return handleKW(kw, schedule, props, n_xyz);        };
         this->keywords = fun::concat( fun::map( handler, section ) );`

gives the original error 

`1>D:\gitroot-magnesj\opm-parser\opm\parser\eclipse\EclipseState\SummaryConfig\SummaryConfig.cpp(263): fatal error C1001: An internal error has occurred in the compiler.
1>  (compiler file 'f:\dd\vctools\compiler\cxxfe\sl\p1\c\symbols.c', line 5785)
1>   To work around this problem, try simplifying or changing the program near the locations listed above.
1>  Please choose the Technical Support command on the Visual C++
`

jokva:
So I looked a bit about and this error seems to occur when the code generator is confused by the optimiser.

Considering the code works in debug mode it's time to figure out exactly what optimisation option it is that causes problems.

Can you try removing all opt flags and re-add them one by one?

magnesj:
Thanks for input. I have not time to investigate this issue further for now, and as suggested in my first comment, I believe a developer with experience with lambdas working on VS2015 will be required. As I already have indicated, my experience with lambdas is very limited.

jokva:
I don't think it has all that much to do with lambdas, but rather MSVC's optimiser that gives up. Have you tried removing the options to the optimiser?

magnesj:
@jokva I have not tested on what compiler option that is related to this issue. You can see the output including std errors mentioned earlier in this thread https://github.com/OPM/opm-parser/pull/818#issuecomment-222960118

alfbr:
Maybe this is the right time to merge the PR?

joakim-hove:
> Maybe this is the right time to merge the PR?

My understanding of the problem is that it is a problem with optimization and the `fun::concat( fun::map( ) )` combination in the `Summary.cpp` file, other calls to `fun:map()` and `fun::concat()` seem unaffected. So I will merge this now - with the following comments:

1. It is still considered 100% legitimate to use the `fun::map()` and `fun::concat()` functions in the parser code; since there already are other call sites which work nicely I exepct (fingers crossed ...) that this will not create problems.

2. Subsequent patches "fixing" the problem, e.g. by compiling the `Summary.cpp` file with special flags on Windows, will be merged.

PR #822, opened 2016-05-29
Avoiding use fun::concat() in SummaryConfig

joakim-hove:
See: https://github.com/OPM/opm-parser/pull/821


magnesj:
Test of this PR does not change the compilation issues for VS2015, an indication of that it is `fun::map ` which is causing trouble.

See comment https://github.com/OPM/opm-parser/pull/818#issuecomment-222960118

PR #823, opened 2016-05-30
Reduce EclipseState interdependencies - IOConfig and Schedule

jokva:
A refactoring patch set that reduces some of the interdependencies between the EclipseState members.

Synopsis:
* Schedule and IOConfig dependencies mostly broken (the IOConfig SOLUTION-handling bit still lingers, as well as reading well RLT/PLT from Schedule)
* A couple of new constructor overloads.
* Overhauled test suite.


opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
```diff
@@ -39,29 +40,241 @@
...
+        const auto& items = kw.getStringData();
+
+        /* if any of the values are pure integers we assume this is meant to be
+         * the slash-terminated list of integers way of configuring. If
+         * integers and non-integers are mixed, this is an error.
+         */
+        const auto ints = std::any_of( items.begin(), items.end(), is_int );
+        const auto strs = !std::all_of( items.begin(), items.end(), is_int );
+
+        if( ints && strs ) throw std::runtime_error(
```
joakim-hove: Splitting this over three lines just looks funny - IMO.

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
```diff
@@ -39,29 +40,241 @@
...
+
+        /* If no BASIC mnemonic is found, either it is not present or we might
+         * have an old data set containing integer controls instead of mnemonics.
+         * BASIC integer switch is integer control nr 1, FREQUENCY is integer
+         * control nr 6.
+         */
+
+        /* mnemonics, but without basic and freq. Effectively ignored */
+        if( !ints ) return {};
+
```
joakim-hove: Can I have named constants for the magic numbers 0 and 5?

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
```diff
@@ -39,29 +40,241 @@
...
+        return restartConfig( step, basic, freq );
+    }
+
+    IOConfig::restartConfig IOConfig::rptsched( const DeckKeyword& keyword ) {
+        size_t restart = 0;
+        bool restart_found = false;
+
+        const auto& items = keyword.getStringData();
+        const auto ints = std::any_of( items.begin(), items.end(), is_int );
```
joakim-hove: This little test is duplicated in two neighbouring functions.

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
```diff
@@ -39,29 +40,241 @@
...
+        /* No RESTART or NOTHING found, but it is not an integer list */
+        if( strs ) return {};
+
+        /* We might have an old data set containing integer controls instead of
+         * mnemonics. Restart integer switch is integer control nr 7
+         */
+
+        if( items.size() < 7 ) return {};
+
+        return restartConfig( std::stoul( items[ 6 ] ) );
```
joakim-hove: Magic number.

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -115,19 +124,7 @@ namespace Opm {
...
-        if (deck.hasKeyword("START")) {
-             const auto& startKeyword = deck.getKeyword("START");
-            startTime = TimeMap::timeFromEclipse(startKeyword.getRecord(0));
-        }
-
-        m_timeMap.reset(new TimeMap(startTime));
-    }
-
-    void Schedule::iterateScheduleSection(const ParseContext& parseContext , const SCHEDULESection& section, IOConfigPtr ioConfig) {
+    void Schedule::iterateScheduleSection(const ParseContext& parseContext , const SCHEDULESection& section, IOConfigPtr ) {
```
pgdr: Why is the `IOConfigPtr` here?  To keep other things outside from breaking?  If so, how much work is it to remove?

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
```diff
@@ -39,29 +40,241 @@
...
+
+        size_t current_step = 1;
+        bool ignore_RPTSCHED_restart = false;
+        restartConfig unset;
+
+        DynamicState< IOConfig::restartConfig >
+            restart_config( timemap, restartConfig( 0, 0, 1 ) );
+
+        for( const auto& keyword : schedule ) {
+            const auto& name = keyword.name();
```
joakim-hove: The body in this loop is not very easy to read; I would have preferred if it was reorganized slightly so that two phases; i.e. check if we should update - and then the actual update were more clearly separated - know it is a bit of a blur.
```
    if (cond1)  continue;
    if (cond2)  continue;
    if (cond3)  continue;

    // OK - now we know that we should update the restart configuration; create the rs object and get going.
  
    rs = ....
```

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
```diff
@@ -218,7 +370,7 @@ namespace Opm {
     void IOConfig::initFirstOutput(const Schedule& schedule) {
         m_first_restart_step = -1;
         m_first_rft_step = -1;
-        assertTimeMap( schedule.getTimeMap() );
+        assertTimeMap( this->m_timemap );
```
joakim-hove: Is it really necessary to internalize a `TimeMap` in the `IOConfig` object - I would think it was only used in the restart configuration?

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
```diff
@@ -293,53 +418,14 @@ namespace Opm {
         } //RPTSOL
     }
 
+    void IOConfig::overrideRestartWriteInterval(size_t interval) {
+        size_t step = 0;
+        size_t basic = interval > 0 ? 3 : 0;
```
joakim-hove: Some English for this interplay: `size_t basic = interval > 0 ? 3 : 0;`?

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
pgdr: Please remove `IOConfigPtr` here, and then remove `ioConfig` variable in constructor parameter list.

joakim-hove:

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
jokva: It's really not anymore, but the `DynamicState` will still keep it about. I'll  look into making an extra commit getting rid of it.

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
jokva: Affirmative.

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
jokva: It is, but the test itself won't get any nicer and they both have it as a precondition. It's not the worst thing in the world to duplicate, but I can look into lifting it out.

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
jokva: Yes.

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
pgdr: Possibly add a function `assertUniformType` or something?

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.hpp
```diff
@@ -203,23 +201,44 @@ namespace Opm {
...
+            {}
+
+            restartConfig( size_t step, size_t b, size_t freq ) :
+                timestep( step ),
+                basic( b ),
+                frequency( freq )
+            {}
+
+            bool operator!=(const restartConfig& rhs) const {
+                return !( *this == rhs );
```
pgdr: Good! :)

pgdr:
It's a bit hard to tell from the commits/changelog, but are all the test cases more or less preserved?

That is, is the test coverage in this PR a superset of the test coverage prior to this?

pgdr:
This is ready to be merged after questions and comments have been addressed.

jokva:
Arright, it'll be done & flagged Monday.

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
jokva: Possible, which I probably would have done if I just used the derived boolean. I don't though, I also use the lists (if this test doesn't succeed).

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
jokva: It's largely how it flows now. Unfortunately, the calculate-restart-config functions rely on knowing what timestep we're at, which is what the first two branches manage. After that it's simply exit-early conditions.

Problem is that computing this without the mutable state is quite a bit more expensive, or it's pretty inconvenient in C++ (we're lacking the good abstractions for it really). Considering how fairly straight forward the algorithm now is I don't think it's worth trying to make it flow differently, but it's not possible to make a set of `if( cond ) continue;` at the top of the loop body because there are too many stateful dependencies later on.

Imperative code everybody.

jokva:
The points brought up have been addressed.

pgdr:
I'm merging this when all the tests have gone through on my system.

PR #824, opened 2016-06-03
Travis build downstream

joakim-hove:
Will build downstream modules on travis.


joakim-hove:
jenkins build this with opm-common=128 please

PR #827, opened 2016-06-05
Check for symlink() and compile test conditionally

joakim-hove:
We have some tests checking symlink behavior - disable these tests if we don't have `symlink()`


joakim-hove:
Merging tomorrow!

joakim-hove:
> Merging tomorrow!

Now!

PR #830, opened 2016-06-07
Add Prtinfo message type allows messages go into print file.

qilicun:
See https://github.com/OPM/opm-common/pull/135

### DIFF
--- a/opm/parser/eclipse/Parser/MessageContainer.cpp
+++ b/opm/parser/eclipse/Parser/MessageContainer.cpp
@@ -120,17 +120,17 @@ namespace Opm {
     }
 
 
-    void MessageContainer::prtinfo( const std::string& msg,
+    void MessageContainer::note( const std::string& msg,
                                     const std::string& filename,
                                     const size_t lineno ) {
         m_messages.emplace_back(
-            Message { Message::Prtinfo, msg, { filename, lineno } }
+            Message { Message::Note, msg, { filename, lineno } }
         );
     }
 
 
-    void MessageContainer::prtinfo( const std::string& msg ) {
-        m_messages.emplace_back( Message { Message::Prtinfo, msg, {} } );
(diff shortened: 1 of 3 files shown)


atgeirr:
Please see my comment to OPM/opm-common#135, basically is it feasible to get tagging in MessageContainer to avoid this workaround?

joakim-hove:
> Please see my comment to OPM/opm-common#135, basically is it feasible to get tagging in MessageContainer to avoid this workaround?

Yes - that would be my [suggestion] (https://github.com/OPM/opm-common/pull/135#issuecomment-224412548).

atgeirr:
> Yes - that would be my suggestion.

That would let us avoid this workaround, which would be better. Do you agree, @qilicun?

qilicun:
@atgeirr 
It seems that we agree on this https://github.com/OPM/opm-common/pull/135#issuecomment-224412548. I still think that introducing a new message type could be a good solution for **those messages** that should only go into print file. I will update the two PR according to the comments.

atgeirr:
OK, let's first get the extra message type. I think it should be named Notification or something other than PrtInfo though. Also: its ordering in the enum definitions (here and in opm-common) should reflect the expectation that this is "more chatty than Info", so it should go in between Debug and Info.

opm/parser/eclipse/Parser/MessageContainer.cpp
```diff
@@ -119,6 +119,21 @@ namespace Opm {
         m_messages.emplace_back( Message { Message::Problem, msg, {} } );
     }
 
+
+    void MessageContainer::note( const std::string& msg,
+                                    const std::string& filename,
```
atgeirr: Inconsistent indentation.

atgeirr:
Looks fine, minor nitpick noted above. Otherwise ready to merge.

PR #832, opened 2016-06-08
Log well messages

qilicun:
- Make a much more clear messages for VFP
- Log the history related message to print file.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Well.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Well.cpp
@@ -156,7 +156,7 @@ namespace Opm {
 
     bool Well::setStatus(size_t timeStep, WellCommon::StatusEnum status) {
         if ((WellCommon::StatusEnum::OPEN == status) && getCompletions(timeStep)->allCompletionsShut()) {
-            m_messages.note("when handling keyword for well " + name() + ": Cannot open a well where all completions are shut");
+            m_messages.note("When handling keyword for well " + name() + ": Cannot open a well where all completions are shut");
             return false;
         } else
             return m_status->update( timeStep , status );
--- a/opm/parser/eclipse/EclipseState/Schedule/Well.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Well.cpp
@@ -156,12 +156,15 @@ namespace Opm {
 
     bool Well::setStatus(size_t timeStep, WellCommon::StatusEnum status) {
         if ((WellCommon::StatusEnum::OPEN == status) && getCompletions(timeStep)->allCompletionsShut()) {
-            std::cerr << "ERROR when handling keyword for well "<< name()  << ": Cannot open a well where all completions are shut" << std::endl;
+            m_messages.note("ERROR when handling keyword for well " + name() + ": Cannot open a well where all completions are shut");
             return false;
         } else
             return m_status->update( timeStep , status );
     }
 
+    const MessageContainer& Well::getMessageContainer() const {
+        return m_messages;
+    }
     bool Well::isProducer(size_t timeStep) const {
         return m_isProducer->get(timeStep);
     }
--- a/opm/parser/eclipse/EclipseState/Schedule/Well.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Well.cpp
@@ -156,7 +156,7 @@ namespace Opm {
 
     bool Well::setStatus(size_t timeStep, WellCommon::StatusEnum status) {
         if ((WellCommon::StatusEnum::OPEN == status) && getCompletions(timeStep)->allCompletionsShut()) {
-            m_messages.note("ERROR when handling keyword for well " + name() + ": Cannot open a well where all completions are shut");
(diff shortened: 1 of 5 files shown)


opm/parser/eclipse/EclipseState/Schedule/Well.cpp
```diff
@@ -156,12 +156,15 @@ namespace Opm {
 
     bool Well::setStatus(size_t timeStep, WellCommon::StatusEnum status) {
         if ((WellCommon::StatusEnum::OPEN == status) && getCompletions(timeStep)->allCompletionsShut()) {
-            std::cerr << "ERROR when handling keyword for well "<< name()  << ": Cannot open a well where all completions are shut" << std::endl;
+            m_messages.note("ERROR when handling keyword for well " + name() + ": Cannot open a well where all completions are shut");
```
atgeirr: The string should be changed to not say "ERROR" (since it is not).

opm/parser/eclipse/EclipseState/Tables/VFPProdTable.cpp
```diff
@@ -376,6 +377,11 @@ void VFPProdTable::check() {
                     for (size_type t=0; t<m_data.shape()[0]; ++t) {
                         if (m_data[t][w][g][a][f] < bhp_last) {
                             ++num_decreasing;
+                            points += "At point (FLOW, THP, WFR, GFR, ALQ) = "
```
atgeirr: Will not the numbers here be squashed together and unreadable since there is no space between the to_string-generated strings?

opm/parser/eclipse/EclipseState/Tables/VFPProdTable.cpp
```diff
@@ -385,6 +391,14 @@ void VFPProdTable::check() {
     }
 
     if (num_decreasing > 0) {
+        m_messages.warning("VFP table for production wells has BHP versus THP not " 
+                           + std::string("monotonically increasing.\nThis may cause convergence ")
```
atgeirr: It is amazing how poor old-style C++ string handling is (and therefore also string literals)...

A better way to work around this (cannot do "foo" + "bar") problem is to concatenate the strings:
```
    std::string s("This is\n"
                  "one long"
                  " string.");
```

bska: > (cannot do `"foo" + "bar"`)

True.  When we get C++14, which enables the UDL approach, this becomes more palatable.
```C++
using namespace std::string_literals;  // One of few good reasons for "using namespace"
const auto s = "This is\n"s + "one long" + " string"s;
```

opm/parser/eclipse/EclipseState/Tables/VFPProdTable.cpp
```diff
@@ -385,6 +391,14 @@ void VFPProdTable::check() {
...
+                           + "issues due to switching between BHP and THP control mode."
+                           + std::string("\nIn keyword VFPPROD table number ") 
+                           + std::to_string(m_table_num) 
+                           + ", line " + std::to_string(keyword.getLineNumber()) 
+                           + ", in file " + keyword.getFileName());
+        m_messages.note(points);
         //TODO: Replace with proper log message
         m_messages.debug("VFPPROD bhp versus thp not monotonic increasing: "
```
alfbr: Please remove this message completely, the new one is a full replacement.

opm/parser/eclipse/EclipseState/Tables/VFPProdTable.cpp
```diff
@@ -376,6 +377,11 @@ void VFPProdTable::check() {
                     for (size_type t=0; t<m_data.shape()[0]; ++t) {
                         if (m_data[t][w][g][a][f] < bhp_last) {
                             ++num_decreasing;
```
alfbr: Can be removed if the old message is removed.

atgeirr: No, the if (num_decreasing > 0) is still there.

alfbr: Of course the code will need to be adapted to the new message, but a counter is probably not what you want since all violations will be printed (instead the number of violations, namely num_decreasing).

qilicun: ``num_decreasing``  is removed.

atgeirr:
I think this is ready to be merged (ok @joakim-hove?)

opm/parser/eclipse/EclipseState/Schedule/Well.hpp
```diff
@@ -148,7 +150,7 @@ namespace Opm {
 
         WellCompletion::CompletionOrderEnum m_comporder;
         bool m_allowCrossFlow;
-
+        MessageContainer m_messages;
```
joakim-hove: My oh my - that `MessageContainer` is spreading *everywhere*, but that is not for this PR to consider.

joakim-hove:
The logging and text output facilities are really starting to get in shape - good work!
> I think this is ready to be merged

OK - thank you for the review effort I will merge.

PR #833, opened 2016-06-10
Well related messages should have type MessageType::note

qilicun:
Double check that these messages eclipse print in the terminal, but for the second model, this kind of messages are too much, maybe let it go into print file first is a good choice.


joakim-hove:
I think we agree with respect to the code here; and I am happy to merge this. But we strongly disagree with the literal description of the changes - which I feel hint at a non-existing coupling between messagetypes and logging backends:

> well related message should go into print file

The changes in this PR is that well related messages have messagetype `Note`, what the logging system eventually does with these messages is completely beside the point - and impossible (by design!) to tell from this code. In principle `flow` could have a backend that posted a separate GitHub issue for every message of type `note`.

So - just to verify that we are the same page here; a better heading for this PR would be:

**Well related messages should have type `MessageType::note`**

qilicun:
>Well related messages should have type MessageType::note

Thanks, this heading looks much better, changed.

PR #839, opened 2016-06-14
Added keyword OPERATE.

joakim-hove:
Will parse - *not* internalize - the `OPERATE` keyword.

### DIFF
--- a/opm/parser/eclipse/Parser/tests/ParseContextTests.cpp
+++ b/opm/parser/eclipse/Parser/tests/ParseContextTests.cpp
@@ -27,6 +27,7 @@
 #include <opm/parser/eclipse/Parser/Parser.hpp>
 #include <opm/parser/eclipse/Parser/ParserKeywords/D.hpp>
 #include <opm/parser/eclipse/Parser/ParserKeywords/E.hpp>
+#include <opm/parser/eclipse/Parser/ParserKeywords/O.hpp>
 #include <opm/parser/eclipse/Parser/ParserKeywords/S.hpp>
 #include <opm/parser/eclipse/Parser/ParserKeywords/T.hpp>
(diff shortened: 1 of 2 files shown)


opm/parser/eclipse/Parser/tests/ParseContextTests.cpp
```diff
@@ -82,6 +83,40 @@ BOOST_AUTO_TEST_CASE(TestUnkownKeyword) {
...
+        "SWL    6* LOG10  SWL               / Temp:  SWL=log(1.E10*PERMX)\n"
+        "SWL    6* MULTA  SWL   -0.06  0.91 / Final: SWL=0.31-0.06*log(PERMX)\n"
+        "--SWCR 6* COPY   SWL               / SWCR=SWL\n"
+        "SWCR   6* MULTA  SWL   1.0    0.0  / SWCR=SWL+0.0 (+0.3)\n"
+        "SWCR   6* MAXLIM SWCR  0.7         / max(SWCR)=0.7\n"
+        "SGU    6* MULTA  SWL   -1.0   1.0  / SGU=1-SWL\n"
+        "/\n";
+
+    ParseContext parseContext;
+    Parser parser(false);
```
magnesj: For some reason, if parser is initialized with default keywords addDefaultKeywords
`Parser parser;`

the test fails.

Any idea why?

joakim-hove:
> For some reason, if parser is initialized with default keywords using addDefaultKeywords() with default arguments to the constructor

> Parser parser;

> the test fails.

> Any idea why?

Yes - that is because the first part of the test check assumes that the `OPERATE` keyword is unknown; and if you instantiate parser as `Parser parser( true );` the `OPERATE`keyword is recognized.

Issue #835, opened 2016-06-10
Test case BRILLIG for Eclipse fails to load


magnesj:
Parsing BRILLIG.DATA causes an exception. The problematic section seems to be parsing 
of the COPY keyword. 

```
COPY                                             FIELD   09:58  2 JAN 90
'permx   ' 'permy   '  ,   ,     ,   ,     ,   ,     /
'permx   ' 'permz   '  ,   ,     ,   ,     ,   ,     /
/

MULTIPLY                                         FIELD   09:58  2 JAN 90
'permz   ' 0.05000000  ,   ,     ,   ,     ,   ,     /
/
```

Previous link to case removed, not sure if this is OK to publish.

joakim-hove:
OK; this seems to be at least the ',' issue again - and probably also the `permx` which has two issues:

1. The keyword recognized by the property manipulator is `PERMX` - and not `permx`.
2. There is no trim of keywords; so in this case we will look for the keyword `"permx   "`.

Where does this deck come from? To it seems to be quite archaic.

joakim-hove:
> Parsing BRILLIG.DATA causes an exception. 

Can we have a backtrace?

magnesj:
This case is taken from an installation of Eclipse, provided to us by Håkon, local path from Windows installation is c:/appl/ecl/2013.2/eclipse/data

magnesj:
Here is a backtrace

```
 	KernelBase.dll!000007fefcefa06d()	Unknown
 	vcruntime140d.dll!000007feeea681c0()	Unknown
>	ResInsight.exe!Opm::readValueToken<int>(Opm::string_view view) Line 79	C++
 	ResInsight.exe!Opm::ParserItemScan<Opm::ParserIntItem,int>(const Opm::ParserIntItem * self, Opm::RawRecord & rawRecord) Line 214	C++
 	ResInsight.exe!Opm::ParserIntItem::scan(Opm::RawRecord & rawRecord) Line 90	C++
 	ResInsight.exe!Opm::ParserRecord::parse(const Opm::ParseContext & parseContext, Opm::RawRecord & rawRecord) Line 123	C++
 	ResInsight.exe!Opm::ParserKeyword::parse(const Opm::ParseContext & parseContext, std::shared_ptr<Opm::RawKeyword> rawKeyword) Line 488	C++
[...]

PR #841, opened 2016-06-14
Supporting reading and writing mixed case properties

pgdr:
Uppercase properties before querying (solves #835)

* Added static functions `toupper` in both Eclipse3DProperties and GridProperties
* Renamed some keywords in tests to be mixed case.

This depends on #843.


opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -37,6 +38,12 @@
 
 namespace Opm {
 
+    /// Returns a copy of s where every letter has been upper cased.
+    static std::string toupper(const std::string& s) {
+        auto ret = std::string(s);
+        std::transform(ret.begin(), ret.end(), ret.begin(), ::toupper);
```
jokva: This is pretty dangerous. `::toupper` and friends from C can be implemented as macros and sometimes behave surprising when used with std algorithm. The most common solution is to wrap it in a lambda.

jokva:
With the exception of possibly buggy `toupper` this looks nice, and is a welcome change.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
joakim-hove: > and friends from C can be implemented as macros 

I am not claming this is wrong - but I have not found any reference to a macro based implementation in cpp ref?

joakim-hove:
OK; merge whenever the `std::toupper( )` is either changed - or verified safe.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
jokva: It's one of those cases™ where it depends on whether or not it's provided by types.h, ctypes or something. You can always force it to be a function pointer context.

Regardless, I had a chat with Pål and we'll soon create a separate PR that provides some of the string functions that were thrown out earlier.

Some context: In a refactoring a lot of boost-string use was replaced by a simple inlined transform, because we really only needed to uppercase things once, and the refactoring made data flow easier to follow. Now, on the other hand, the strings being uppercase'd come through a completely different access path, so we cannot rely on input reading doing it for us.

PR #844, opened 2016-06-15
Suppress warnings

atgeirr:
Before you merge this, please check that the unused arguments, especially in `convertInputValue()` do not signify a bug (I have not taken the time to check this carefully).


joakim-hove:
See: #846

PR #846, opened 2016-06-15
Suppress warnings

joakim-hove:
Replaces: #844

### DIFF
--- a/opm/parser/eclipse/EclipseState/EclipseConfig.cpp
+++ b/opm/parser/eclipse/EclipseState/EclipseConfig.cpp
@@ -34,7 +34,6 @@
 namespace Opm {
 
     EclipseConfig::EclipseConfig(const Deck& deck,
-                                 const ParseContext& parseContext,
                                  const Eclipse3DProperties& eclipse3DProperties,
                                  std::shared_ptr< EclipseGrid > inputGrid,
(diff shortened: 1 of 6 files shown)


atgeirr:
A better way, thanks for the improvement.

joakim-hove:
> A better way, thanks for the improvement.

I'll take that as a review - and merge!

opm/parser/eclipse/EclipseState/EclipseConfig.hpp
```diff
@@ -38,7 +38,6 @@ namespace Opm {
     {
     public:
         EclipseConfig(const Deck& deck,
-                      const ParseContext& parseContext,
```
pgdr: Removing this makes `class ParseContext` unnecessary.

opm/parser/eclipse/EclipseState/EclipseConfig.cpp
```diff
@@ -34,7 +34,6 @@
 namespace Opm {
 
     EclipseConfig::EclipseConfig(const Deck& deck,
-                                 const ParseContext& parseContext,
```
pgdr: No longer necessary to `#include <opm/parser/eclipse/Parser/ParseContext.hpp>`

PR #848, opened 2016-06-15
Error msg

joakim-hove:


opm/parser/eclipse/EclipseState/Grid/GridProperty.cpp
```diff
@@ -344,7 +344,7 @@ namespace Opm {
         for (size_t g=0; g < m_data.size(); g++) {
             T value = m_data[g];
             if ((value < min) || (value > max))
-                throw std::invalid_argument("Property element outside valid limits");
+                throw std::invalid_argument("Property element " + std::to_string( value) + " outside valid limits: [" + std::to_string(min) + ", " + std::to_string(max) + "]");
```
bska: Do we have enough information here to print the name of the property too?  If so, would that be useful?

joakim-hove: Yes we do - coming.

pgdr:
A nice change.  Merge at will (after discussed rebase).

PR #849, opened 2016-06-16
Remove shared_ptr Well alias, make Schedule interface more consistent

jokva:
I'm doing some work on output that works with the parser, and I keep finding issues with the parser interface.

This patch set smooths some of the edges in the `Schedule` interface by making `Well` and `Group`-functions have uniform behaviour w.r.t. types. Since this soft-breaks the interface anyway, it's about time to retire `WellPtr` and friends.

Ideally I'd make the Well communication here a reference right away, but that would mean a lot of `->` to `.` changes, which is work for another day. By removing the ownership sharing (which was a ruse) from the interface, replacing `shared_ptr` with raw pointers, little changes in client code.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Group.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Group.cpp
@@ -96,7 +96,7 @@ namespace Opm {
     Group::Group(const std::string& name_, TimeMapConstPtr timeMap , size_t creationTimeStep) :
         m_injection( new GroupInjection::InjectionData(timeMap) ),
         m_production( new GroupProduction::ProductionData( timeMap )),
-        m_wells( new DynamicState<WellSetConstPtr>(timeMap , WellSetConstPtr(new WellSet() ))),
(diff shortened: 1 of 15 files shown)


atgeirr:
The PRs in opm-core and opm-simulators look ok (minor nitpick about auto over-use aside). I tried to focus on possible ownership issues, and could not find any problems in my quick look.

I agree about using references, and also agree that it can come later.

jokva:
There aren't any because all this data is owned by, and was always owned by, `EclipseState`.

atgeirr:
> There aren't any because all this data is owned by, and was always owned by, EclipseState.

Clients could have participated in ownership by retaining the shared_ptr. There is no way to guarantee that does not happen. However, it seems to not actually happen with the objects here.

jokva:
Doesn't matter since we mandate `EclipseState` liveness. There's no reason for clients to participate in ownership, and it was always a bug in my opinion. I've yet to read a single piece of code in opm that uses `shared_ptr` for actual shared ownership.

PR #850, opened 2016-06-16
use boost::regex unconditionally

andlaus:
this fixes OPM/opm-simulators#702. which turned out to be a quirk of the
build system which is due to the fact that opm-parser likes to do some
things differently than the rest of OPM. (in this case the culprit was
that it does not generate a `config.h` file which resulted in the
HAVE_REGEX macro to be different between opm-parser and downstream
which in turn causes the ABI to be incompatible.)

Note that it would be nice to use `std::regex` instead of Boost, but
this is currently blocked by the requirement that the oldest supported
compiler is GCC 4.8. (GCC 4.8 only provides a stub implementation of
`std::regex`.)

### DIFF
--- a/opm/parser/eclipse/Parser/ParserKeyword.cpp
+++ b/opm/parser/eclipse/Parser/ParserKeyword.cpp
@@ -527,11 +527,7 @@ namespace Opm {
 
     void ParserKeyword::setMatchRegex(const std::string& deckNameRegexp) {
         try {
-#ifdef HAVE_REGEX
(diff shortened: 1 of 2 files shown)


jokva:
Nice work! How did you spot it?

andlaus:
I had some suspicions and this one turned out to be true. (when changing this macro in the downstream's config.h it started working...)

jokva:
Sure, and ABI incompatibility was discussed.

PR #852, opened 2016-06-16
Cornerpoint based EclipseGrid constructor.

joakim-hove:
Micro step towards proper EclipseGrid representation of the simulation grid.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.hpp
+++ b/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.hpp
@@ -72,14 +72,21 @@ namespace Opm {
     public:
         explicit EclipseGrid(const std::string& filename);
         explicit EclipseGrid(const EclipseGrid& srcGrid);
-        explicit EclipseGrid(size_t nx, size_t ny, size_t nz,
-                             double dx = 1.0, double dy = 1.0, double dz = 1.0);
+	EclipseGrid(size_t nx, size_t ny, size_t nz,
+		    double dx = 1.0, double dy = 1.0, double dz = 1.0);
+	
+	EclipseGrid(const std::vector<int>& dims , 
+		    const std::vector<double>& coord , 
+		    const std::vector<double>& zcorn , 
+		    const int * actnum = nullptr,
+		    const double * mapaxes = nullptr);
+
 
         /// EclipseGrid ignores ACTNUM in Deck, and therefore needs ACTNUM
         /// explicitly.  If a null pointer is passed, every cell is active.
-        explicit EclipseGrid(const Deck& deck, const int * actnum = nullptr);
+	EclipseGrid(const Deck& deck, const int * actnum = nullptr);
         /// [deprecated]
-        explicit EclipseGrid(const std::shared_ptr<const Deck>& deck, const int * actnum = nullptr);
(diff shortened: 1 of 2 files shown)


atgeirr:
This is probably fine, but I'm a little confused about what an EclipseGrid (and the underlying ert structure) is supposed to represent. My assumption has been that they represent the input specification, but in a way that is agnostic to which keywords were used to specify it. On top of that it does compute some cell-wise geometric properties such as volumes. However the topology or connectivity structure is not part of the concept.

Is that about to change?

joakim-hove:
> Is that about to change?

No - your understanding of the situation is correct; and it is not about to change. However - in addition to input the `EclipseGrid` structure is also used for output - essentially to save an `EGRID` file. Currently the `EGRID` file we save corresponds to the input grid + `ACTNUM` modifications; minor geometry tweaking (essentially `ZCORN` adjustments) in the grid processor is *not* accounted for in the output, that is what I would like to change.

To achieve this we need to be able to create an `EclipseGrid` from a simulation grid, this is probably a 'medium-size' task - and maybe a bit overboard.

atgeirr:
Thanks for the explanation. Another question: why is it important that the EGRID file contains the exact same zcorn that are used in the simulation? If it was used for restarting that would be a very good reason, but we recreate the deck from input upon restart (I think). So is it for visualization of results?

joakim-hove:
> Another question: why is it important that the EGRID file contains the exact same zcorn that are used in the simulation? If it was used for restarting that would be a very good reason, but we recreate the deck from input upon restart (I think). So is it for visualization of results?

It is only for visualization - and possibly other post simulation tasks; maybe not important enough after all?

atgeirr:
If it is only for visualization, I think it may not need to be a high priority in the short term. I expect that for most users, the differences are not going to be important. If you want to inspect a few key cells to really see what is going on there it may be significant, but that opens a new question: do we want to visualize properties on the original (input) cells, or on the modified cells (if using the OPMFIL option)? I don't think the answer is clear-cut.

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.hpp
```diff
@@ -72,14 +72,21 @@ namespace Opm {
     public:
         explicit EclipseGrid(const std::string& filename);
         explicit EclipseGrid(const EclipseGrid& srcGrid);
-        explicit EclipseGrid(size_t nx, size_t ny, size_t nz,
-                             double dx = 1.0, double dy = 1.0, double dz = 1.0);
+	EclipseGrid(size_t nx, size_t ny, size_t nz,
+		    double dx = 1.0, double dy = 1.0, double dz = 1.0);
+	
+	EclipseGrid(const std::vector<int>& dims ,
```
pgdr: Could we have a very short comment on the content of `mapaxes`?

joakim-hove: > Could we have a very short comment on the content of mapaxes?

MAPAXES is an Eclipse input keyword which can be used to rotate and move the grid to real-world location, i.e. UTM coordinates.

pgdr:
Could you add a test?

joakim-hove:
> Could you add a test?

This is well covered by existing Grid tests - so I will self-merge this now.

PR #853, opened 2016-06-20
Minor refactor. Changed shared_ptr to references in EclipseState and related objects

pgdr:
In stabilizing the OPM-parser API, some minor modifications of internal parser API.

* removed unused `MULTREGTScanner` constructor argument
* removed `shared_ptr` usage in `MULTREGTScanner`
* reduced `shared_ptr` in `EclipseState` (`Section`, `GridSection`, `FaultCollection`, `TransMult`)
* some fixes in includes (removed unused, imported used, sorted)

We will soon remove shared pointers from the EclipseState API, but since that requires a larger synchronized PR, we will wait with that until the internal API seems more stable.  This PR contains just some of the internal modifications necessary that do not require downstream changes.

### DIFF
--- a/opm/parser/eclipse/EclipseState/EclipseState.cpp
+++ b/opm/parser/eclipse/EclipseState/EclipseState.cpp
@@ -207,7 +207,7 @@ namespace Opm {
 
 
 
-    void EclipseState::setMULTFLT(const Section section) {
+    void EclipseState::setMULTFLT(const Section& section) {
         for (size_t index=0; index < section.count("MULTFLT"); index++) {
             const auto& faultsKeyword = section.getKeyword("MULTFLT" , index);
             for (auto iter = faultsKeyword.begin(); iter != faultsKeyword.end(); ++iter) {
--- a/opm/parser/eclipse/EclipseState/EclipseState.cpp
+++ b/opm/parser/eclipse/EclipseState/EclipseState.cpp
@@ -76,7 +76,12 @@ namespace Opm {
 
         initTransMult();
         initFaults(deck);
-        initMULTREGT(deck);
+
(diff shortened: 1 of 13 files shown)


opm/parser/eclipse/EclipseState/EclipseState.cpp
```diff
@@ -188,25 +193,23 @@ namespace Opm {
...
 
         m_transMult->applyMULTFLT( m_faults );
     }
 
 
 
-    void EclipseState::setMULTFLT(std::shared_ptr<const Section> section) {
-        for (size_t index=0; index < section->count("MULTFLT"); index++) {
-            const auto& faultsKeyword = section->getKeyword("MULTFLT" , index);
+    void EclipseState::setMULTFLT(const Section section) {
```
joakim-hove: I guess this should be a reference?

opm/parser/eclipse/EclipseState/Grid/FaultCollection.cpp
```diff
@@ -33,42 +37,51 @@
...
-                if (!hasFault(faultName)) {
-                    addFault( faultName );
-                }
-
-                {
-                    Fault& fault = getFault( faultName );
-                    fault.addFace( face );
-                }
+
+                initFault(grid, faultRecord, faultName);
```
joakim-hove: This will typically be called repeatedly - i.e. it is not a typical `init` function. A better name would be `addFaultFaces()`

joakim-hove:
This is ready to be merged when the minor comments have been addressed.

pgdr:
Thanks for the feedback.  I have fixed the comments you had.  I will merge when Travis green-lights.

joakim-hove:
> I will merge when Travis green-lights.

Go ahead.

PR #854, opened 2016-06-20
Introducing GridDims, a wrapper around (x,y,z), superclassing EclipseGrid

pgdr:
This PR introduces GridDims, a class wrapping ints x,y,z
* Many classes only need x,y,z info from EclipseGrid, now these can use GridDims
* GridDims is a superclass for EclipseGrid, extracting simple x,y,z functionality
* GridDims takes a Deck and reads dimens/specgrid keyword to obtain x,y,z
* Config, NNC, and Schedule takes GridDims instead of EclipseGrid

This facilitates a future "lightweight EclipseState" when we don't have a full Deck, we can use GridDims.

I would like feedback on the `m_messages` usage in the `GridDims(Deck&)` constructor.


jokva:
getDims is statically sized and can use std::array. There's also a bug when you assign gridDims to a vector ref, since it's returned by value.

pgdr:
Thanks for the feedback, I have fixed the bug.  I also rewrote the EclipseGrid object to now use `std::array<int, 3>` where we indeed expect `x,y,z`.  Fixed in GridDims as well.

jokva:
Regarding the MessageContainer I need to think on it. If it can wait a week I can provide feedback, but probably not before.

opm/parser/eclipse/EclipseState/Grid/GridDims.hpp
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        GridDims(size_t nx, size_t ny, size_t nz) :
+            m_nx(nx), m_ny(ny), m_nz(nz)
+        {
+        }
+
+        GridDims(const Deck& deck, MessageContainer messages = MessageContainer())
+        {
+            std::vector<int> xyz;
```
joakim-hove: This should in my opinion be simplified massively:
```
private void GridDims::init(const DeckKeyword& keyword) {
   m_nx = keyword.getItem("NX").getInt(0);
   m_ny = keyword.getItem("NY").getInt(0);
   m_nz = keyword.getItem("NZ").getInt(0);
}

GridDims::GridDims(const Deck& deck) {
   if (deck.hasKeyword("SPECGRID"))
      init( deck.getKeyword("SPECGRID"))
   else if (deck.hasKeyword("DIMENS"))
      init( deck.getKeyword("DIMENS"));
   else
      throw std::invalid_argument("Must have either SPECGRID or DIMENS to indicate grid dimesnions");
}
```
This implementation exploits the fact that the `SPECGRID` and `DIMENS` share the `NX,NY,NZ` part of the configuration - if going that way we should probably add a test to enforce that behavior.

opm/parser/eclipse/EclipseState/Grid/GridDims.hpp
```diff
@@ -0,0 +1,174 @@
...
+#include <opm/parser/eclipse/Deck/DeckKeyword.hpp>
+#include <opm/parser/eclipse/Deck/DeckRecord.hpp>
+#include <opm/parser/eclipse/Deck/Section.hpp>
+
+#include <opm/parser/eclipse/Parser/MessageContainer.hpp>
+#include <opm/parser/eclipse/Parser/ParserKeywords/D.hpp> // DIMENS
+#include <opm/parser/eclipse/Parser/ParserKeywords/S.hpp> // SPECGRID
+
+namespace Opm {
+    class GridDims
```
joakim-hove: When there is no templating involved I very much prefer a `GridDims.hpp` & `GridDims.cpp` pair.

joakim-hove:
> I would like feedback on the m_messages usage in the `GridDims(Deck&)` constructor.

I would ignore it completely - and just throw in the constructor. An input which does not have `DIMENS` or `SPECGRID` is broken beyond repair - there is no way we can construct any meaningful `EclipseState` datastructures from such a deck.

joakim-hove:
It has been a (anti)feature of the `EclipseGrid` class that it is possible to instantiate a grid which only has dimension information, and before calling functions to e.g. get the cell volume we must call the `assertCellInfo()` method first. The `GridDims` class should fill this role, and we should now *require* that an `EclipseGrid` has all the relevant information; i.e.

1. The whole `assertCellInfo()` functionality should be removed - an `EclipseGrid` instance should have a valid `ecl_grid_type` pointer - period.
2. The `initGrid( )` should throw [here](https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp#L225).
3. The `initGrid( )` still initializes the dimensions - that has already been done in the `GridDims` constructor.

opm/parser/eclipse/EclipseState/Grid/GridDims.hpp
```diff
@@ -0,0 +1,174 @@
...
+        GridDims() : m_nx(0), m_ny(0), m_nz(0) {}
+
+        const std::array<int, 3> getDims() {
+            return std::array<int, 3> {static_cast<int>(m_nx),
+                                       static_cast<int>(m_ny),
+                                       static_cast<int>(m_nz)};
+        }
+
+        // keyword must be DIMENS or SPECGRID
+        std::vector<int> getDims( const DeckKeyword& keyword ) {
```
joakim-hove: This looks like it could be `static` - or even a loose function?

opm/parser/eclipse/EclipseState/Grid/GridDims.hpp
```diff
@@ -0,0 +1,174 @@
...
+        void assertIJK(size_t i, size_t j, size_t k) const {
+            if (i >= getNX() || j >= getNY() || k >= getNZ())
+                throw std::invalid_argument("input index above valid range");
+        }
+
+
+    protected:
+        GridDims() : m_nx(0), m_ny(0), m_nz(0) {}
+
+        const std::array<int, 3> getDims() {
```
joakim-hove: Any reason to have both `getDims()` and `getNXYZ()`?

joakim-hove:
Good if the [`FaultCollection`](https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Grid/FaultCollection.hpp) is modified to use `GridDims` argument - that will benefit Resinsight usage, and is probably the first call site where this refactoring gives value.

joakim-hove:
The new file `GridDims.hpp` is not added to `CMakeLists.txt`; the build also fails on travis due to a non-satisfied dependency on the generated keywords. This would come for free if the implementation was in a `.cpp` file which was added [here](https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/CMakeLists.txt#L111) - i.e. with a dependency on the `build-parser` target.

atgeirr:
Just my $0.02 based on the PR description: I think EclipseGrid inheriting GridDims sounds a bit odd, inheritance should model an "is-a" relationship, and an EclipseGrid is not a GridDims in the sense that you might really want or need to substitute an EclipseGrid for a GridDims object. I would suggest a "has-a" relationship instead (EclipseGrid containing a GridDims object).

joakim-hove:
>I think EclipseGrid inheriting GridDims sounds a bit odd, 

Might be - the situation we are trying to solve/improve upon is the following:

1. Many of the large subobjects in the `EclipseState` take `EclipseGrid` arguments in their constructor; but the only real requirement is the triplet `(n_x, n_y, n_z)`.

2. As part of the full `EclipseState` construction we will create an `EclipseGrid` instance, so there we will indeed have a full `EclipseGrid` lying around.

What we achieve with inheritance is that we can transparently use the `GridDims` object when explicitly creating e.g. the `Schedule` object from the outside, and just use the `EclipseGrid` instance from inside the `EclipseState` as part of the "large construction process".

The main point about this change is that e.g. the `Schedule`constructor more clearly communicates what it needs, and that it becomes simpler to create explicitly. The relationship between `EclipseGrid` and `GridDims` is not important - both "has-a" and "creates-a-on-the-fly" would work.

pgdr:
I disagree with Atgeirr on this one.  I definitely think that `EclipseGrid` is a 3D object with some more features.  The _name_ `GridDims` might be suboptimal, but if it were called `3DGrid` I assume we would agree that `EclipseGrid` "would-be-a" `3DGrid`.

opm/parser/eclipse/EclipseState/Grid/GridDims.cpp
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    const std::array<int, 3> GridDims::getDims() const {
+        return std::array<int, 3> { static_cast<int>(m_nx), static_cast<int>(m_ny), static_cast<int>(m_nz) };
+    }
+
+    // keyword must be DIMENS or SPECGRID
+    const std::vector<int> GridDims::getDims(const DeckKeyword& keyword) {
+        const auto& record = keyword.getRecord(0);
+        return std::vector<int> { record.getItem("NX").get<int>(0), record.getItem("NY").get<int>(0), record.getItem(
```
joakim-hove: Now - this was really nicely formatted :-)

opm/parser/eclipse/EclipseState/Grid/GridDims.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+        void assertIJK(size_t i, size_t j, size_t k) const;
+
+    protected:
+        GridDims();
+
+        const std::array<int, 3> getDims() const;
+
+        // keyword must be DIMENS or SPECGRID
+        const std::vector<int> getDims(const DeckKeyword& keyword);
```
joakim-hove: `static`?

PR #856, opened 2016-06-21
adding keyword WECON

GitPaean:
There are some part I am not sure. 

Like the item 12, DEFAULT, the same action as in item 7.


joakim-hove:
> There are some part I am not sure.

Everything looks fine to me; what exactly was your question?

GitPaean:
> Like the item 12, DEFAULT, the same action as in item 7.

When we use `DEFAULT` for item 12, it means it should take the value from item 7. I think for this we can only handle it later, right?
I do not have other question.

joakim-hove:
> When we use DEFAULT for item 12, it means it should take the value from item 7.

OK - I understand. You are right - that must be handled in the application layer. I'll merge.

PR #857, opened 2016-06-21
fixed: broken keyword in e576f399f18da63172aa983d3f630e05111e2b0d

akva2:
A mere compilation test showed the failure here.

Repeat after me: I WILL NEVER EVER PULL UNTESTED CODE INTO A PRODUCTION GIT.  Now go repeat it 10000 times bart simpson style until it sticks.

### DIFF
--- a/opm/parser/share/keywords/000_Eclipse100/W/WECON
+++ b/opm/parser/share/keywords/000_Eclipse100/W/WECON
@@ -5,7 +5,7 @@
        {"name" : "MAX_WATER_CUT", "value_type" : "DOUBLE",   "default" : 0.0},
        {"name" : "MAX_GAS_OIL_RATIO", "value_type" :  "DOUBLE",   "default" : 0.0},
        {"name" : "MAX_WATER_GAS_RATIO", "value_type" :  "DOUBLE",  "default" : 0.0},
-       {"name" : "WORKOVER_RATIO_LIMIT", "value_type" : "STRING", "default" : "NONE},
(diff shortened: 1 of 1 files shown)


akva2:
note that my fix may we off if an enum value NONE was intended.

GitPaean:
Sorry. I did compile, while I am not sure why I did not see the problem. I will try it again now. Thanks for the fix.

akva2:
that's a bit disconcerning. it's likely that a clean build that caused the problem. is there a dependency tracking issue in the buildsystem. @joakim-hove perhaps ?

GitPaean:
Now I see it is broken. Thanks for the fix. I saw another error when compiling and I corrected it. I did not see this one. Sorry for that.

akva2:
right. that explains it. the build system is based on globbing for keywords. since your commit was a pure keyword-add it wouldn't cause a rebuild and you wouldn't see the issue. you likely edited the file after it built right?

GitPaean:
Maybe, while I am not sure about that.

joakim-hove:
Ok: my bad. Travis was slow to start so I just tought what the ....

The build system has a dependency checking on the keywords which I thougt worked,  but I guess you can fool it.

Will wait for Travis now.

pgdr:
While I normally _insist_ on waiting for Travis, in this case I would say that we can just merge.  If it still fails, it's something more that needs to be done, but this fix needs to go in anyway.

PR #858, opened 2016-06-24
Added IOConfig::getOutputBase()

joakim-hove:
The purpose of this PR is to get the combination of output path and basename as one variable; that can then be sent directly to e.g. the summary output routine.

### DIFF
--- a/opm/parser/eclipse/EclipseState/IOConfig/IOConfig.hpp
+++ b/opm/parser/eclipse/EclipseState/IOConfig/IOConfig.hpp
@@ -161,6 +161,10 @@ namespace Opm {
         const std::string& getBaseName() const;
         void setBaseName(std::string baseName);
 
+        /// Return a string consisting of outputpath and basename;
+        /// i.e. /path/to/sim/CASE
+        std::string fullBasePath( ) const;
+
     private:
 
         IOConfig( const GRIDSection&,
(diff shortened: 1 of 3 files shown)


opm/parser/eclipse/EclipseState/IOConfig/IOConfig.hpp
```diff
@@ -161,6 +161,10 @@ namespace Opm {
         const std::string& getBaseName() const;
         void setBaseName(std::string baseName);
 
+        /// Return a string consisting of outputpath and basename;
+        /// i.e. /path/to/sim/CASE
+        std::string getOutputBase( ) const;
```
pgdr: The naming `baseName` and `outputBase` confuses me.  Could it be possible to rename `outputBase` to `fullPath` or `fullBasePath` or something?

Why does `getOutputBase` need to contain _output_ while `getBaseName` doesn't?

pgdr:
Could you add a description for the PR?  E.g. why/where is this needed?

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.hpp
joakim-hove: > Could it be possible to rename outputBase to fullPath or fullBasePath or something?

`fullBasePath` was a good idea.

joakim-hove:
OK - this is green and trivial; self-merging.

PR #859, opened 2016-06-27
Add missing namespace and include file

ANerd:
Functions from `<cmath>` are used without namespace and without including the header.

I guess these have leaked from other include files, but with gcc 6.1.1 I get compile errors.


pgdr:
Thanks! I'll merge this once the tests finish.

pgdr:
Thanks for the contribution!

PR #862, opened 2016-06-28
Improved well properties interface

jokva:
Two convenince call for downstream users that allows querying well rates
with the assumption that if something is unsupported or of the wrong
type, return plain zero.

Adds the "identity" unit which is a no-op under to_si/from_si
conversion.

These are both features I find useful for downstream development.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Well.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Well.cpp
@@ -88,6 +88,33 @@ namespace Opm {
     }
 
 
+    double Well::production_rate( Phase::PhaseEnum phase, size_t timestep ) const {
+        if( !this->isProducer( timestep ) ) return 0.0;
+
+        const auto& p = this->getProductionProperties( timestep );
+
+        switch( phase ) {
+            case Phase::WATER: return p.WaterRate;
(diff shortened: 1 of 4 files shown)


opm/parser/eclipse/EclipseState/Schedule/Well.cpp
```diff
@@ -88,6 +88,33 @@ namespace Opm {
...
+                                "This is likely a programming error." );
+    }
+
+    double Well::injection_rate( Phase::PhaseEnum phase, size_t timestep ) const {
+        if( !this->isInjector( timestep ) ) return 0.0;
+
+        const auto& i = this->getInjectionProperties( timestep );
+        const auto type = i.injectorType;
+
+        if( phase == Phase::WATER && type != WellInjector::WATER ) return 0.0;
```
joakim-hove: Just a question out of curiosity: Would it be possible to let one enum be a subclass of the other; and thereby write:
```C++
if (phase != type) return 0.0;
```

Don't spend time on it ...

jokva: Would be nice, but unfortunately there's no such mechanism in C++.

What we **could** do is let it be a bitset and let mixed injectors be an or'd enum, but that requires some other considerations.

PR #863, opened 2016-06-28
TimeMap bug for RPTRST BASIC=4 terminated before a year!

joakim-hove:
Downstream: 

opm-output: https://github.com/OPM/opm-output/pull/54
~~opm-data: https://github.com/OPM/opm-data/pull/112~~

PR #864, opened 2016-06-29
Moved NOSIM handling to IOConfig + allow override.

joakim-hove:
The IOConfig object is maybe not the perfect place for it - but that object is already mutable.

The purpose of this is to allow easier testing of INIT and EGRID files from flow, by allowing somthing like 'flow --nosim FILE.DATA'


joakim-hove:
Any objections to merging this?

jokva:
As a temporary measure I think it's ok. As evident by https://github.com/OPM/opm-parser/pull/863#issuecomment-229347291 however I think we need to figure out a better way to communicate the NOSIM option.

jokva:
Also, won't this break downstream?

PR #865, opened 2016-06-29
Support  of Keyword WECON. 

GitPaean:
To handle the automatic shut-in related, added the flag to indicate automatic shut-in to the Well class.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Well.hpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Well.hpp
@@ -45,8 +45,10 @@ namespace Opm {
 
     class Well {
     public:
-        Well(const std::string& name, std::shared_ptr<const EclipseGrid> grid , int headI, int headJ, Value<double> refDepth , Phase::PhaseEnum preferredPhase,
-             std::shared_ptr< const TimeMap > timeMap, size_t creationTimeStep, WellCompletion::CompletionOrderEnum completionOrdering = WellCompletion::TRACK,
+        Well(const std::string& name, std::shared_ptr<const EclipseGrid> grid, int headI,
+             int headJ, Value<double> refDepth , Phase::PhaseEnum preferredPhase,
+             std::shared_ptr< const TimeMap > timeMap, size_t creationTimeStep,
+             WellCompletion::CompletionOrderEnum completionOrdering = WellCompletion::TRACK,
              bool allowCrossFlow = true, bool automaticShutIn = true);
         const std::string& name() const;
 
@@ -107,7 +109,7 @@ namespace Opm {
         bool                           setSolventFraction(size_t timeStep , const double fraction);
         const double&                  getSolventFraction(size_t timeStep) const;
 
-        bool                            setEconProductionLimits(const size_t timeStep, const WellEconProductionLimits& properteis);
+        bool                            setEconProductionLimits(const size_t timeStep, const WellEconProductionLimits& productionlimits);
         const WellEconProductionLimits& getEconProductionLimits(const size_t timeStep) const;
 
         int  firstRFTOutput( ) const;
--- a/opm/parser/eclipse/EclipseState/Schedule/Well.hpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Well.hpp
@@ -46,7 +46,8 @@ namespace Opm {
     class Well {
     public:
         Well(const std::string& name, std::shared_ptr<const EclipseGrid> grid , int headI, int headJ, Value<double> refDepth , Phase::PhaseEnum preferredPhase,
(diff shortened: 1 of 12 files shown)


joakim-hove:
This looks good - test and all! Will review more thoroughly tomorrow; only minor comment so far is that `WellEconProductionLimits` is *really* long - what about `WellEconLimits`?

GitPaean:
Yes. We can do that, though it is really only applied to production wells. I will address that with other comments together tomorrow.

opm/parser/eclipse/EclipseState/Schedule/Well.hpp
```diff
@@ -105,6 +107,9 @@ namespace Opm {
         bool                           setSolventFraction(size_t timeStep , const double fraction);
         const double&                  getSolventFraction(size_t timeStep) const;
 
+        bool                            setEconProductionLimits(const size_t timeStep, const WellEconProductionLimits& properteis);
```
pgdr: properties*

opm/parser/eclipse/EclipseState/Schedule/Well.hpp
```diff
@@ -45,7 +46,8 @@ namespace Opm {
     class Well {
     public:
         Well(const std::string& name, std::shared_ptr<const EclipseGrid> grid , int headI, int headJ, Value<double> refDepth , Phase::PhaseEnum preferredPhase,
-             std::shared_ptr< const TimeMap > timeMap, size_t creationTimeStep, WellCompletion::CompletionOrderEnum completionOrdering = WellCompletion::TRACK, bool allowCrossFlow = true);
+             std::shared_ptr< const TimeMap > timeMap, size_t creationTimeStep, WellCompletion::CompletionOrderEnum completionOrdering = WellCompletion::TRACK,
```
pgdr: I don't have strong opinions on this, but I think 160 character lines is a bit over the top.  Could you perhaps break these lines?

testdata/integration_tests/SCHEDULE/SCHEDULE_WECON
```diff
@@ -0,0 +1,44 @@
+START
+  10 MAI 2007   /
```
pgdr: MAY?

pgdr:
It looks good!

joakim-hove:
I am not saying you *should* do this; but have you considered extending the [`WellProductionProperties`](https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.cpp) structure instead of adding a completely new structure?

GitPaean:
> I am not saying you should do this; but have you considered extending the WellProductionProperties structure instead of adding a completely new structure?

Yes. I thought about that. `WellProductionProperties` is mostly about the well control for the production wells, and it will enter the well equations (maybe not everything).  `WECON` is really about whether the running of wells still economical. We shut/close some wells or connections when the output of the well is not good enough anymore after running for each report step. They are kind of independent of the purpose and usage. But for sure, they are all about the production wells. We can possibly put the whole `WellEconLimits` as a whole in the `WellProductionProperties`, while it may not bring much benefit. Implementing the `WECON` on top of `WellProductionProperties` will make the `WellProductionProperties` much longer and more complicated, i do not think it is a good idea.

testdata/integration_tests/SCHEDULE/SCHEDULE_WECON
joakim-hove: Eclipse seems to prefer 'MAI'?!

joakim-hove:
> i do not think it is a good idea.

Fair enough - it was just a question. I will merge when you addressed the micro issues raised in the review:

1.  [Long lines](https://github.com/OPM/opm-parser/pull/865#discussion_r68969691)
2.  [Spelling](https://github.com/OPM/opm-parser/pull/865#discussion_r68969463)
3. Optional: [rename] (https://github.com/OPM/opm-parser/pull/865#issuecomment-229382710)

testdata/integration_tests/SCHEDULE/SCHEDULE_WECON
GitPaean: `testdata/integration_tests/SCHEDULE/` there are lot of decks using `MAI`. I am not sure how Eclipse and OPM handle this. From the manual, it should be `MAY`.

GitPaean:
> Fair enough - it was just a question. I will merge when you addressed the micro issues raised in the review:

> Long lines
> Spelling
> Optional: rename

I broke the long lines and corrected the spelling. I chose to keep the name. To me, it is slightly better to remind that the keyword is for production well only.  Please let me know if you insist in changing it.

joakim-hove:
In she goes!

PR #866, opened 2016-06-30
Added EclipseGrid::getActiveMap().

joakim-hove:
For *now* mainly for testing of the output writer.


opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
```diff
@@ -674,7 +674,17 @@ namespace Opm {
         ecl_grid_init_zcorn_data_double( c_ptr() , zcorn.data() );
     }
 
-
+    std::vector<int> EclipseGrid::getActiveMap() const {
+        std::vector<int> activeMap( getNumActive() );
+	for (int global_index = 0; global_index < static_cast<int>(getCartesianSize()); global_index++) {
```
jokva: There's no reason to change just because of this, but for reference:

`int( getCartesianSize() )` does the same thing but better.

PR #867, opened 2016-07-01
Disable Python in jenkins build.

joakim-hove:


joakim-hove:
Desperately trying to fix.

PR #869, opened 2016-07-04
[SummaryConfig] Refactoring & remove duplicate entries

jokva:
This PR gives some love to the `SummaryConfig` internals and also adds a new feature to address https://github.com/OPM/opm-output/issues/57

Notable additions:
* `SummaryConfig` **no longer contains duplicate entries**.
* Some initialization list support for `Deck`
* Support for merging two `SummaryConfig` objects
* `SummaryConfig` internals refactored.

### DIFF
--- a/opm/parser/eclipse/Deck/Deck.cpp
+++ b/opm/parser/eclipse/Deck/Deck.cpp
@@ -132,6 +132,14 @@ namespace Opm {
         m_dataFile("")
     {}
 
+    Deck::Deck( std::initializer_list< DeckKeyword > ilist ) :
+        Deck( std::vector< DeckKeyword >( ilist.begin(), ilist.end() ) )
+    {}
+
+    Deck::Deck( std::initializer_list< std::string > ilist ) :
+        Deck( std::vector< DeckKeyword >( ilist.begin(), ilist.end() ) )
+    {}
+
     void Deck::addKeyword( DeckKeyword&& keyword ) {
         this->keywordList.push_back( std::move( keyword ) );
 
--- a/opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp
+++ b/opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp
@@ -47,7 +47,11 @@
 namespace Opm {
 
 namespace {
-    constexpr auto ALL_keywords =  {
+    /* A dummy deck that holds a Summary section with the keyword list that ALL
+     * expands to, plus the SUMMARY header
+     */
+    const Deck ALL_keywords = {
+        "SUMMARY",
         "FAQR",  "FAQRG", "FAQT", "FAQTG", "FGIP", "FGIPG", "FGIPL",
         "FGIR",  "FGIT",  "FGOR", "FGPR",  "FGPT", "FOIP",  "FOIPG",
         "FOIPL", "FOIR",  "FOIT", "FOPR",  "FOPT", "FPR",   "FVIR",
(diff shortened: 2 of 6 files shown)


opm/parser/eclipse/Deck/Deck.cpp
```diff
@@ -132,6 +132,14 @@ namespace Opm {
         m_dataFile("")
     {}
 
+    Deck::Deck( std::initializer_list< DeckKeyword > ilist ) :
+        Deck( std::vector< DeckKeyword >( ilist.begin(), ilist.end() ) )
+    {}
```
joakim-hove: Out of curiosity: will this `move` out of the initializer list?

opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp
```diff
@@ -66,42 +66,42 @@ namespace {
...
-        if (parseContext.get( ParseContext::SUMMARY_UNKNOWN_GROUP) == InputError::WARN)
-            std::cerr << "ERROR: " << msg << std::endl;
+    2. Print a message on stderr.
 
-        parseContext.handleError( ParseContext::SUMMARY_UNKNOWN_GROUP , msgContainer , msg );
-    }
+    The case of incorrectly/missing well/group names in the SUMMARY
+    section did just not seem important enough to warrant the
+    refactoring required to pass a mutable proper MessageContainer
+    all the way down here.
```
joakim-hove: I take it you are "happy" with the cheating?!

joakim-hove:
I liked the `initializer_list` based constructors - I will try to add those to my C++ toolchest.

opm/parser/eclipse/Deck/Deck.cpp
jokva: Maybe. 

C++11:
> The underlying array is not guaranteed to exist after the lifetime of the original initializer list object has ended. The storage for std::initializer_list is unspecified (i.e. it could be automatic, temporary, or static read-only memory, depending on the situation).

C++14:
> The underlying array is a temporary array (until C++17)a materialized array prvalue (since C++17) of type const T[N], in which each element is copy-initialized (except that narrowing conversions are invalid) from the corresponding element of the original initializer list. The lifetime of the underlying array is the same as any other temporary object, except that initializing an initializer_list object from the array extends the lifetime of the array exactly like binding a reference to a temporary (with the same exceptions, such as for initializing a non-static class member). The underlying array may be allocated in read-only memory. 

prvalue:
> A prvalue ("pure rvalue") expression is an expression that does not have identity and can be moved from. 

So yes, it will be moved if it can or is feasible, typically, but initializer_lists can largely be considered a compile-time construct so my bet is that most of the time the constructor is only invoked statically.

jokva: ... and for convenience, sources.
http://en.cppreference.com/w/cpp/utility/initializer_list
http://en.cppreference.com/w/cpp/language/value_category

atgeirr:
I'm pretty sure this required an ERT update (adding a move constructor to handle the sorting of `std::vector< ERT::smspec_node >& vec`). Please send an email to the OPM mailing list when ERT updates are required! If you are afraid you will be spamming because it is happening very often, consider doing it less often :-)

blattms:
On Tue, Jul 05, 2016 at 12:38:05AM -0700, Atgeirr Flø Rasmussen wrote:
> I'm pretty sure this required an ERT update (adding a move constructor to handle the sorting of `std::vector< ERT::smspec_node >& vec`).

Is this  a typo or are we really using a vector of references? I doubt
that this is valid in C++

atgeirr:
It's a reference to a vector, so no problem (text copied from the function argument).

jokva:
It does, and it's a bug that it was missing in the first place.

PR #873, opened 2016-07-11
Appveyor support

jokva:
Appveyor http://appveyor.com/ is a CI service for Windows. Since parser is meant to be Windows compatible we need this feature. Like travis and gitlab, it is driven by an yml file.

In addition, this PR fixes tests that break on Windows.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,36 @@
+version: 1.0.{build}
+clone_depth: 1
+
+configuration:
+    - Debug
+    - Release
+
+os: Visual Studio 2015
+image: Visual Studio 2015
+
+platform:
+    - x64
+
+environment:
+    BOOST_ROOT: C:\Libraries\boost_1_59_0\boost
+    BOOST_INCLUDEDIR: C:\Libraries\boost_1_59_0
+    BOOST_LIBRARYDIR: C:\Libraries\boost_1_59_0\lib64-msvc-14.0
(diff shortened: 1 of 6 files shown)


appveyor.yml
```diff
@@ -0,0 +1,36 @@
...
+    - Release
+
+os: Visual Studio 2015
+image: Visual Studio 2015
+
+platform:
+    - x64
+
+environment:
+    BOOST_ROOT: C:\Libraries\boost_1_59_0\boost
```
joakim-hove: Hehe - I recognize a hard coded filename when I see it :-)

joakim-hove:
Thank you for a herkulean effort getting Windows CI in place. I guess there is a good reason for it - and I am happy to merge this - but boost configuration in the `.yml` file looked kind-of-funny?

appveyor.yml
jokva: That's ok though, it's the path to the appveyor provided install, which is documented and all.

joakim-hove: Ok - merging.

jokva:
It does, but it's rather reasonable. Appveyor provides a few pre-installed packages in the image, amongst them boost, and they all reside in `C:\Libraries`. Now, without a BOOST_ROOT suggestion, boost is not discovered, and there is no dynamic or relative way to obtain the path. Shouldn't change too often so I don't really consider it a problem.

PR #877, opened 2016-07-13
Remove warnings

jokva:
This PR accomplishes two things:

1. It increases the warning level significantly. The previous warning level wasn't properly set because `CMAKE_CXX_FLAGS` typically was defined. The warnings have been made default by unconditionally setting them (when not microsoft), but only in debug mode. Debug mode is now the default build configuration.
2. I've gone through all warnings generated by this new warning set and fixed them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,12 @@ enable_language(C)
 include(CTest)
 include(TestCXXAcceptsFlag)
 
+if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+    message(STATUS "No build type specified - defaulting to 'Debug'.")
+    set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "Choose build type." FORCE)
+    set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Release" "Debug" "RelWithDebInfo" "MinSizeRel")
+endif()
+
 # project information is in dune.module. Read this file and set variables.
 # we cannot generate dune.module since it is read by dunecontrol before
 # the build starts, so it makes sense to keep the data there then.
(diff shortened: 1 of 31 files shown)


CMakeLists.txt
```diff
@@ -77,20 +77,19 @@ endif (USE_RUNPATH)
 
 if (MSVC)
   add_definitions( "/W3 /D_CRT_SECURE_NO_WARNINGS /wd4996 /wd4244 /wd4267")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
+  set(CMAKE_CXX_FLAGS "/MP ${CMAKE_CXX_FLAGS}")
 else()
-  if (NOT DEFINED CMAKE_CXX_FLAGS)
```
blattms: We should give people the chance to overwrite these flags.

Maybe introducing a CACHE variable OPM_CMAKE_CXX_FLAGS_DEBUG and and adding them to CMAKE_CXX_FLAGS_DEBUG is an option
```
set(OPM_CMAKE_CXX_FLAGS_DEBUG  "-pipe -std=c++0x -pedantic -Wall -Wextra -Wformat-nonliteral -Wcast-align -Wpointer-arith -Wmissing-declarations -Wcast-qual -Wshadow -Wwrite-strings -Wchar-subscripts -Wredundant-decls"
   CACHE STRING "Default flags for OPM")
set(CMAK_CXXFLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} ${OPM_CMAKE_CXX_FLAGS_DEBUG}")
```
IMHO the same should be done for the other flags (RELEASE, DEBUG), too.

By the way, what happens if we rerun CMake? Do they get duplicated? I guess that might have been  a reason for the former  if statement.

jokva: They get to override them now (by passing the respective -Wno or -std= flags).

I can try running cmake twice, give me a second.

jokva: Running cmake twice doesn't seem to duplicate the list.

joakim-hove:
Merge?

PR #878, opened 2016-07-13
Gridproperties iterator

joakim-hove:
Basic support to iterate over `GridProperty<T>` instances from a `GridProperties<T>` collection. Slightly over engineered solution to avoid the `std::pair< >` of the `std::map<>::iterator` and also to esacpe from the `std::shared_ptr< >`


opm/parser/eclipse/EclipseState/Grid/GridProperties.hpp
```diff
@@ -58,13 +58,22 @@ namespace Opm {
...
-
     class Eclipse3DProperties;
 
+    template <typename T> class GridPropertyIterator;
+
+    template <typename T>
+    using storageType = typename std::map<std::string , std::shared_ptr<GridProperty<T> > >;
+
+    template <typename T>
+    using  storageIteratorType = typename std::map<std::string , std::shared_ptr<GridProperty<T> > >::const_iterator;
```
jokva: Why don't you leverage your own aliases?

`template< typename T >` using const_iterator = storage< T >::const_iterator`

jokva:
Isn't it better to just kill `shared_ptr` as storage, rather than going through this?

opm/parser/eclipse/EclipseState/Grid/GridProperties.cpp
```diff
@@ -204,6 +204,16 @@ namespace Opm {
 
 
     template< typename T >
+    void GridProperties<T>::insertKeyword(const SupportedKeywordInfo& supportedKeyword) const {
+        int nx = m_eclipseGrid.getNX();
+        int ny = m_eclipseGrid.getNY();
+        int nz = m_eclipseGrid.getNZ();
+
+        m_properties.emplace( std::make_pair(supportedKeyword.getKeywordName() , GridProperty<T>( nx, ny , nz , supportedKeyword )));
```
jokva: If you're using `emplace` you don't have to go via pair (that's the point). Save yourself some trouble:

`emplace( keywordname, GridProperty< T >( nx, ny, nz, supported )  )`

You might even be able to omit GridProperty< T > by using `{}`.

opm/parser/eclipse/EclipseState/Grid/GridProperties.hpp
```diff
@@ -124,6 +146,17 @@ namespace Opm {
         void handleCOPYREGRecord( const DeckRecord& record, const GridProperty<int>& regionProperty );
 
 
+        struct iterator : public storageIteratorType {
+            iterator( storageIteratorType iter ) : storageIteratorType( iter ) { }
```
jokva: Is this even necessary? Anyway, `using storageIterator::storageIterator` accomplishes the same.

opm/parser/eclipse/EclipseState/Grid/GridProperties.hpp
```diff
@@ -124,6 +146,17 @@ namespace Opm {
...
+
+            const GridProperty<T>& operator*( ) {
+                const auto& pair = storageIteratorType::operator*( );
+                return pair.second;
```
jokva: `return this->storageIteratorType::operator*().second` ;)

joakim-hove:
Damn - I'm proud!

jokva:
This currently has a subtle bug: Properties are created twice *if* they're auto generated. I suppose `autogeneratedProperties` should be `reference_wrapper` or just plain pointer.

opm/parser/eclipse/EclipseState/Grid/GridProperties.hpp
joakim-hove: Well - readability is in the mind of the beholder; I prefer the one with the temporary.

joakim-hove:
> This currently has a subtle bug: Properties are created twice if they're auto generated. I suppose autogeneratedProperties should be reference_wrapper or just plain pointer. 

I might misunderstand you here - but I think you are wrong; the member `m_autoGeneratedProperties`is *only* a `std::set<std::string` of keyword names - not actual keywords.

joakim-hove:
> I might misunderstand you here

And in case I am in the wrong (heaven forbid!) - well spotted; I still haven't found it.

jokva:
You're absolutely right and I'm a potato.

joakim-hove:
Merge?

jokva:
Let me have a quick second look and we're good to go.

PR #880, opened 2016-07-14
COMPDAT I/J can be defaulted

jokva:
According to the manual, defaulting or putting 0 for I and J coordinates
will default to the well head I/J.


opm/parser/eclipse/EclipseState/Schedule/Completion.cpp
```diff
@@ -136,12 +138,22 @@ namespace Opm {
...
         std::vector<CompletionPtr> completions;
-        std::string well = compdatRecord.getItem("WELL").getTrimmedString(0);
+
         // We change from eclipse's 1 - n, to a 0 - n-1 solution
-        int I = compdatRecord.getItem("I").get< int >(0) - 1;
-        int J = compdatRecord.getItem("J").get< int >(0) - 1;
+        // I and J can be defaulted with 0 or *, in which case they are fetched
+        // from the well head
+        const auto& itemI = compdatRecord.getItem( "I" );
+        const auto defaulted_I = itemI.defaultApplied( 0 ) || itemI.get< int >( 0 ) == 0;
```
joakim-hove: Well  - in my opinion this is a bit easier on the eye:
```C++
int I;
if (itemI.defaultApplied( 0 ) || (itemI.get< int >( 0 ) == 0))
   I = I = well.getHeadI( );
else
  I = itemI.get< int >( 0 ) - 1
```

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -1118,13 +1118,12 @@ namespace Opm {
 
 
     void Schedule::handleCOMPDAT( const DeckKeyword& keyword, size_t currentStep) {
-        std::map<std::string , std::vector< CompletionPtr> > completionMapList = Completion::completionsFromCOMPDATKeyword( keyword );
-        std::map<std::string , std::vector< CompletionPtr> >::iterator iter;
+        const auto wells = this->getWells( currentStep );
+        auto completions = Completion::fromCOMPDAT( keyword, wells );
+
+        for( const auto pair : completions )
```
joakim-hove: Reference?

testdata/integration_tests/SCHEDULE/SCHEDULE_COMPDAT1
```diff
@@ -1,3 +1,19 @@
+DIMENS
```
joakim-hove: Maybe a bit excessive grid size for a test?

opm/parser/eclipse/EclipseState/Schedule/Completion.cpp
joakim-hove: Alternatively you can abuse that fact that the default value is 0:
```C++
int I = itemI.get< int >( 0 ) - 1;
if (I < 0)
   I = well.getHeadI( );
```

jokva: Both have the drawback that they leave the value non-const, unfortunately, and this is inherently an *expression*. I might be able to beautify it a tad though, I'll have a look.

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
jokva: `getWells` returns a vector by value.

I'd like us to start returning proxies with iterators for things like this, but we're not quite there yet unfortunately.

testdata/integration_tests/SCHEDULE/SCHEDULE_COMPDAT1
jokva: Oh yeah. I'll update that, essentially I just put in something massive to make sure I didn't get out-of-bounds errors.

opm/parser/eclipse/EclipseState/Schedule/Completion.cpp
joakim-hove: Well - you are *never* going to compose anything from this expression, the interesting unit here is the entire method.

jokva:
Aaaand updated.

opm/parser/eclipse/EclipseState/Schedule/Completion.cpp
jokva: Which still requires scanning and understanding all code paths, and leaves the question of *why*.

Syntactically it's less noisy obviously, 100%.

joakim-hove: >and leaves the question of *why*.

No - in my opinion it does not a leave a question at all; your alternative is way more cumbersome to read. Not to mention the `defaulted_I` and `default_J` variables.

But if you strongly prefer the current version we'll leave at that. I saw the testing had improved significantly :+1:

jokva: Weak phrasing on my part. Here's a slightly more elaborate justification:

* `defaulted_I` should make it obvious that the boolean expression on the right hand side are the two conditions that would lead to this field being considered defaulted.
* The ternary on the `defaulted` should make it clear that the one case is behaviour in the case of defaulted, the other is when you simply read the value.
* Most important, it doesn't leave the question "why would you conditionally re-assign a value you've already read" open (the "why" I was referring to).
* And my not-so-humble perspective: I definitely don't have to scan the rest of the function for *more* possible reassignments, I know that we're dealing with a value.

joakim-hove: > but I'd like for the ternary to stay

fair enough - I have never liked ternaries; but I realize I am in the minority - guess I just have to suck it up. I'll merge when travis is green.

jokva: Nah, it's not something to just "suck up" to be honest, we should rather figure out something that works for us based on its merits. I (obviously) think the ternary has merit **because** it's an expression, and I have yet to be convinced that the alternative provides any benefits, but I'm open to be persuaded.

joakim-hove: > we should rather figure out something that works for us based on its merits

[meritocracy](http://www.dictionary.com/browse/politics) :-)

PR #881, opened 2016-07-15
Well matching returns empty list when no wells are found; Summary/Completion use Well matching

jokva:
**BEHAVIOURAL CHANGE**

The "soft" change here is that instead of simply accepting well names or * for well identifiers in the Completion family of summary keywords, patterns are accepted. This covers the documented behaviour *and more*, so hopefully everyone are pleased.

The more radical change, however, is  the modification to the well matching algorithm. It **used** to expand and check wild cards (denoted by *) and then include every well matching, returning possibly the empty list. However, when looking up a well by name (a set pattern, if you will) it would *throw* if the well wasn't found. This has been changed to, in the case of an invalid well name, returning an empty list.

This is consistent to the matching behaviour, but some code (in particular WELTARG) relied on the throwing behaviour. This behaviour has been preserved by checking if the list of matched wells is empty or not.

### DIFF
--- a/opm/parser/eclipse/EclipseState/SummaryConfig/tests/SummaryConfigTests.cpp
+++ b/opm/parser/eclipse/EclipseState/SummaryConfig/tests/SummaryConfigTests.cpp
@@ -55,8 +55,9 @@ static DeckPtr createDeck( const std::string& summary ) {
             "/\n"
             "COMPDAT\n"
             "'PRODUCER'   5  5  1  1 'OPEN' 1* -1  0.5  / \n"
-            "'W_1'   3    7    1    3      'OPEN'  1*     32.948      0.311   3047.839  2*         'X'     22.100 / \n"
             "'W_1'   3    7    2    2      'OPEN'  1*          *      0.311   4332.346  2*         'X'     22.123 / \n"
+            "'W_1'   2    2    1    1      /\n"
+            "'WX2'   2    2    1    1      /\n"
             "/\n"
             "SUMMARY\n"
             + summary;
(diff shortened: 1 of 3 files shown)


GitPaean:
If I remember correctly, Eclipse does not accept `*`. It has to be `1*` or `2*`, and so on. 
People can help to confirm about that.

jokva:
For "repeat this N times" you would be right, but in the context of summary config and completion output it's explicitly considered "defaulted well" (aka all wells).

opm/parser/eclipse/EclipseState/SummaryConfig/tests/SummaryConfigTests.cpp
```diff
@@ -55,8 +55,9 @@ static DeckPtr createDeck( const std::string& summary ) {
             "/\n"
             "COMPDAT\n"
             "'PRODUCER'   5  5  1  1 'OPEN' 1* -1  0.5  / \n"
-            "'W_1'   3    7    1    3      'OPEN'  1*     32.948      0.311   3047.839  2*         'X'     22.100 / \n"
             "'W_1'   3    7    2    2      'OPEN'  1*          *      0.311   4332.346  2*         'X'     22.123 / \n"
```
GitPaean: I mean here. I just realized that you deleted the above line while this line is not added by you. I misread it. Sorry about that. 

`*` indicating defaulted values are not supported by Eclipse, while we have so many this kinds of examples in our data decks in the tests.

jokva: Oh right. Not a problem!

Apparently "not supported" means "I'll happily accept this" in eclipse speak, which means we must deal with it as well, unfortunately.

OPMUSER:
Correct.

ThunderBird Signature File

opm/parser/eclipse/EclipseState/SummaryConfig/tests/SummaryConfigTests.cpp
OPMUSER: Correct in Eclipse you have to enter the form n* (1*, 2* etc.) * on its 
own will throw a warning/error.

ThunderBird Signature File

PR #882, opened 2016-07-15
Extracted configuration of *when* to write restart

joakim-hove:
The configuration of when to write restart files is quite complex,
involving at least the three keywords RPTSOL, RPTRST and RPTSCHED. This
commit moves that configuration from the IOConfig class to the
RestartConfig class.

At a later stage the RestartConfgig class will be extended with
information of *what* to save in the restart file.


jokva:
Forgot to add a source file?

joakim-hove:
> Forgot to add a source file?

Ehh - yes.

jokva:
I feel that `RestartConfig` should be a first-class citizen of `EclipseConfig` (or `EclipseState` for that matter) rather than a member of `IOConfig`. Thoughts?

joakim-hove:
> I feel that RestartConfig should be a first-class citizen of EclipseConfig

Yes - I agree; and that is my plan as well. But I wanted to complete the refactoring before affecting downstream, especially now when downstream merges are slow to to come by.

jokva:
If it's on your TODO-list then fine, this is a merge. Thumbs up.

PR #883, opened 2016-07-18
[Summary] W-keywords accept patterns

jokva:
The W-family of keywords accept a pattern to expand, rather than just
names or defaulted-all. This is the actual behaviour, according to the
manual.

### DIFF
--- a/opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp
+++ b/opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp
@@ -109,23 +109,24 @@ inline void keywordW( std::vector< ERT::smspec_node >& list,
                       const Schedule& schedule ) {
 
     const auto type = ECL_SMSPEC_WELL_VAR;
+    static const std::vector< std::string > wildcard = { "*" };
 
-    if( keyword.size() == 0 ||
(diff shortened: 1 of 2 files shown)


andlaus:
actually this was the reason why regexp based keywords were introduced in the first place ;) (and hence the need for `boost::regex`; be aware that the same equally applies to the B- and R- keywords)

jokva:
Maybe, but there's not that much support for B- and R-keywords yet. For some reason the documentation doesn't specify anything w.r.t groups and name matching, but that doesn't necessarily mean that it doesn't still accept patterns.

Regardless, having a regex engine somewhere is probably not a bad idea.

joakim-hove:
> actually this was the reason why regexp based keywords were introduced in the first place

Not that it matters - but I think I disagree; the initial introduction of regular expressions in the `xxxx_PROBE` keywords was to match keywords, whereas the current use of `*` matching (not proper regular expressions) is for matching content.

joakim-hove:
Looks fine - I will force merge this: https://github.com/OPM/opm-simulators/pull/763 tomorrow and restart Travis; assuming Travis I will merge.

jokva:
Splendid.

andlaus:
> Not that it matters - but I think I disagree

yeah, it does not matter, but at least it was _my_ reason for implementing this ;)

PR #884, opened 2016-07-19
WIP: Restart config what

joakim-hove:
Started to build a configuration object of which keywords should be included in restart. This is work in progress and not ready for merge.

### DIFF
--- a/opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.cpp
+++ b/opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.cpp
@@ -24,6 +24,7 @@
 #include <boost/lexical_cast.hpp>
 #include <boost/filesystem.hpp>
 
+#include <opm/parser/eclipse/Utility/Functional.hpp>
 #include <opm/parser/eclipse/Deck/DeckItem.hpp>
 #include <opm/parser/eclipse/Deck/DeckKeyword.hpp>
 #include <opm/parser/eclipse/Deck/DeckRecord.hpp>
@@ -50,10 +51,97 @@ namespace Opm {
                 && std::all_of( x.begin() + 1, x.end(), is_digit );
         }
 
+
+        std::set<std::string> notSupported = {"NORST", "SFREQ" , "STREAM"};
+
+        constexpr const char*  integerKeywords[] = {""     ,      //  1 BASIC=?? - ignored
+                                                    "FLOWS",      //  2
+                                                    "FIP",        //  3
+                                                    "POT",        //  4
+                                                    "PBPD",       //  5
+                                                    "",           //  6 FREQ=?? - ignored
+                                                    "PRES",       //  7
+                                                    "VISC",       //  8
+                                                    "DEN",        //  9
+                                                    "DRAIN",      // 10
+                                                    "KRO",        // 11
+                                                    "KRW",        // 12
+                                                    "KRG",        // 13
+                                                    "PORO",       // 14
+                                                    "NOGRAD",     // 15
+                                                    "NORST",      // 16 NORST - not supported
+                                                    "SAVE",       // 17
+                                                    "SFREQ",      // 18 SFREQ=?? - not supported
+                                                    "ALLPROPS",   // 19
+                                                    "ROCKC",      // 20
+                                                    "SGTRAP",     // 21
+                                                    "",           // 22 - Blank - ignored.
+                                                    "RSSAT",      // 23
+                                                    "RVSAT",      // 24
+                                                    "GIMULT",     // 25
+                                                    "SURFBLK",    // 26
+                                                    "PCOW,PCOG",  // 27 - Multiple, split on "," [Special cased]
+                                                    "STREAM",     // 28 STREAM=?? - not supported
+                                                    "RK",         // 29
+                                                    "VELOCITY",   // 30
+                                                    "COMPRESS" }; // 31
+
+
+        static void insertKeyword(std::set<std::string>& keywords, const std::string& keyword) {
+            const auto pos = keyword.find( "=" );
+            if (pos != std::string::npos) {
+                std::string base_keyword( keyword , 0 , pos );
+                insertKeyword( keywords, base_keyword );
+            } else {
+                /*
+                  We have an explicit list of not supported keywords;
+                  those are keywords requesting special output
+                  behavior which is not supported. However - the list
+                  of keywords is long, and not appearing on this list
+                  does *not* imply that the keyword is actually
+                  supported by the simulator.
+                */
+                if (notSupported.find( keyword ) == notSupported.end())
+                    keywords.insert( keyword );
+            }
+        }
+
     }
 
-    RestartConfig::restart RestartConfig::rptrst( const DeckKeyword& kw, size_t step ) {
 
+    RestartSchedule::RestartSchedule( size_t sched_restart) :
+        rptsched_restart_set( true ),
+        rptsched_restart( sched_restart )
+    {
+    }
+
+    RestartSchedule::RestartSchedule( size_t step, size_t b, size_t freq) :
+        timestep( step ),
+        basic( b ),
+        frequency( freq )
+    {
+    }
+
+    bool RestartSchedule::operator!=(const RestartSchedule & rhs) const {
+        return !( *this == rhs );
+    }
+
+    bool RestartSchedule::operator==( const RestartSchedule& rhs ) const {
+        if( this->rptsched_restart_set ) {
+            return rhs.rptsched_restart_set
+                && this->rptsched_restart == rhs.rptsched_restart;
+        }
+
+        return this->timestep == rhs.timestep &&
+            this->basic == rhs.basic &&
+            this->frequency == rhs.frequency;
+    }
+
+
+
+    void RestartConfig::handleRPTRST( const DeckKeyword& kw, size_t step ) {
+        RestartSchedule rs;
+        RestartSchedule unset;
         const auto& items = kw.getStringData();
 
         /* if any of the values are pure integers we assume this is meant to be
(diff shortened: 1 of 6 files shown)


opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.hpp
```diff
@@ -20,6 +20,8 @@
 #ifndef OPM_RESTART_CONFIG_HPP
 #define OPM_RESTART_CONFIG_HPP
 
+#include <vector>
+
 #include <boost/date_time/gregorian/gregorian_types.hpp>
```
jokva: Seems to me like this boost include can go away too. Happy days!

opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.cpp
```diff
@@ -195,6 +205,118 @@ namespace Opm {
...
 
+    void RestartConfig::update( size_t step, const RestartSchedule& rs) {
+        if (step == 0)
+            this->restart_schedule->updateInitial( rs );
+        else
+            this->restart_schedule->update( step, rs );
+    }
+
+
+    const std::vector<std::string> RestartConfig::integerKeywords = {""     ,      //  1 BASIC=?? - ignored
```
jokva: Why is this a member in the first place?

Unless there's a very good reason for it, this should probably be `static constexpr const char* integerKeywords[] = { ... }.` (or anon namespace as you see fit).

Why? Well, first of all it's a simpler data type for what is static (as in binary region) in the binary anyway, and you use indexing for lookup by design. Second, it opens up a few optimisations for no syntactic cost.

joakim-hove: > Why is this a member in the first place?

Old habits.

> static constexpr const char* integerKeywords[] = { ... }. (or anon namespace as you see fit).

I'll update - guess I'll go for the anon namespace.

This patch has some miles to go anyway.

As a side comment: I would encourage anyone skeptical to the value of tests to refactor this code; I have been guided semicolon by semicolon by the tests!

joakim-hove:
OK -this should be conent-wise ready; did not have time for rebasing.

jokva:
I'll have a look.

opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.cpp
```diff
@@ -24,6 +24,7 @@
 #include <boost/lexical_cast.hpp>
```
jokva: Good opportunity to replace lexical_cast, maybe?

opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.cpp
```diff
@@ -50,10 +51,97 @@ namespace Opm {
...
+                                                    "SURFBLK",    // 26
+                                                    "PCOW,PCOG",  // 27 - Multiple, split on "," [Special cased]
+                                                    "STREAM",     // 28 STREAM=?? - not supported
+                                                    "RK",         // 29
+                                                    "VELOCITY",   // 30
+                                                    "COMPRESS" }; // 31
+
+
+        static void insertKeyword(std::set<std::string>& keywords, const std::string& keyword) {
+            const auto pos = keyword.find( "=" );
```
jokva: There's actually a `basic_string::find( CharT, size_type pos = 0 )` overload. Since we're looking for a character we can use that rather than string-based. Minor, but should be slightly faster.

opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.cpp
```diff
@@ -70,52 +158,65 @@ namespace Opm {
...
-        // is done to make flow write restart file at the same intervals
-        // as eclipse for the Norne data set. There might be some rules
-        // we are missing here.
-        if( 0 == basic ) return {};
+                rs = RestartSchedule( step, basic, freq );
+            }
 
-        if( items.size() > FREQ_index ) // if frequency is set
-            freq = std::stoul( items[ FREQ_index ] );
+            updateKeywords( step , fun::map( []( const std::string& s) { return std::stoi(s); } , items ));
```
jokva: `fun::map( std::stoi, items )`.

opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.hpp
```diff
@@ -20,7 +20,165 @@
...
+
+     RESTART=1 : As RPTRST,BASIC=1
+     RESTART>1 : As RPTRST,BASIC=2
+     RESTART>2 : Flow is added to restart file
+     RESTART>3 : Fluid in place is added to restart file
+     RESTART=6 : Restart file for every timestep.
+
+
+
+  RPTSOL: The RPTSOL keyword is very similar to the RPTCHED keyword,
```
jokva: Typo? I presume you meana RPTSCHED

opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.hpp
```diff
@@ -114,7 +273,36 @@ namespace Opm {
...
+        RestartSchedule() = default;
+        RestartSchedule( size_t sched_restart);
+        RestartSchedule( size_t step, size_t b, size_t freq);
+        bool writeRestartFile( size_t timestep , const TimeMap& timemap) const;
+        bool operator!=(const RestartSchedule& rhs) const;
+        bool operator==( const RestartSchedule& rhs ) const;
+
+
+
+        //private:
```
jokva: Re-privatise the members?

opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.hpp
```diff
@@ -155,55 +347,15 @@ namespace Opm {
...
-
-        std::shared_ptr<DynamicState<restart>> m_restart_output_config;
+        void handleScheduleSection( const SCHEDULESection& schedule);
+        void update( size_t step, const RestartSchedule& rs);
+        void addKeywords( size_t step, const std::vector<std::string>& keywords);
+        void updateKeywords( size_t step, const std::vector<int>& integer_controls);
+        void handleRPTRST( const DeckKeyword&, size_t );
+        static RestartSchedule rptsched( const DeckKeyword& );
+
+        std::shared_ptr<DynamicState<RestartSchedule>> restart_schedule;
```
jokva: Can we kill the `shared_ptr`s here?

jokva:
I guess we should re-enable the old tests as well, no?

opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.cpp
```diff
@@ -50,10 +51,97 @@ namespace Opm {
                 && std::all_of( x.begin() + 1, x.end(), is_digit );
         }
 
+
+        std::set<std::string> notSupported = {"NORST", "SFREQ" , "STREAM"};
```
jokva: Just a general observation here: I don't think this needs to be a set. It's three elements, and it doesn't change at all. We can make it a constexpr array and just use `any_of`, meaning the algorithm for unsupported later becomes:

```
bool unsupported( const std::string& kw ) {
    static constexpr const char* keywords[] = { "NORST", "SFREQ", "STREAM" };   
return std::any_of( std::begin( keywords ), std::end( keywords ), std::equal< std::string >() );
}
[...]

if( !unsupported( keyword ) ) insertKeyword( keyword );
```

Obviously, making it an array or something doesn't change the real benefit here which is providing a function that determines if a keyword is unsupported or not, that can also be accomplished with a string. It does lift the burden of comparison from the insert algorithm.

jokva:
Personally I feel this gets a bit unwieldy now. Little bit too much going on in too many branches.

It was already pretty hard to follow, but the flow was largely linear in the sense that there were no multiple branches, only exit conditions.

If I get the overall idea of the patch right, we want to
1. Register at what timesteps to restart (this was already supported)
2. Register *what* to output at some time step.

which is user configurable through the same two keywords. Assuming this is correct, I want to ask the following:

1. How many keywords control this behaviour? The 26 `integerKeywords` or more? Is it feasible to not use a string set to represent this?
2. Can we split the operations and assembly up a bit? In its current state it feels a bit ad-hoc and case-A-then-maybe-B-unless-C-then-D.

opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.cpp
```diff
@@ -155,19 +256,86 @@ namespace Opm {
...
+      keywords - there are some of them.
+    */
+
+    void RestartConfig::addKeywords( size_t step, const std::vector<std::string>& keywords) {
+        if (keywords.size() > 0) {
+            std::set<std::string> kw_set = this->restart_keywords->back();
+            for (const auto& kw : keywords )
+                insertKeyword( kw_set , kw );
+
+            if (step == 0)
```
jokva: This can be encapsulated in `update`.

jokva:
Furthermore I'm not quite sure I see the value of the add/update keyword interface. Care to shed some light on it?

opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.cpp
```diff
@@ -155,19 +256,86 @@ namespace Opm {
...
+      to remove keywords with the string based menonics.
+    */
+
+    void RestartConfig::updateKeywords( size_t step, const std::vector<int>& integer_controls) {
+        std::set<std::string> keywords = this->restart_keywords->back();
+
+        for (size_t index = 0; index < integer_controls.size(); index++) {
+            if (index == 26) {
+                if (integer_controls[index] > 0) {
+                    insertKeyword( keywords , "PCOW" );
```
jokva: This is confusing. If POCW+PCOG hasn't been added because integer_controls[26] (can be done outside of the loop by the way, clearer that way) <= 0, why would you need to erase them?

opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.cpp
```diff
@@ -184,35 +352,65 @@ namespace Opm {
...
+                                             );
 
-            restart_config.update( current_step, rs );
+                update( current_step , rs );
+            }
         }
+    }
+
+    bool RestartSchedule::writeRestartFile( size_t input_timestep , const TimeMap& timemap) const {
+        if (this->rptsched_restart_set && (this->rptsched_restart > 0))
```
jokva: Nice.

opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.cpp
joakim-hove: I thought so to - did not work at first attempt; will try again.

opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.hpp
joakim-hove: Yes.

opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.hpp
joakim-hove: Yes - that was the plan; forgot it.

opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.cpp
joakim-hove: >This can be encapsulated in update.

You mean the `DynamicState::update()`?

jokva: No, `RestartConfig::update`

joakim-hove:
OK @jokva - it's all yours!

Observe that there are two failing tests - these test test that the `RESTART=n` mnenonic on the `RPTSOL` and `RPTSCHED` keywords can configure that the `FIP` and `FLOW` keywords should be added to list of keywords. That is only implemented in the test - in true TDD style.

jokva:
Great. Since the tests are failing I'll use this branch as a starting point rather than merge into master and then go ham, and my final PR will be the one schedule for merge.

joakim-hove:
> and my final PR will be the one schedule for merge.

Yes - that was my plan; do merge your own future PR when you are happy with it.

jokva:
Will do, great work!

pgdr:
Am I to understand that we can close this PR and that #887 is the one that eventually will be merged?

jokva:
Yes.

PR #885, opened 2016-07-19
Unit system unit name

jokva:
Extends unit system support by adding three new units, all ratios, which are non-scaling. The more major feature is the ability to look up a string representation of the unit (M, M3, KG, FT, BARSA etc) from a `measure`.

The need is largely driven from opm-output.

### DIFF
--- a/opm/parser/eclipse/Units/UnitSystem.cpp
+++ b/opm/parser/eclipse/Units/UnitSystem.cpp
@@ -77,6 +77,29 @@ namespace {
         Metric::ReservoirVolume / Metric::Time,
         Metric::Transmissibility,
         Metric::Mass,
+        1, /* gas-oil ratio */
+        1, /* oil-gas ratio */
+        1, /* water cut */
+    };
+
+    static constexpr const char* metric_names[] = {
+        "",
+        "M",
+        "DAY",
+        "KG/M3",
+        "BARSA",
+        "K",
+        "C",
+        "CP",
+        "MD",
+        "SM3",
+        "SM3",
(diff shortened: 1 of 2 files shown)


opm/parser/eclipse/Units/UnitSystem.cpp
```diff
@@ -57,6 +57,9 @@ namespace {
         1 / ( Metric::ReservoirVolume / Metric::Time ),
         1 / Metric::Transmissibility,
         1 / Metric::Mass,
+        1, /* gas-oil ratio */
+        1, /* oil-gas ratio */
```
joakim-hove: Do we need both `gas-oil` and `oil-gas`?

jokva: Not necessarily, from a quick glance I can only find gas-oil ratio, gas-liquid ratio (uses same unit as gas-oil) and liquid-liquid ratio (water cut). I did choose to include it for some completeness, since oil-gas has a dedicated entry in the manual.

OPMUSER: Yes, oil-gas is for a gas condensate field/reservoir condensate yield
normally called CGR

Regards

David Baxendale
david.baxendale@gmail.com

OPMUSER: Yes, oil-gas is for a gas condensate field/reservoir condensate yield
normally called CGR

Regards

David Baxendale
david.baxendale@gmail.com

PR #887, opened 2016-07-26
RestartConfig - generalised implementation, support for options

jokva:
Based work by @joakim-hove (https://github.com/OPM/opm-parser/pull/884).

Synopsis of the interesting changes:
* IOConfig has been split in two - RestartConfig and IOConfig - IOConfig governs all paths etc., RestartConfig governs what and when to write restart files.
* Integer list input is first translated to their corresponding
  mnemonics.
* Mnemonic implementation support turning off input parameters via
  mnemonic=0, handled generally rather than special-cased for BASIC/FREQ
* Restart configuration setup is handled in one else-less, linear code
  path. RPTRST and RPTSCHED are interpreted by their separate helper
  functions, but they produce the same resulting object. Should be
  easier to follow
* Ignores unknown mnemonics, has a list of known/supported mnemonics
  that will be internalised.


opm/parser/eclipse/EclipseState/Schedule/DynamicState.hpp
```diff
@@ -68,6 +68,17 @@ namespace Opm {
...
 
 
+        const T& back() const {
+            if (m_data.size() > 0)
+                return m_data.back();
+            else
+                return m_initialValue;
+        }
+
+        T& back() {
```
joakim-hove: Is this necessary. I am slighly uncomfortable to let this be a mutable return, since modifying the initialvalue semantically feels different from modifying an arbitrary last value.

Maybe it could throw if size == 0?

joakim-hove:
Apart from the DynamicState I am happy to see this merged, but a proxy/shim could be retained temporarily to keep downstream happy?

opm/parser/eclipse/EclipseState/Schedule/DynamicState.hpp
jokva: It's actually not needed anymore at all, I think it can be removed. Old iteration's patch that got through, it seems. I'll try to remove it.

jokva:
Nah, better to update downstream I think. It can wait till you're back regardless (enjoy your holiday!), but I think we should use this opportunity to redesign the `RestartConfig` (and `IOConfig`) interface to not use the `override` methods, but rather a `copy-then-modify-local` paradigm. Since this is a breaking change anyway we may as well do it right. Nothing is about to use these new features anyway, so we have some time.

That being said, there are a couple of bugs I found with @flikka that I'm fixing now.

jokva:
DynamicState mutability gone.

opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.hpp
```diff
@@ -114,13 +281,42 @@ namespace Opm {
...
+            RestartSchedule() = default;
+            RestartSchedule( size_t sched_restart);
+            RestartSchedule( size_t step, size_t b, size_t freq);
+            bool writeRestartFile( size_t timestep , const TimeMap& timemap) const;
+            bool operator!=(const RestartSchedule& rhs) const;
+            bool operator==( const RestartSchedule& rhs ) const;
+
+
+
+        //private:
```
pgdr: uncomment?

pgdr:
This looks like a very good addition and I would be more than happy to see this merged soon.

Where are we on the compile issues reported by Travis?

opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.hpp
jokva: Yes, eventually. The experimental implementation was easier this way, and the full interface needs some work, which will be done once @joakim-hove returns.

In short, `RestartConfig` should be "overridable" by applications, i.e. for simulations the preferred way of overriding deck specified configurations should be to provide your own `RestartConfig` instance, rather than the ugly `override*` calls we use now. Since that will break this interface anyway, re-privatising these vars can be done in the same sweep.

jokva:
Except for merge conflicts the compilation problems are just interface breakage (and not behaviour), so this is easily fixable by just replacing a few calls. Regardless, I'd like to review the entire override-approach as mentioned and that will also update downstream.

I've tested a locally-modified `flow` with this patch applied and it works as expected, once you take the new interface into account.

joakim-hove:
It is not clear to me exactly what is the state of this patch; but if it is reasonably compatible with the current state I would wish for:

1. The patch is updated with shims/proxies so that downstream builds.
2. We merge.
3. We start a new development removing the `overrideXXX( )` methods and changing downstream accordingly.

I would prefer this approach - to guarantee that the changes hitting downstream are as small as possible. Although it would of course be possible to directly do the `overrideXXX( )` deprecation in this patch we will then get a quite unfocused patch hitting several repositories - that will in my opinion increase the chance of f...up.

jokva:
Then I'll clean up to accomplish this. We don't need shims really though, so we're good like that.

joakim-hove:
> Then I'll clean up to accomplish this

Very good!

PR #890, opened 2016-08-08
Removing some use of shared_ptr

pgdr:
Remove returning of shared_ptr.  The opm-parser and opm-output are moving towards a shared_ptr-free state (except from the cases where we actually use shared ownership)

* TransMult, InitConfig, SimulationConfig are now returned as references, and not as shared_ptr's.
* The method `applyModifierDeck` takes a reference, not shared_ptr

This PR has downstream dependencies:
* OPM/opm-output#73
* OPM/opm-core#1056
* OPM/opm-simulators#783
* OPM/ewoms#63

### DIFF
--- a/opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
+++ b/opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
@@ -212,7 +212,7 @@ namespace Opm {
         m_output_enabled = enabled;
     }
 
-    std::string IOConfig::getOutputDir() {
+    std::string IOConfig::getOutputDir() const {
         return m_output_dir;
     }
 
--- a/opm/parser/eclipse/Applications/opmi.cpp
+++ b/opm/parser/eclipse/Applications/opmi.cpp
@@ -51,7 +51,7 @@ inline void loadDeck( const char * deck_file) {
     std::cout << "Loading deck: " << deck_file << " ..... "; std::cout.flush();
     deck = parser->parseFile(deck_file, parseContext);
     std::cout << "parse complete - creating EclipseState .... ";  std::cout.flush();
-    state = std::make_shared<Opm::EclipseState>( deck , parseContext );
+    state = std::make_shared<Opm::EclipseState>( *deck , parseContext );
     std::cout << "complete." << std::endl;
 
     dumpMessages( deck->getMessageContainer() );
--- a/opm/parser/eclipse/EclipseState/EclipseConfig.cpp
+++ b/opm/parser/eclipse/EclipseState/EclipseConfig.cpp
@@ -39,8 +39,8 @@ namespace Opm {
                                  const Schedule& schedule,
                                  const ParseContext& parseContext) :
             m_ioConfig(        std::make_shared<IOConfig>(deck)),
-            m_initConfig(      std::make_shared<const InitConfig>(deck)),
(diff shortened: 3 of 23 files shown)


pgdr:
jenkins build this opm-output=73 opm-core=1056 opm-simulators=783 please

jokva:
Thumbs up.

One question though, as I feel our interfaces in general are rather clunky. Does it make much sense to keep on using names such as `getXXXXConfig`? I mean, they're already part of `EclipseConfig` as straight-up members more-or-less directly exposed, and already provide their logic in isolation. Seems very hungarian notation-esque tacked on redundant information that in my not-so-humble opinion becomes rather noisy.

```
EclipseState state;
auto& summary = state.cfg.summary;
// or
auto& summary = state.cfg().summary();
// or even
auto& summary = state.summary_config()
```

rather than:

```
auto& summary = state.getEclipseConfig().getSummaryConfig();
auto& summary = state.getSummaryConfig();
```

I'm rambling. This is not for this patch obviously (which looks good by the way, but I'll have a second read though that's more thorough).

pgdr:
That's a very good idea.

I'm for the `auto& summary = state.cfg().summary()` approach.  And I think even I can squeeze it into this PR, it shouldn't be too much work.

jokva:
More about it being a different scope and goal rather than the work.

Regardless,  the two names can live side-by-side for a while, since it would be a pure addition, and then later we can look to deprecate old interfaces.

pgdr:
jenkins build this opm-output=73 opm-core=1056 opm-simulators=783 please

pgdr:
I added the proposed shortform of the API.  Now we can write
```c++
const auto& ioConfig   = state.cfg().io();
const auto& simulation = state.cfg().simulation();
const auto& summary    = state.cfg().summary();
const auto& initConf   = state.cfg().init();
```

I also made `string getOutputDir()` into `string getOutputDir() const`

opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp
```diff
@@ -212,7 +212,7 @@ namespace Opm {
         m_output_enabled = enabled;
     }
 
-    std::string IOConfig::getOutputDir() {
+    std::string IOConfig::getOutputDir() const {
```
jokva: Can this be a reference as well? Not sure why it's a copy.

jokva:
Thumbs up from me.

Whenever downstream passes and all downstream maintainers are happy this is good to go.

joakim-hove:
jenkins build this opm-output=73 opm-core=1056 opm-simulators=783 with downstreams please

pgdr:
jenkins build this opm-output=73 opm-core=1056 opm-simulators=783 ewoms=63 with downstreams please

atgeirr:
The downstream PRs in opm-core and opm-simulators look fine to me, feel free to merge when ready.

I have one question though: after this, do any of the "public" classes (EclipseState, Deck etc.) have value semantics (i.e. copy constructor will make a deep copy)? For now I assume that no classes have value semantics.

jokva:
They're about to (some remnants left that don't), but that's definitely the goal.

jokva:
Oh, and since your question was more precise than my vague answer. From my recollection, `Deck` has value semantics already, but `EclipseState` does not. Most of the objects that build `EclipseState` are default deep copyable, but the aggregate is not (yet).

PR #891, opened 2016-08-08
Optimisations

jokva:
A series of micro optimisations which reduces execution time by some 20%.

Nothing should change for users except a slightly lower power bill. For developers, highlights are:
* `SCHEDULESection` is no longer special; the file has been killed and `DeckTimeStep` is gone.
* `is_separator` and `is_quote` use lookup tables and are function objects for inlining when used with `algorithm`
* `RawRecord` expansion no longer uses an explicit list, instead registers the same `string_view` multiple times.


joakim-hove:
Cool!

PR #892, opened 2016-08-09
Shut wells where all completions are shut

totto82:
Shut the well if all its completions are shut. 
The check is applied in handleCOMPDAT(...). 
The same logic is already applied when completions are SHUT using WELOPEN.


joakim-hove:
Thank you for writing test!

PR #893, opened 2016-08-15
Quell some compiler warnings

akva2:
This contains two commits which were done to target unused variable compiler warnings.

- One commit which moves unit-conversion constant generation to compile time. Constexprs can be unused and safely pruned from the binary by the optimizer. There is equivalent code in opm-core for this, i did not target that one as i'm not entirely sure what the deal with the duplication is about.
- a trivial removal of some unsed references in some tests.

### DIFF
--- a/opm/parser/eclipse/EclipseState/tests/EclipseStateTests.cpp
+++ b/opm/parser/eclipse/EclipseState/tests/EclipseStateTests.cpp
@@ -497,7 +497,6 @@ BOOST_AUTO_TEST_CASE(TestIOConfigCreation) {
     DeckPtr deck = parser->parseString(deckData, ParseContext()) ;
     EclipseState state(*deck , ParseContext());
 
-    const IOConfig& ioConfig = state.cfg().io();
     const RestartConfig& rstConfig = state.cfg().restart();
 
     BOOST_CHECK_EQUAL(false, rstConfig.getWriteRestartFile(0));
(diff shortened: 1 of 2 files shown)


joakim-hove:
> There is equivalent code in opm-core for this, [...]  i'm not entirely sure what the deal with the duplication is about.

This code started in opm-core, also opm-parser needs access to unit constants and due to dependencies opm-parser can not use the constants defined in opm-parser. At the time opm-parser was an optional dependency, and opm-core kept it's constants around, but now I guess it would make sense to let opm-core use the constants from opm-parser, or move them all to opm-common?

atgeirr:
They will be moved to opm-common, once I get around to making another pruning round in opm-core. This fix will be good to have there of course as well.

jokva:
Is there a good reason to move this to common? Is there a problem with it living in parser?

atgeirr:
I'm not suggesting to remove it from opm-parser, since opm-parser will want to be independent. For other clients of the code however, it makes sense to have it in opm-common rather than opm-core.

jokva:
So it's not motivated by the duplication at all? I suppose moving it to common would make it accessible from non-core dependent modules that also don't necessarily require parser (that's material, no? maybe ewoms).

Seems to me like another nail in core's coffin.

atgeirr:
> Seems to me like another nail in core's coffin.

That's the idea, yes.

PR #894, opened 2016-08-19
Do not use constexpr pair constructor.

atgeirr:
That constructor is not available in C++11, only from C++14 on.

(Causes compile error with -std=c++11 for me.)

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.cpp
@@ -95,7 +95,7 @@ namespace Opm {
 
         namespace wp = WellProducer;
         using mode = std::pair< const char*, wp::ControlModeEnum >;
-        static constexpr mode modes[] = {
+        static const mode modes[] = {
             { "ORAT", wp::ORAT }, { "WRAT", wp::WRAT }, { "GRAT", wp::GRAT },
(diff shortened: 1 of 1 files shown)


joakim-hove:
>That constructor is not available in C++11, only from C++14 on.

No problem for me to merge - but I though we were targeting the gcc4.8  - and not necessarily strict C++11?

jokva:
Splendid, will merge as soon as test runners are green.

jokva:
> No problem for me to merge - but I though we were targeting the gcc4.8 - and not necessarily strict C++11?

I suppose we consider clang best effort, but regardless this isn't all that important I think and not worth breaking compiler compatibility for.

atgeirr:
> I though we were targeting the gcc4.8 - and not necessarily strict C++11?

It can be that gcc 4.8 has this constructor in its library, but on my compiler (clang), it's only available when passing -std=c++14 and not with -std=c++11, and I don't think we are (yet) at the point where we want to require c++14. If gcc accepts it with -std=c++11 that is strictly speaking a (tiny) gcc extension.

atgeirr:
By the way, what is the appveyor thing, and who has set that up?

joakim-hove:
> By the way, what is the appveyor thing, and who has set that up?

Windows CI - like Travis. Set up by @jokva

atgeirr:
> Windows CI - like Travis. Set up by @jokva

👍

PR #895, opened 2016-08-19
Restart keywords

joakim-hove:


opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.cpp
```diff
@@ -360,6 +360,24 @@ RPTSCHED( const DeckKeyword& keyword ) {
 }
 
 
+    void RestartConfig::addRestartKeywords( size_t timeStep , const std::map<std::string, int>& input_mnemonics) {
```
jokva: Ok, so I'm getting my nitpicking magnifying glass.

nitpick

This approach can be made *far* more elegant by acknowledging that we're doing a copy anyway.

```
add( size_t ts, map< std::string, int > mnemonics ) {
    auto allprops = mnemonics.find( "ALLPROPS" );
    if( allprops == mnemonics.end() ) return this->restart_keywords.update( ts, mnemonics );
    auto value = allprops->second;
    mnemonics.erase( allprops );
    for( const auto& kw : { "BG", "BO", "..." } )
        mnemonics[ kw ] = value;
    this->restart_keywords.update( ts, mnemonics );
}
```

joakim-hove: Hehe - as you say MMV; personally I maintain a good-treatment-of-datastructures ordering which looks like this:

1. Immutable - best.
2. Add elements - middle.
3. Remove elements - worst.

Anyway - I'll look into the possibility of a hybrid approach.

jokva: Externally it's immutable (which is what really matters, honestly), and this requires¹ mutation anyway. You're effectively doing remove-then-add anyway, it's just via conditional addition, preventing whole-structure copying.

Besides, what happens when ALLPROPS is set to 0, and BG is set to 1 later? Which one kicks in?

¹ Not really, but in C++ we can safely operate under this assumption.

joakim-hove: OK; code updated according to suggestion.

PR #896, opened 2016-08-19
add bare-bones support for the MULTREGP keyword

andlaus:
it is parsed and put into the deck, but it is not internalized in any
way. (Note that, given that this keyword is quite simple, adding
sophisticated code to absorb it into EclipseState would be overkill
in my not so humble opinion.)

### DIFF
--- a/opm/parser/share/keywords/000_Eclipse100/M/MULTREGP
+++ b/opm/parser/share/keywords/000_Eclipse100/M/MULTREGP
@@ -1,0 +1,8 @@
+{"name": "MULTREGP",
+ "sections": ["GRID", "EDIT"],
+ "items": [
+    {"name": "REGION", "value_type": "INT"},
+    {"name": "MULTIPLIER", "value_type": "DOUBLE"},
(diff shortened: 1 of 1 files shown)


andlaus:
> What is the use case?

I've been told that it is used by Model 2. The opm-simulators part is still to do, but it should be pretty easy as well.

OPMUSER:
Why not add the MULTREGT keword as well as it multiplies the 
transmissibility between flux or MULTNUM regions.

Engineers use this to isolate the different regions or fault blocks y 
setting the value to zero by:

-- 
-- SET TRANSMISSIBILITES ACROSS DIFFERENT RESERVOIRS TO ZERO TO ISOLATE 
RESERVOIRS
-- 
--    REGION    REGION      TRANS    DIREC NNC
--    FROM      TO          MULT     OPT OPTS
MULTREGT
       1*        1*          0.0      1*      'ALL'   / ALL REGIONS SEALED
/

ThunderBird Signature File

andlaus:
> Why not add the MULTREGT keword as well

because it already is there and has full support by `flow` (this has been the case for about two years, IIRC.)

PR #899, opened 2016-08-23
add multiconfiguration support to jenkins build script

akva2:
also unifies build.sh and build-pr.sh and gets rid of build-opm-parser.sh.

Depends on https://github.com/OPM/opm-common/pull/169 and changes to the jenkins configuration at http://ci.opm-project.org


pgdr:
Is there a reason why `build.sh` isn't put in opm-common, and then all we have to do in parser is to "git clone common" && `opm-common/jenkins/build.sh`?  I think that's what we do with Travis, and that works quite well.

akva2:
yes. it works because travis is "stupid", it doesn't keep track of the forks in the dependency trees. i do not want to make that simplification, so i need to keep the list of up / downstreams in each module.

akva2:
@joakim-hove due to this being sensitive to the order of changes and needing syncronized external changes to the jenkins server, i request authority to merge myself when deemed ready.

akva2:
jenkins build this serial please

akva2:
jenkins build this serial opm-common=169 please

akva2:
this is ready for merge. i have made an attempt at jenkins to keep the builder working in the mean time, but no guarantees how well those attempts work.

joakim-hove:
> due to this being sensitive to the order of changes and needing syncronized external changes to the jenkins server, i request authority to merge myself when deemed ready.

Sorry - I lost this one. *Yes 007* - you have the permission to merge!

PR #900, opened 2016-08-25
EclipseGrid::activeMap( ) will cache the results.

joakim-hove:


jokva:
We save ourselves some trouble and remove a source-of-errors by not heap-alloc'ing the cache.

I'll send you the proposal as a patch.

joakim-hove:
Ahhh - sorry; merged a bit prematurely.

PR #901, opened 2016-08-25
Active cell cache with automatic storage

jokva:
Simplify slightly by using automatic memory management, rather than
heap-alloc.


joakim-hove:
Thank you - the plan was to merge this as part of #900. Anyway - going in when Travis is green.

PR #902, opened 2016-08-25
Scan a property and return cells equal to a value.

joakim-hove:
Convenience functionality for region calculations.


jokva:
Press the button whenever green.

joakim-hove:
> Press the button whenever green.

Thanks - pushing now.

PR #903, opened 2016-08-26
Replace number parser with boost::spirit::qi

jokva:
The hand-written number parser functions implemented using strtod and
friends were rather slow (profiling indicates that typically 30% of the
program is spent inside of strtod internals). By using
boost::spirit::qi, which we already depend on through boost-filesystem
and others this portion typically seem to be reduced to 20% (via
instruction count) and with somewhat better cache performance.
Rudimentary measuring indicates ~15% speedup overall.

Additionally, the intention is a lot clearer this way, so readability
received a boost. Compilation time of StarToken goes through the roof.

### DIFF
--- a/opm/parser/eclipse/RawDeck/StarToken.cpp
+++ b/opm/parser/eclipse/RawDeck/StarToken.cpp
@@ -24,10 +24,12 @@
 #include <stdexcept>
 #include <cstdlib>
 
+#include <boost/spirit/include/qi.hpp>
+
 #include <opm/parser/eclipse/RawDeck/StarToken.hpp>
 #include <opm/parser/eclipse/Utility/Stringview.hpp>
 
-
+namespace qi = boost::spirit::qi;
 
 namespace Opm {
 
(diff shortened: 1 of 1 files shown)


andlaus:
I agree that the resulting code is nicer to read. A few question come to the mind nevertheless:

- I'm stupid; what is `boost::spirit`? is it some obscure corner of boost or should this be known?
- isn't the goal to reduce the use of boost (if possible) with the prospect of getting rid of it entirely in the long term?
- for which use-cases is the parser still a performance bottleneck? (for the simulators it has been "fast enough" for a while IMO, so do you eye visualization?)

opm/parser/eclipse/RawDeck/StarToken.cpp
```diff
@@ -24,10 +24,12 @@
 #include <stdexcept>
 #include <cstdlib>
 
+#include <boost/spirit/include/qi.hpp>
+
 #include <opm/parser/eclipse/RawDeck/StarToken.hpp>
 #include <opm/parser/eclipse/Utility/Stringview.hpp>
 
-
+namespace qi = boost::spirit::qi;
```
andlaus: this is unfortunate IMO: I'd prefer if you replaced calls to `qi::` below by the full thing because -- while adding some additional verbosity -- it takes away the mental burden of remembering that `qi::` is actually `boost::spirit::qi`

jokva:
@andlaus boost::spirit is their parser (generator) library. I wouldn't call it obscure per se, it receives love and has been a part of boost since... 1.44? The newest addition, x3 is apparently also even faster than qi.

It's not *the goal* per se, and I don't think it'll happen before we bump to a C++17 compiler anyway. And since we depend on it for filesystem and regex anyway we might as well have faster number parsing.

My problem with boost is more it being of varying quality (not all components are that good) and the sheer size of it, also some features are slow. This is a fast one though.

It's  never fast enough I suppose, but largely for visualisation and other applications than simulation. Flow is so slow that it doesn't matter for simulation.

alfbr:
> for which use-cases is the parser still a performance bottleneck? (for the simulators it has been "fast enough" for a while IMO, so do you eye visualization?)

Yes, visualization is interactive, so here every second counts. With opm-parser now in use within Resinsight we are seeing slow-downs.

joakim-hove:
I'll merge this. Just a final check: The `boost:spirit` is *header-only* - or is there an extra library somewhere - which has just been linked in by magic???

opm/parser/eclipse/RawDeck/StarToken.cpp
```diff
@@ -69,55 +71,38 @@ namespace Opm {
 
     template<>
     int readValueToken< int >( string_view view ) {
+        int n = 0;
```
joakim-hove: I really don't understand this code - to me it seems the code should return `n= 0`; or raise an exception. I would have expected to pass `&n`?

jokva:
It's header only, yes.

opm/parser/eclipse/RawDeck/StarToken.cpp
jokva: It's just to silence the maybe-uninitialised warning.

joakim-hove: Sure - the main point was that as an old C hand I was expecting a modifiable scalar argument to be passed as `&n`; but I guess the signature of the parse routine is `(....,  T& value )`.

jokva: More-or-less.

```
    template <typename Iterator, typename Expr>
    inline bool 
    parse(
        Iterator& first
      , Iterator last 
      , Expr const& expr)
```

PR #905, opened 2016-08-29
WIP: Added EclipseGrid processed copy.

joakim-hove:
Will create an EclipseGrid copy where zcorn and zctnum have been adjusted.

Depends on: https://github.com/Ensembles/ert/pull/1217

### DIFF
--- a/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
+++ b/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
@@ -727,7 +727,45 @@ namespace Opm {
         this->getActiveMap();
     }
 
+    ZcornMapper EclipseGrid::zcornMapper() const {
+        return ZcornMapper( getNX() , getNY(), getNZ() );
+    }
 
+    ZcornMapper::ZcornMapper(size_t nx , size_t ny, size_t nz)
+        : dims( {nx,ny,nz} ),
+          stride( {2 , 4*nx, 8*nx*ny} ),
+          cell_shift( {0 , 1 , 2*nx , 2*nx + 1 , 4*nx*ny , 4*nx*ny + 1, 4*nx*ny + 2*nx , 4*nx*ny + 2*nx + 1 })
+    {
+    }
+
+
+    /* lower layer:   upper layer  (higher value of z - i.e. lower down in resrvoir).
+
+         2---3           6---7
+         |   |           |   |
+         0---1           4---5
+    */
+
+    size_t ZcornMapper::index(size_t i, size_t j, size_t k, int c) const {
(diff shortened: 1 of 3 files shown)


opm/parser/eclipse/EclipseState/Grid/EclipseGrid.hpp
```diff
@@ -55,6 +55,7 @@ namespace Opm {
     public:
         explicit EclipseGrid(const std::string& filename);
         explicit EclipseGrid(const EclipseGrid& srcGrid);
+        explicit EclipseGrid(const EclipseGrid& src, const std::vector<double>& zcorn , const std::vector<int>& actnum);
```
jokva: Not that it matters, but `explicit` does nothing in non-1-arity constructors, since the possibility for implicit conversion just isn't there due to the arity.

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
```diff
@@ -118,6 +118,20 @@ namespace Opm {
...
+    EclipseGrid::EclipseGrid(const EclipseGrid& src, const std::vector<double>& zcorn , const std::vector<int>& actnum)
+        : GridDims(src.getNX(), src.getNY(), src.getNZ()),
+          m_messages( src.m_messages ),
+          m_minpvValue( src.m_minpvValue ),
+          m_minpvMode( src.m_minpvMode ),
+          m_pinch( src.m_pinch ),
+          m_pinchoutMode( src.m_pinchoutMode ),
+          m_multzMode( src.m_multzMode )
+    {
+        m_grid.reset( ecl_grid_alloc_processed_copy( src.c_ptr(), zcorn.data() , actnum.data() ));
```
jokva: You can pass this to the `m_grid` constructor in the init list as well.

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.hpp
joakim-hove: > but explicit does nothing in non-1-arity constructors

Copy & Paste; best technique there is.

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
joakim-hove: > You can pass this to the m_grid constructor in the init list as well.

Good point.

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.hpp
jokva: Is that why they call it "clue code".

I'll see myself out.

PR #907, opened 2016-09-01
UnitSystem: Add skeleton implementation of LAB conventions

bska:
This commit adds conversion factors for the ECL LAB unit conventions--notably Atmospheres for pressures, Hours for time, Grams for mass, and cubic centimetres for volumes.

I am not sure if the unit names in [`lab_names`](https://github.com/OPM/opm-parser/blob/ba69320b528d20aabeff0a5f061495a3eb962428/opm/parser/eclipse/Units/UnitSystem.cpp#L226-L246) match those of ECL, but this is my best suggestions at present.

On a second note, if I may be so bold as to suggest an interface change, I'd propose that the `UnitSystem::new*()` constructor methods at some point be modified to return either `std::unique_ptr<>`s or even objects by value.

### DIFF
--- a/opm/parser/eclipse/Units/ConversionFactors.hpp
+++ b/opm/parser/eclipse/Units/ConversionFactors.hpp
@@ -112,6 +112,7 @@ namespace Opm {
             /// \name Mass
             /// @{
             constexpr const double kilogram = 1;
+            constexpr const double gram     = 1.0e-3 * kilogram;
             // http://en.wikipedia.org/wiki/Pound_(mass)#Avoirdupois_pound
             constexpr const double pound    = 0.45359237 * kilogram;
             /// @}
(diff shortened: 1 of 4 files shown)


jokva:
I think `UnitSystem` should return by value (in fact I've been wanting it for a while but who has the time). Personally I'd see the entire old `UnitSystem` parsing go away in favour of table lookup, but that requires some more discussion as it's somewhat breaking.

joakim-hove:
> This commit adds conversion factors for the ECL LAB

Cool going in when all is green.

>On a second note, if I may be so bold as to suggest an interface change, I'd propose that the UnitSystem::new*() constructor methods at some point be modified to return either std::unique_ptr<>s or even objects by value.

Nothing is sacred - of course you can suggest! Maybe we will make a change in that direction. Is your suggestion based on a general style consideration - or do you face a concrete problem with the current implementation? Anyway - I am personally going from one end of the spectrum to the other; from an excessive (over)use of `std::shared_ptr<T>` to trying to reduce the amount of smart pointers. At some stage before I retire I might have discovered a good balance.

Anyway - my current guiding principle on smart pointers is that they should only be implementation details; an API should be based on the fundamental types(?) - i.e. values, references and naked pointers.

jokva:
> Anyway - my current guiding principle on smart pointers is that they should only be implementation details; an API should be based on the fundamental types(?) - i.e. values, references and naked pointers.

A case for return-by-value, isn't it?

bska:
> my current guiding principle on smart pointers is that they should only be implementation details; an API should be based on the fundamental types(?) - i.e. values, references and naked pointers.

I agree completely, and to be honest I'd prefer return-by-value in this particular case.  My suggestion of `unique_ptr<>` was primarily based on it being a relatively smaller syntactical change at call sites--usually at moste a `std::move()` into a `shared_ptr<>` with remaining statements remaining unchanged.

On a general note, I think a case can be made for returning `unique_ptr<>` from some factory-style functions and containers of `unique_ptr<>`s are useful datastructures within the implementation of a class--especially if the objects pointed to are somewhat large.

joakim-hove:
> A case for return-by-value, isn't it?

I guess so; I am just not sufficiently familiar with the call sites.

jokva:
> I guess so; I am just not sufficiently familiar with the call sites.

Neither am I (although I suspect it's contained to when the state is handed to the simulator and when output takes it back), in which case it won't be that bad to track down. It's a task for another day though.

jokva:
Only one error in unit names as far as I can tell, "hour" is "hr", not "h". Do you want to submit a patch, @bska, or do you want me to do it for you?

bska:
> Do you want to submit a patch, @bska, or do you want me to do it for you?

If you have the time to do it, please go ahead.  I'm otherwise occupied at the moment.

jokva:
Not a problem.

PR #909, opened 2016-09-01
Implement support for MULTREGP

andlaus:
This PR implements full support for the MULTREGP keyword. It tries to faithfully implement the documented ECLIPSE bugs w.r.t. this keyword, but lacking the real thing, I cannot check if MULTREGP is considered when removing cells due to MINPV. (I assume that it is considered, but somebody should check to make sure.)

Finally, for this to be consistent within `flow`, the simulator must retrieve the pore volume of the active cells from opm-parser (i.e., it requires a small patch for opm-simulators).

### DIFF
--- a/opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
+++ b/opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
@@ -33,7 +33,6 @@
 #include <opm/parser/eclipse/EclipseState/Tables/TableManager.hpp>
 #include <opm/parser/eclipse/Utility/String.hpp>
 
-
 namespace Opm {
 
     namespace {
@@ -50,9 +49,26 @@ namespace Opm {
             }
         }
 
-        /// initPORV uses doubleGridProperties: PORV, PORO, NTG, MULTPV
+        // this funcion applies a single pore volume multiplier (i.e., it deals with a single
+        // record of the MULTREGP keyword)
+        void applyPorosityRegionMultiplier_(std::vector<double>& porev,
+                                            const std::vector<int>& regionId,
+                                            int multRegionId,
+                                            double multValue)
+        {
+            for (unsigned i = 0; i < porev.size(); ++i) {
+                if (regionId[i] == multRegionId)
+                    porev[i] *= multValue;
+            }
+        }
+
+        /// this function initializes the pore volume of all cells. it uses the raw keyword
+        /// 'MULTREGP', the integer grid properties 'FLUXNUM', 'MULTNUM' and 'OPERNUM' as
+        /// well as the double grid properties 'PORV', 'PORO', 'NTG' and 'MULTPV'
         void initPORV( std::vector<double>&    values,
+                       const Deck* deck,
                        const EclipseGrid*      eclipseGrid,
+                       const GridProperties<int>* intGridProperties,
                        const GridProperties<double>* doubleGridProperties)
         {
             const auto iter = std::find_if( values.begin(), values.end(), []( double value ) { return std::isnan(value); });
@@ -80,6 +96,54 @@ namespace Opm {
                 const auto& multpvData = multpvKeyword.getData();
                 for (size_t globalIndex = 0; globalIndex < multpvData.size(); globalIndex++)
                     values[globalIndex] *= multpvData[globalIndex];
+            }
+
+            // deal with the region multiplier for porosity
+            if (deck->hasKeyword("MULTREGP")) {
+                const DeckKeyword& multregpKeyword = deck->getKeyword("MULTREGP");
+                for (unsigned recordIdx = 0; recordIdx < multregpKeyword.size(); ++recordIdx) {
+                    const DeckRecord& multregpRecord = multregpKeyword.getRecord(recordIdx);
+
+                    int regionId = multregpRecord.getItem("REGION").template get<int>(0);
+                    std::string regionType = multregpRecord.getItem("REGION_TYPE").template get<std::string>(0);
+                    double multValue = multregpRecord.getItem("MULTIPLIER").template get<double>(0);
+                    uppercase(regionType, regionType);
+
+                    // deal with the "convenience" feature of ECL: if the region index is
+                    // zero or negative, the record is ignored.
+                    if (regionId <= 0)
+                        continue;
+
+                    // implement a (documented) ECL bug: the region index must be unique
+                    // (i.e., it is impossible to specify multipliers for different
+                    // region types with the same index). Also, only the last occurence
+                    // of each region counts.
+                    unsigned record2Idx = recordIdx + 1;
+                    for (; record2Idx < multregpKeyword.size(); ++record2Idx) {
+                        const DeckRecord& multregpRecord2 = multregpKeyword.getRecord(record2Idx);
+                        int region2Idx = multregpRecord2.getItem("REGION").template get<int>(0);
+                        if (region2Idx == regionId)
+                            break;
+                    }
+                    if (record2Idx < multregpKeyword.size())
+                        // the region was specified twice
+                        continue;
+
+                    if (regionType == "M") {
+                        const GridProperty<int>& multnumProp = intGridProperties->getKeyword("MULTNUM");
+                        applyPorosityRegionMultiplier_(values, multnumProp.getData(), regionId, multValue);
+                    }
+                    else if (regionType == "F") {
+                        const GridProperty<int>& fluxnumProp = intGridProperties->getKeyword("FLUXNUM");
+                        applyPorosityRegionMultiplier_(values, fluxnumProp.getData(), regionId, multValue);
+                    }
+                    else if (regionType == "O") {
+                        const GridProperty<int>& opernumProp = intGridProperties->getKeyword("OPERNUM");
+                        applyPorosityRegionMultiplier_(values, opernumProp.getData(), regionId, multValue);
+                    }
+                    else
+                        throw std::logic_error("Unknown or illegal region type for MULTREGP keyword: '"+regionType+"'");
+                }
             }
         }
 
@@ -396,7 +460,9 @@ namespace Opm {
         {
             auto initPORVProcessor =  std::bind(&initPORV,
                                       std::placeholders::_1,
+                                      &deck,
                                       &eclipseGrid,
+                                      &m_intGridProperties,
                                       &m_doubleGridProperties);
 
             m_doubleGridProperties.postAddKeyword( "PORV",
--- a/opm/parser/eclipse/EclipseState/Grid/tests/PORVTests.cpp
+++ b/opm/parser/eclipse/EclipseState/Grid/tests/PORVTests.cpp
@@ -228,6 +228,39 @@ static Opm::DeckPtr createDeckWithNTG() {
     return parser->parseString(deckData, Opm::ParseContext()) ;
 }
 
+static Opm::DeckPtr createDeckWithMULTREGP() {
+    const char *deckData =
+        "RUNSPEC\n"
+        "\n"
+        "DIMENS\n"
+        " 10 10 10 /\n"
+        "GRID\n"
+        "DX\n"
+        "1000*0.25 /\n"
+        "DYV\n"
+        "10*0.25 /\n"
+        "DZ\n"
+        "1000*0.25 /\n"
+        "TOPS\n"
+        "100*0.25 /\n"
+        "PORV\n"
+        "1000*77 /\n"
+        "MULTNUM\n"
+        "100*1 700*2 200*3 / \n"
+        "FLUXNUM\n"
+        "200*1 700*2 100*3  / \n"
+        "MULTREGP\n"
+        "1 10.0 / \n"
+        "1 2.0 F/ \n"
+        "0 123.0 F/ \n" // ignored
+        "2 20.0 M / \n"
+        "/\n"
+        "\n";
+
+    Opm::ParserPtr parser(new Opm::Parser());
+    return parser->parseString(deckData, Opm::ParseContext()) ;
+}
+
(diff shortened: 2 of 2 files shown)


opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -81,6 +99,57 @@ namespace Opm {
...
+            // deal with the region multiplier for porosity
+            if (deck->hasKeyword("MULTREGP")) {
+                const DeckKeyword& multregpKeyword = deck->getKeyword("MULTREGP");
+                for (int recordIdx = 0; recordIdx < multregpKeyword.size(); ++recordIdx) {
+                    const DeckRecord& multregpRecord = multregpKeyword.getRecord(recordIdx);
+
+                    int regionIdx = multregpRecord.getItem("REGION").template get<int>(0);
+                    std::string regionType = multregpRecord.getItem("REGION_TYPE").template get<std::string>(0);
+                    double multValue = multregpRecord.getItem("MULTIPLIER").template get<double>(0);
+                    boost::to_upper(regionType);
```
jokva: Is `regionType` an eclipse keyword, or a name? I assume the former, in which case it must be whitespace-sanitised as well, right?

atgeirr:
> the simulator must retrieve the pore volume of the active cells from opm-parser

I believe we already do that by default? GeoProps.hpp lines 117-128 or so.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
andlaus: it represents the value of the third item of the current MULTREGP record and is supposed to be a user-supplied string. I've got no idea what Eclipse does here (I'm not even sure if it is case-insensitive) -- could you check?

jokva: The regions are "M", "F" and "O" as you check for, but I suspect that it still would swallow something like `'M     '` and be happy.

Regardless, that's not something we need to worry about until the problem shows up I think.

andlaus: > I suspect that it still would swallow something like 'M ' and be happy.

that's quite possible, and I would not mind implementing it. But before I do this, could you check what the reference implementation does? (you know, that expensive Fortran-based thingy which does not even have its source publicly available.) If you do, could you please also check if it accepts ' M', 'MULTNUM' and 'MFOO'?

joakim-hove: You could check the MULTREGT code, I am quite certain there already is code for this minor detail there.

andlaus: also, it would be nice to check if it swallows lower case letters ('m'). the patch suggested here does, but I don't want to bet a beer that ECL does so too...

andlaus: > You could check the MULTREGT code, I am quite certain there already is code for this minor detail there.

sure: for MULTREGT, the current implementation does not seem to do anything:

https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp#L475
and https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Grid/MULTREGTScanner.cpp#L36

which corresponds to the documented behaviour. I'm just wondering whether that's what the reference implementation does, because there are other cases that are quite similar IMO where the documentation deviates from its implementation. (lower-case keywords anyone?)

I suppose you're right that if MULTREGP needs to be fixed, the MULTREGT code probably needs changes, too.

joakim-hove: > I suppose you're right that if MULTREGP needs to be fixed, the MULTREGT code probably needs changes, too.

That was my main point; so my suggestion is that this patch is refactored slightly to ensure that both `MULTREGT`and `MULTREGT` share the same code. The current `MULTREGT` implementation has at least worked for ~2 years.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -81,6 +99,57 @@ namespace Opm {
...
+
+                    // deal with the "convenience" feature of ECL: if the region index is
+                    // zero or negative, the record is ignored. (why are such lines not
```
joakim-hove: While I fully agree with your sigh here I think it should be removed from the code.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -81,6 +99,57 @@ namespace Opm {
...
+                    // simply commented out? so that mere mortals do not understand
+                    // what's going on?)
+                    if (regionIdx <= 0)
+                        continue;
+
+                    // implement a (documented) ECL bug: the region index must be unique
```
joakim-hove: Again - the `MULTREGT` implementation implements the same quirks (I think) - would be good if the implementation could be shared.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -81,6 +99,57 @@ namespace Opm {
...
+                    // (i.e., it is impossible to specify multipliers for different
+                    // region types with the same index). Also, only the last occurence
+                    // of each region counts. Once again, this shows the very high
```
joakim-hove: Although I certainly agree with your slightly opinionated comments I would prefer if they were removed from the code.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
andlaus: > refactored slightly to ensure that both MULTREGTand MULTREGT share the same code.

I suppose that this is a cure worse than the disease: the added complexity needed for sharing that trivial piece of code would be much larger than simply duplicating it as far as I can see.

> The current MULTREGT implementation has at least worked for ~2 years.

sure. my point was rather: it worked for all decks which we tested so far, but not necessarily for all decks floating around in the wild  ;)

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
andlaus: ok. changed.

joakim-hove: Thank you!

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
jokva: It's actually not that hard for us to implement "worst case" support:

```
kw = uppercase( kw );
switch( kw.front() ) {
    case 'M': /* update */ break;
    case 'F': /* update */ break;
    [...]
}
```

For the record I'm not suggesting this is a good idea.

jokva: On that note, I see you've used `boost` for uppercasing the string. It's obviously possible, but it's not necessarily all that fast because it deals with encodings, and we offer this https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/Utility/String.hpp#L10

And yes; it's surely a NIH, but it was really the only thing we used from `boost::algorithm::string`, and it improved compilations time quite a lot.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
andlaus: for this particular issue, the code can hardly be shared because MULTREGT has two regions here,  while MULTREGP only has one. I haven't looked at MULTREGT too much, though, so I'm not even sure wheter the (sourceRegion, targetRegion) pairs must be unique. (Also, technically it would obviously be possible to do this using the same code path, but as far as I can see the abstraction layer required to hide the difference between a pair of integers and a single integer would be much larger than the code itself, so it would IMO be rather like an "obfuscation layer".)

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
andlaus: Its only optionated if you can see the sarcasm. anyway, its removed.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
andlaus: > It's actually not that hard for us to implement "worst case" support

yeah, true, but do we want to eat more than eclipse does? (i.e., is it okay if decks get accepted by opm-parser but not by Eclipse?)

> we offer this https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/Utility/String.hpp#L10

that's news to me: it was not around too long ago. anyway, it's changed.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
joakim-hove: > so it would IMO be rather like an "obfuscation layer".

OK - fair enough.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
jokva: > that's news to me: it was not around too long ago. anyway, it's changed.

Yeah, it's rather recent. It was added during a boost purge where most of these (really only uppercase) boost dependencies were removed, in an effort to improve performance and reduce redundant work.

> yeah, true, but do we want to eat more than eclipse does? (i.e., is it okay if decks get accepted by opm-parser but not by Eclipse?)

I suppose, and it's not the first time we're doing that I think, but it provides us with another dilemma when people start using undocumented accidental behaviour and we want to change something.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -50,9 +49,26 @@ namespace Opm {
             }
         }
 
-        /// initPORV uses doubleGridProperties: PORV, PORO, NTG, MULTPV
+        // this funcion applies a single pore volume multiplier (i.e., it deals with a single
+        // record of the MULTREGP keyword)
+        void applyPorosityRegionMultiplier_(std::vector<double>& porev,
+                                            const std::vector<int>& regionIdx,
```
joakim-hove: Minor nitpick: Can we have e.g. `regionValue` (or `regionId`) instead of `regionIdx`? The point is that the values in the region field are not necessarily zero offset - and can also have "holes" in the sequence of values.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -81,6 +97,54 @@ namespace Opm {
                 for (size_t globalIndex = 0; globalIndex < multpvData.size(); globalIndex++)
                     values[globalIndex] *= multpvData[globalIndex];
             }
+
+            // deal with the region multiplier for porosity
+            if (deck->hasKeyword("MULTREGP")) {
+                const DeckKeyword& multregpKeyword = deck->getKeyword("MULTREGP");
+                for (int recordIdx = 0; recordIdx < multregpKeyword.size(); ++recordIdx) {
```
joakim-hove: I think (and hope ....) you can now do:
```C++
for (const auto& multregpRecord : multregpKeyword) { 
    ...
}
```

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -81,6 +97,54 @@ namespace Opm {
...
+                    const DeckRecord& multregpRecord = multregpKeyword.getRecord(recordIdx);
+
+                    int regionIdx = multregpRecord.getItem("REGION").template get<int>(0);
+                    std::string regionType = multregpRecord.getItem("REGION_TYPE").template get<std::string>(0);
+                    double multValue = multregpRecord.getItem("MULTIPLIER").template get<double>(0);
```
joakim-hove: Excuse my C++ ignorance; but what is the `.template get<T>` thing?

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
jokva: The index is used for look-at-next, unfortunately.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
jokva: Dependent name lookup.

> 

    After name lookup (3.4) finds that a name is a template-name, if this name is followed by a <, the < is always taken as the beginning of a template-argument-list and never as a name followed by the less-than operator.

Shouldn't be needed here though.

joakim-hove:
OK - barring very minor comments I am ready to merge this.

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
andlaus: changed to `regionId`. nice spot!

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
andlaus: that's because in some instances the compiler can't figure out if this is meant to be a call to a template method or `get` is a public attribute, i.e. whether the call should be interpreted as

```c++
((multregpRecord.get < T) >)()
```

(which would be a syntax error, but I believe there are corner cases where it would result in valid code.) this happens if the type of the `multregpObject` object is a template argument.

> Shouldn't be needed here though.

I prefer to leave it like it is because it will work even if the type of the `multregpRecord` is made a template...

PR #910, opened 2016-09-02
Do not apply the MAPAXES transform.

joakim-hove:
With this PR we will not at all apply the MAPEXES transform when creating the EclipseGrid / simulation grid. However - if a MAPAXES is specified in the input deck that will be retained, and written out to disk when the EGRID file is saved again.

Depends on: https://github.com/Ensembles/ert/pull/1221

Unless something pops up I intend to merge this - and https://github.com/Ensembles/ert/pull/1221 together on monday.


joakim-hove:
jenkins build this ert=1221 please

joakim-hove:
Merging this now; observe that ert must be updated to contain: https://github.com/Ensembles/ert/pull/1221

PR #912, opened 2016-09-05
Added some tests to verify ALLPROPS handling.

joakim-hove:


jokva:
Plus one and merge when green.

joakim-hove:
> Plus one and merge when green.

Thanks - merging.

PR #914, opened 2016-09-06
Added units required for formation volume factors

babrodtk:
This adds the missing formation volume factors to UnitSystem used for outputting for bw, bo, bg.


opm/parser/eclipse/Units/UnitSystem.cpp
```diff
@@ -129,6 +138,9 @@ namespace {
         1, /* gas-oil ratio */
```
atgeirr: While not a part of this PR, is not this wrong (also oil-gas ratio and the reverse relations in from_field[]): it should be GasSurfaceVolume/LiquidSurfaceVolume (and vice versa in the right spots), that is MSCF/STB.

jokva: MSCF/STB is the field unit (metric is sm³/sm³). However, this is the unit transformation **scaling**. At least for metric, since the computation is done in SI (and ratios shouldn't be affected by units), there's no scaling to be done. But since you bring it up, I don't know if we have to scale for field. Lab is scc/scc, and should require no scaling.

atgeirr: > if we have to scale for field

We do, simply since the numerator and denominator of the unit are different and therefore different from 1. This is a good example of a "fun" case where the *dimension* is 1 but the *unit* is not (I did not believe such weirdness would ever happen until I saw eclipse decks with field units...)

But I wonder about this: does not the conversion of input depend on these factors as well? And judging from being able to run the SPE cases well we do that conversion correctly.

jokva: Are there many ratios in inputs? I'd assume not, but regardless I think input is mostly handled by the string-based lookup which hasn't been changed at all.

atgeirr: > Are there many ratios in inputs?

Formation volume factors are input in PVDO, PVDG, PVTO, PVTG, PVTW etc. (pvt data).

jokva: Fair enough - they might be buggy for field (but should be fine for lab and metric).

atgeirr:
Can this be merged? Downstream PRs depend on it, so it would be good to get in.

PR #915, opened 2016-09-07
Added EclipseGrid move constructor.

joakim-hove:


jokva:
You don't move messages and friends, is that intentional? Except for that it looks fine obviously, but it makes me thing maybe the m_grid itself should learn how to be move/copyable, making all these copy/move constructors propagating obsolete.

PR #916, opened 2016-09-08
Remove Grid from Schedule object + Python wrap Schedule and EclipseGrid

joakim-hove:
1. Cleaned up and simplified the `Schedule <-> EclipseGrid` interaction.
2. Added *very thin* Pyhon wrappers for `EclipseGrid` and `Schedule`.


opm/parser/eclipse/EclipseState/Schedule/Completion.cpp
```diff
@@ -179,7 +182,8 @@ namespace Opm {
         const WellCompletion::DirectionEnum direction = WellCompletion::DirectionEnumFromString(compdatRecord.getItem("DIR").getTrimmedString(0));
 
         for (int k = K1; k <= K2; k++) {
-            CompletionPtr completion(new Completion(I , J , k , state , connectionTransmissibilityFactor, diameter, skinFactor, direction ));
+            double depth = grid.getCellDepth( I,J,k );
+            CompletionPtr completion(new Completion(I , J , k , depth , state , connectionTransmissibilityFactor, diameter, skinFactor, direction ));
```
jokva: TODO: remove this vector-of-pointers.

(not for this task, just making a note of it).

joakim-hove:
Unless there are objections I will self-merge this tomorrow.

jokva:
No objections from me.

Well, except I think we should look into boost.python rather than using cwrap for this. Seems to save us a lot of boilerplate, is only a compile time dependency, native C++-support and should be fewer cross platform issues.

joakim-hove:
> Well, except I think we should look into boost.python rather than using cwrap for this.

Fair point - starting from scratch that might have been the best choice.

PR #918, opened 2016-09-09
Return identity when compressing active-only cells.

jokva:
See https://github.com/OPM/opm-output/issues/88


joakim-hove:
Thanks for fixing this - please merge ås soon it is Green.

jokva:
Affirmative.

joakim-hove:
Merging #920 instead - where the comment has a different wording.

PR #921, opened 2016-09-13
Get print and stop limits from MESSAGES keyword.

qilicun:
This PR enables the user codes to get print and stop limits from MESSAGES keyword in parser.


joakim-hove:
**Time dependence**: The `MESSAGES` keyword should have time-dependence, i.e. it should be possible to reset it in the Schedule section. This means that the messages behavior should be internalized in the `Schedule`object with a `DynamicState<T>` and not in the `EclipseState`. I guess the best solution would be to add a:
```
DynamicState<Messages> messages;
```
to the `Schedule` class - alternatively the `Messages` class could *contain* several `DynamicState<int>` instances.

**Data structure**: I don't particularly like the `std::vector<int>` as a way to access the limit values. My suggestion would actually be to just use a plain `struct` with public members, alternatively a full class with various `getXXXStopLimit()` and `getXXXPrintLimit()` methods. The default constructor of the `Messages` class/struct should initialize all members with the default values:
```
warning_stop_limit = ParserKeywords::MESSAGES::WARNING_STOP_LIMIT::defaultValue;
...
```

qilicun:
>Time dependence: The MESSAGES keyword should have time-dependence, i.e. it should be possible to reset it in the Schedule section. 

I did not aware of that, thanks for notice!

>getXXXStopLimit() and getXXXPrintLimit() methods. 

That's fine. But the intention of ``std::vector`` is that I thought if the custom codes need  to access to these limits, the most possible case is that they want all the values at one time.

joakim-hove:
> I thought if the custom codes need to access to these limits, the most possible case is that they want all the values at one time.

That might be right - but then calling scope needs to check the internal ordering in this vector, and it can not be changed without breaking. That is in my opinion not a good design.

qilicun:
>but then calling scope needs to check the internal ordering in this vector

That's the drawback, Ok then we go to the separate methods.

joakim-hove:
Just a comment on names: I guess the class should be called `MessageLimits` - i.e. *limits* should be part of the name.

opm/parser/eclipse/EclipseState/EclipseState.hpp
```diff
@@ -53,6 +54,7 @@ namespace Opm {
     class SimulationConfig;
     class TableManager;
     class UnitSystem;
+    class Messages;
```
pgdr: This can be removed.

qilicun:
Sorry for waiting for this so long, just back from vacation. 
It seems the new changes have some conflicts, will solve this ASAP.

pgdr:
@qilicun Welcome back!

Try also to reduce (preferably completely remove) the use of shared pointers, e.g.
```c++
std::shared_ptr<DynamicState<int>> m_message_print_limit;
...
```

Just use
```c++
DynamicState<int> m_message_print_limit;
...
```

The use of shared pointers is an _antipattern_ that we are trying to remove completely from opm-parser.

qilicun:
>@qilicun Welcome back!

:smile: 

>The use of shared pointers is an antipattern that we are trying to remove completely from opm-parser.

Done.

opm/parser/eclipse/EclipseState/Schedule/MessagesLimits.cpp
```diff
@@ -0,0 +1,172 @@
...
+*/
+
+#include <opm/parser/eclipse/Deck/DeckItem.hpp>
+#include <opm/parser/eclipse/Parser/ParserKeywords/M.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/MessagesLimits.hpp>
+
+namespace Opm {
+
+
+    MessagesLimits::MessagesLimits(TimeMapConstPtr timemap):
```
jokva: `TimeMapConstPtr` is gone, use a const ref.

opm/parser/eclipse/EclipseState/Schedule/MessagesLimits.cpp
```diff
@@ -0,0 +1,172 @@
...
+        m_message_print_limit(DynamicState<int>(timemap, ParserKeywords::MESSAGES::MESSAGE_PRINT_LIMIT::defaultValue)),
+        m_comment_print_limit(DynamicState<int>(timemap, ParserKeywords::MESSAGES::COMMENT_PRINT_LIMIT::defaultValue)),
```
jokva: There's no reason to construct a temporary here, just pass the arguments directly to the constructor.

opm/parser/eclipse/EclipseState/Schedule/MessagesLimits.hpp
```diff
@@ -0,0 +1,83 @@
...
+    class MessagesLimits {
+    public:
+
+        /*
+           This constructor will create a new Messages object which is
+           a copy of the input argument, and then all items explicitly
+           set in the record are modified.
+        */
+
+        MessagesLimits(std::shared_ptr<const TimeMap> timemap);
```
pgdr: Please use const ref, and then you must include TimeMap as well.

(Ps., add WIP prefix to title if it's not ready for review.)

opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp
```diff
@@ -47,6 +48,7 @@ namespace Opm
     class TimeMap;
     class UnitSystem;
     class Well;
+    class MessagesLimits;
```
pgdr: This can be removed

opm/parser/eclipse/EclipseState/Schedule/MessagesLimits.cpp
joakim-hove: The class - and source files - should be named `MessageLimits` - the double plural looks funny.

opm/parser/eclipse/EclipseState/Schedule/MessagesLimits.hpp
jokva: Technically not if he doesn't use it outside of a reference. A forward declaration is plenty.

PR #922, opened 2016-09-14
quell signed/unsigned comparison warning

akva2:
the simpler fix would be to make ```i``` size_t, but i figured it may as well use the stl function.


bska:
> i figured it may as well use the stl function.

Strictly speaking you need
```C++
#include <numeric>
```
in order to use `std::iota()`.

PR #925, opened 2016-09-21
Add method to return number of open wells

totto82:


joakim-hove:
See: https://github.com/OPM/opm-output/pull/93

PR #927, opened 2016-09-23
A well group can be injection and production group at the same time

GitPaean:


opm/parser/eclipse/EclipseState/Schedule/Group.hpp
```diff
@@ -114,6 +115,7 @@ namespace Opm {
         std::shared_ptr<GroupProduction::ProductionData> m_production;
         std::shared_ptr<DynamicState<std::shared_ptr< const WellSet >> > m_wells;
         std::shared_ptr<DynamicState<bool> > m_isProductionGroup;
+        std::shared_ptr<DynamicState<bool> > m_isInjectionGroup;
```
pgdr: This is perhaps a stupid question, but does this really have to be a `shared_ptr`?  Or could this (and `m_isProductionGroup`) be a `unique_ptr` or (better yet) a pure value?

opm/parser/eclipse/EclipseState/Schedule/Group.hpp
```diff
@@ -51,6 +51,7 @@ namespace Opm {
         bool isProductionGroup(size_t timeStep) const;
         bool isInjectionGroup(size_t timeStep) const;
         void setProductionGroup(size_t timeStep, bool isProductionGroup);
+        void setInjectionGroup(size_t timeStep, bool isInjectionGroup_);
```
joakim-hove: The fact that a group can be both an injection group and a production group at the same time is so special that even I would like a comment😀.

opm/parser/eclipse/EclipseState/Schedule/Group.cpp
```diff
@@ -97,7 +97,8 @@ namespace Opm {
         m_injection( new GroupInjection::InjectionData(timeMap) ),
         m_production( new GroupProduction::ProductionData( timeMap )),
         m_wells( new DynamicState< std::shared_ptr< const WellSet > >( timeMap , std::make_shared< const WellSet >() ) ),
-        m_isProductionGroup( new DynamicState<bool>(timeMap, true))
+        m_isProductionGroup( new DynamicState<bool>(timeMap, false)),
+        m_isInjectionGroup( new DynamicState<bool>(timeMap, false))
```
joakim-hove: When both are initialised with false we are in a situation where the well is created as a ??? well? That might be OK, but it seems a bit weird!?

opm/parser/eclipse/EclipseState/Schedule/Group.hpp
joakim-hove: I don't really know - but I guess imposing value semantics on the `DynamicState` instances is a large job. Before we do an all-in conversion I guess the shared_ptr can remain here.

joakim-hove (commented):
OK - with a minor comment, and possibly a discussion of the initial state, this will be merged.

opm/parser/eclipse/EclipseState/Schedule/Group.hpp
jokva: It's not, it's already used as value elsewhere.

opm/parser/eclipse/EclipseState/Schedule/Group.hpp
GitPaean: In my understanding, a group that contains injection wells and production wells at the same time provides some mechanism to let injection wells and production wells interact with each other, such like making the injection rate for some phase dependent on the production rate.

opm/parser/eclipse/EclipseState/Schedule/Group.hpp
GitPaean: > (better yet) a pure value?

maybe not. I think it changes during the simulation.

joakim-hove:
OK; as seen from the parser I am happy to merge this. But - is the simulator ready for this? Is this a needed feature?

GitPaean:
> OK; as seen from the parser I am happy to merge this. But - is the simulator ready for this? Is this a needed feature?

Yes. It is a needed feature for sure. The simulator side is not ready, while we have been using this PR and make some progress in the correct direction. 

And from all the aspects, this PR makes things more correct (No final solution and understanding yet).

PR #928, opened 2016-09-27
Made production group and injection group non-share_ptr

pgdr:


jokva:
Forcing a rebuild on appveyor.

PR #929, opened 2016-09-28
Python test: locate testdata + add declarations.

joakim-hove:


pgdr:
Very good!  This should get rid of most of the compiler warnings for opm-parser!

Will merge very soon.

pgdr:
OPM warnings dropped from 51 to 10:

opm-warns

PR #930, opened 2016-10-05
Remove shared ptrs from schedule

jokva:
Removes most shared_ptr members from the Schedule class. Some other changes are mandated by this:

* `DynamicState::get` returns a *reference*, not a value copy. This has implications for `std::vector<bool>` which can't have its `at` returned by reference since it would be a temporary. All `DynamicState<bool>` are instead represented by `int`, and appropriate bool conversions are added.
* `OilVaporizationProperties` has its default constructor public, but the default constructed version will be `UNDEF` and comparing to this will always be `false`.
* GroupTree has a copy constructor that maps to `GroupTree::deepCopy`.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/DynamicState.hpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/DynamicState.hpp
@@ -91,14 +91,8 @@ namespace Opm {
         }
 
 
-        T get(size_t index) const {
-            if (index >= m_timeMap->size())
-                throw std::range_error("Index value is out range.");
-
-            if (index >= m_data.size())
-                return m_currentValue;
-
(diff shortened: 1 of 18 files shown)


pgdr:
jenkins build this with opm-output=102 opm-core=1082 opm-simulators=852 please

pgdr:
Yes!  Looks great.  I've reviewed it and have no further questions.

I see no reasons why this shouldn't be merged immediately, but I'll wait to see if anyone chimes in.

I'll merge this (together with the PRs in output, core, and simulators) later today.

pgdr:
jenkins build this with downstreams opm-output=102 opm-core=1082 opm-simulators=852 please

PR #931, opened 2016-10-05
Added vector overload of unit onversion methods.

joakim-hove:


jokva:
Good idea, but can't we just generalise it right away to work with iterators and transform?

joakim-hove:
>but can't we just generalise it right away to work with iterators 

I guess we could do that - something like:

```C++
from_si( measure m, std::output_iterator begin, std::iterator end) { ... }
```
but I also guess the `std::vector<double>` will be the only used type for the foreseeable future, so maybe not so much point?

> and transform?

I guess the transformation will *always*[1] be multiply with scalar, so not so much point in generalizing  that?


[1]: That is not actually true, the temperature transformation between Celcius and Kelvin actually involves an addition and a shift. That is ugly I do NOT suggest we look into the required generalizations now.

jokva:
Or even `template< typename InputIt, typename OutputIt > OutputIt from_si( measure m, InputIt begin, InputIt end, OutputIt out ) {}`.

... but it still "requires" a single-type container (so maps are out). Alternatively, we can do:

`template< typename T > T& from_si( measure m, T& )` or (in my opinion better) `T from_si( measure, T )`. We can also handle the kelvin/celcius shift, it's a rather trivial special case.

Anywho, the current implementation suffers from bind1st and bind2nd being deprecated in C++11 (and fully removed in C++17). It's probably better to just use a lambda.

joakim-hove:
OK - `std::bind` changed to use lambda. Tried to implement the methods as template methods with universal reference arguments but then resolving the correct method failed, so the original simple in place update approach is retained.

jokva:
Merge when green.

PR #933, opened 2016-10-06
Add JFUNC

andlaus:
this adds the basic support for the JFUNC keyword and this required to parse the deck of "model 2.2" without hickups. the support is "basic" because the keyword is not internalized in any way. (given that the keyword is pretty simple and not present in most decks, I don't think adding it to `EclipseState` is worth the hassle.)

### DIFF
--- a/opm/parser/eclipse/Units/ConversionFactors.hpp
+++ b/opm/parser/eclipse/Units/ConversionFactors.hpp
@@ -133,6 +133,7 @@ namespace Opm {
             /// \name Force
             /// @{
             constexpr const double Newton = kilogram*meter / square(second); // == 1
+            constexpr const double dyne   = 1e-5*Newton;
             constexpr const double lbf    = pound * gravity; // Pound-force
             /// @}
 
@@ -211,6 +212,7 @@ namespace Opm {
         constexpr const double Salinity             = kilogram/cubic(meter);
         constexpr const double Viscosity            = centi*Poise;
         constexpr const double Timestep             = day;
+        constexpr const double SurfaceTension       = dyne/(centi*meter);
     }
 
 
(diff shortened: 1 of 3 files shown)


pgdr:
It seems good, but I would really like to see a test somewhere just verifying that the parser don't choke on the keyword if it is properly used.

opm/parser/eclipse/Units/ConversionFactors.hpp
```diff
@@ -133,6 +133,7 @@ namespace Opm {
             /// \name Force
             /// @{
             constexpr const double Newton = kilogram*meter / square(second); // == 1
+            constexpr const double dyn    = 1e-5*Newton; // cf. wikipedia
```
andlaus: well, it's just the hard, cold truth. but if you prefer, I'll remove the comment.

andlaus: I just noticed this:

https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/Units/ConversionFactors.hpp#L116

do you still want to change the comment? both?

andlaus:
thanks for your review.

> see a test somewhere just verifying that the parser don't choke on the keyword if it is properly used.

doesn't that statement apply for > 70% of all opm-parser keywords? in particular the ones for the SUMMARY section are grossly undertested IIRC...

opm/parser/eclipse/Units/ConversionFactors.hpp
pgdr: Yes, just remove it.

jokva:
That's not true, most summary keywords have plenty of testing love. Was true before the big summary job started this spring.

opm/parser/eclipse/Units/ConversionFactors.hpp
andlaus: ok. done.

andlaus:
> That's not true, most summary keywords have plenty of testing love. Was true before the big summary job started this spring.

ok, I wasn't aware of that...

PR #935, opened 2016-10-10
update debian packaging

akva2:
add packaging of binaries (opmi)

Sorry for the multiple prs but I need to organize my work and they make it easier.


jokva:
Not a problem. Merge when you deem it ready.

PR #936, opened 2016-10-10
fully move the units code from opm-core to opm-parser

andlaus:
this fixes some annoying inconsistencies (e.g., the recently
introduced 'dyne' unit was unavailable in the opm-core version) and
gets rid of some compiler ambiguity errors if 'using namespace
Opm::details' was used by code inside the 'Opm' namespace or when
both headers were included. (I haven't checked if this somehow worked 
anyway, but I suppose it did not.)

Since opm-parser is a hard dependency of opm-core, the only measure
which must be taken by higher-level code is to include
'opm/parser/eclipse/Units/Units.hpp' instead of
'opm/parser/eclipse/Units/ConversionFactors.hpp' or
'opm/core/utility/Units.hpp'.

Note that a potentially better location for this code would be
opm-common, but this would break the Windows build of opm-parser.

Also be aware that this PR must be merged synchronously with the follow-up PRs for the other modules


andlaus:
jenkins build this please

atgeirr:
Actually, I wanted to move the units from opm-core to opm-common!

This is because other modules (initially the opm-flowdiagnostic-applications module) needs this, and the move would enable that module to not depend on opm-core (or opm-parser for that matter).

Don't you agree it should rather be there?

In my opinion opm-parser is the "odd man out" here, requiring it to be completely independent is fine, but the additional burden should be put on the module that is "special" (opm-parser), in summary I think the canonical unit stuff should be in opm-common, and the opm-parser version should keep up with it if necessary.

andlaus:
> Actually, I wanted to move the units from opm-core to opm-common!

I fully agree that this would be the logically best place to be. the reason I do not propose it (for now) is that opm-parser cannot use opm-common on windows? Though I'd be happy to be informed that this is outdated information and I'll update the PRs as soon as possible...

>This is because other modules (initially the opm-flowdiagnostic-applications module) needs this, and the move would enable that module to not depend on opm-core (or opm-parser for that matter).

hm. sounds like all choices are bad ones. the best way would probably be to make opm-common available to opm-parser on windows, but I've got no idea how much effort that is... 

> Don't you agree it should rather be there?

certainly, but my opinion is not too relevant here IMO.

> In my opinion opm-parser is the "odd man out" here, requiring it to be completely independent is fine, but the additional burden should be put on the module that is "special" (opm-parser), in summary I think the canonical unit stuff should be in opm-common, and the opm-parser version should keep up with it if necessary.

I agree. Would it make things easier to make opm-common a "header only" module?

joakim-hove:
> Don't you agree it should rather be there?

I fully agree that the technically best&trade; solution would be to have all unit related code in opm-common. But when that solution is not available to us we must agree on a second best: as I see it there is no clear second best - and importantly the second best is way behind the best.

> in summary I think the canonical unit stuff should be in opm-common, and the opm-parser version should keep up with it if necessary.

I guess that argument has some merit - but I must admit the solution makes me cringe. The current situation with duplication between opm-parser and opm-core is a legacy, that is in a way fair enough - but to actively introduce such a duplication with open eyes .... 

When it comes purely to flow I would say that @andlaus suggestion makes a certain sense. In flow units should only be used in input, i.e. opm-parser and output i.e. opm-output. But of course - I can see that this does not sit particularly well with e.g. the new flow-diagnostics module.

The technical reasoning that we do not want to move these units to opm-common is that opm-parser should be fully self-contained. It is not 100% clear to me what this implies, and different views/understandings have been expressed in the past. The two views I am aware of include:

1. Opm-parser should be stand-alone for as-simple-as-possible use in entirely different projects.
2. Opm-parser should work on windows. Opm-parser should be stand-alone to ensure that the burden of keeping the windows build functioning is minimized.

But - we still require the build system from opm-common, so in my opinion the gain in opm-parser independence by banning code dependencies between opm-parser and opm-common is limited.

andlaus:
@magnesj: since you seem to be responsible for the windows port, could you please comment on the issue?

atgeirr:
The required independence of opm-parser is at this point a given I think, and I do not propose to change that. Needing units in other modules makes me want to put units in opm-common (rather than opm-core), but that does not have to touch opm-parser at all, since it has its own version which I have thought of as internal to opm-parser. This PR makes that version public in some sense, I guess that is what makes me uneasy. I'm sure we'll figure out a good solution, though!

joakim-hove:
>[....], since it has its own version which I have thought of as internal to opm-parser

Yes - it has its own version, but that is certainly an unfortunate legacy detail. There is no such thing as units internal to opm-parser. I must say say I feel more and more reluctant to change the current situation - right now it is slightly broken, and everyone knows - changing it to something which only &epsilon; less broken is in my opinion of dubious value.

andlaus:
@atgeirr 
> The required independence of opm-parser is at this point a given I think, and I do not propose to change that. Needing units in other modules makes me want to put units in opm-common (rather than opm-core), but that does not have to touch opm-parser at all, since it has its own version which I have thought of as internal to opm-parser.

My motivation for coming up with this patch in the fist place was that when there are two copies of the file, they *will* get out of sync sooner or later: e.g., I recently added the "dyne" unit for force to the opm-parser version, but besides the parser it is also needed in the user code to implement the JFUNC keyword correctly. (if there's a better alternative than converting all quantities to the ECL units and applying the equation outlined in the documentation, let me know.)

So to keep everyone's sanity I'd like to have only a single canonical place for this functionality. (I don't really care where it is, though.)

@joakim-hove 
> right now it is slightly broken, and everyone knows

"everyone" did not include me until recently, so I had to learn it the hard way ;)

andlaus:
about the issues of dependencies: how is ERT dependency handled on windows?

alfbr:
Actually opm-parser was intended to be stand-alone from day one, it had little to do with windows. Supporting window properly does come with a cost, and introducing interdependent modules does complicate it. At this point in time we have already exhausted more resources than anticipated on using opm-parser in Resinsight (and we are still not there), so from my end this is as much about risk as anything else.

With regards to the J-function. Maybe you can share what you are trying to accomplish. Do you intend to do the unit conversion in opm-parser, calculating capillary pressure curves and presenting  them SI units for the simulator?

joakim-hove:
OK - my last comment came out a bit wrong. My preferences are:

1. Make the code in opm-parser the canonical unit code; as suggested by this PR.
2. Move the canonical unit code to opm-common, and let opm-parser handle that dependency manually by maintaining a local copy.

But I will accept either solution.

joakim-hove:
> [....] Do you intend to do the unit conversion in opm-parser, calculating capillary pressure curves and presenting them SI units for the simulator?

I would actually say this is the only way to do it; you have to insist on using the API incorrectly to get non-SI values out from the parser.

andlaus:
@alfbr:
> With regards to the J-function. Maybe you can share what you are trying to accomplish.

that's simple: I'd like to get the same results as Eclipse. Now, if you look at the equation for JFUNC, it features something like

gamma * (poro^alpha / perm^beta) * Uconst

the problem is that, as far as I can see, the permeability (unit: m^2) needs to be converted to the unit which is used by Eclipse (mD) because the perm^beta term cannot be converted to SI units by multiplying with a constant (Uconst) since the value of beta is user-definable. Also, I'd like to use the same constant as outlined in the documentation, so the surface tension gamma must be converted from SI units (N/m) to ECL ones (dyn/cm).

atgeirr:
> the equation for JFUNC

Perhaps this page can be of use: https://en.wikipedia.org/wiki/Leverett_J-function

andlaus:
> Perhaps this page can be of use: https://en.wikipedia.org/wiki/Leverett_J-function

true, ECL likes do things slightly differently, though (as usual)

alfbr:
> I would actually say this is the only way to do it; you have to insist on using the API incorrectly to get non-SI values out from the parser.

The J-function is a special case. It is dimensionless, i.e., it has no unit. Hence, passing it as is to, e.g., opm-material for further handling will not violate the SI requirement on the parser API.

andlaus:
> The J-function is a special case. It is dimensionless, i.e., it has no unit.

let's say it is *supposed* to be dimensionless. what Eclipse produces is can only be dimensionless in the default case, though (i.e., beta=0.5).

> Hence, passing it as is to, e.g., opm-material for further handling will not violate the SI requirement on the parser API.

correct. there was no problem if this was a simple scaling factor. (it's not :()

andlaus:
> I would actually say this is the only way to do it; you have to insist on using the API incorrectly to get non-SI values out from the parser.

alright. do you mean it's better to just hard-code the conversion factors?

andlaus:
what I also just noticed: if beta != 0.5 the JFUNC keyword probably behaves significantly different for decks which use e.g. field units from ones that are based on metric ones because it uses different Uconst values. I would be a bit surprised if most reservoir engineers are aware of that...

alfbr:
>  gamma * (poro^alpha / perm^beta) * Uconst

Nice catch with the beta. I have to admit I am in the dark as to whether anybody uses that option. It may be one of those peculiarities we do not want to support.

andlaus:
I just had a hard look at the manual again: the 'Uconst' factor should be identical for all unit systems (because all input quantities except for the unscaled capillary pressure exhibit the same units and the input capillary pressure is supposed to be just scaled by a dimensionless factor), i.e. if the ECL documentation is correct, the effect of JFUNC on decks which use FIELD should always be about factor 15 larger than for decks which use METRIC.

alfbr:
>  'Uconst' factor should be identical for all unit systems 

Not according to the technical description, different factors are listed for each unit system.

andlaus:
> Not according to the technical description, different factors are listed for each unit system.

yep, that was my point: if you look at the equation, the pc(S) term should be scaled by the same factor regardless of the unit system used by the deck. Now, for FIELD decks the factor is about 15 times larger than for METRIC ones because Uconst is different...

andlaus:
well, I should add that with "should be scaled by the same factor" I mean "should be scaled by the same factor if the effect of the JFUNC keyword is supposed to be identical"

OPMUSER:
JFUNC is used all the time to initialize models, not only in Eclipse but 
also other simulators. It is a standard form for normalizing capillary 
pressure, so it is necessary to support this. It is also used with 
endpoint scaling, but that is another interesting topic for another day.
ThunderBird Signature File

andlaus:
> so it is necessary to support this.

I don't doubt this, I just have the suspicion that most people get something different than what they think they get...

alfbr:
> yep, that was my point: if you look at the equation, the pc(S) term should be scaled by the same factor regardless of the unit system used by the deck.

I don't follow you. It is the J-function that is scaled. The factor differs according to what units are used for the other quantities, it is in effect the unit conversion. Maybe we should do a hangout to clarify any misunderstandings?

alfbr:
> JFUNC is used all the time to initialize models, not only in Eclipse but also other simulators. It is a standard form for normalizing capillary pressure, so it is necessary to support this. It is also used with endpoint scaling, but that is another interesting topic for another day.

Indeed, what I am wondering however is whether we need to support something other than the square root of permeability. If you look at record five for JFUNC (that is the beta), you will see that you can use just about any root of permeability, not just the square root. Then again the technical description seems to ignore this and only refers to the standard J-function formulation with the square root of porosity over permeability.

andlaus:
> The factor differs according to what units are used for the other quantities, it is in effect the unit conversion.

if you look up how the permeability, porosity and surface tension are specified, you'll notice that they are actually identical for all ECL unit systems: [mD], [-] and [dyn/cm]. this means that the gamma * (poro^alpha / perm^beta)  part of the equation is going to be the same for all unit systems which implies that Uconst also needs to be the same in order to get the same scaling factor?

alfbr:
> if you look up how the permeability, porosity and surface tension are specified, you'll notice that they are actually identical for all ECL unit systems: [mD], [-] and [dyn/cm]. this means that the gamma * (poro^alpha / perm^beta) part of the equation is going to be the same for all unit systems which implies that Uconst also needs to be the same in order to get the same scaling factor?

The difference is caused by the unit for the pressure. It is bar for metric, psi for field, atm for lab, this is why the constant differs.

OPMUSER:
Indeed, what I am wondering however is whether we need to support
    something other than the square root of permeability. If you look at
    record five for JFUNC (that is the beta), you will see that you can
    use just about any root of permeability, not just the square root.
    Then again the technical description seems to ignore this and only
    refers to the standard J-function formulation with the square root
    of porosity over permeability.


Short answer is no. The form is standard, but there are other methods to 
normalized cap curves  Brooks-Corey Function, Lambda Function, Thomeer 
Function, Skelt-Harrison, and one from your part of the world 
Skjaeveland. And then geologists who just use a saturation height 
function aka Petrel.
ThunderBird Signature File

atgeirr:
Apart from the JFUNC discussion, I think that this PR should be closed as not needed. Further on, the opm-core unit implementation should be moved to opm-common (and gain the dyne unit). For the foreseeable future opm-parser will retain its own version. Agree @andlaus and @joakim-hove?

andlaus:
yes

joakim-hove:
No

andlaus:
> No

re-open the PR if you disagree; I'm fine with anything as long as we can get our stuff done. (IMO the technically best solution would be to make opm-parser depend on opm-common.)

joakim-hove:
My main  objection is that I dø not want to move units to opm-common. The parser and output are the main users of units in the flow line, and then it is really funny if the parser should have second class access to the units. But I realise this involves several developers and I will bend over to a large opposing agreement.

andlaus:
> My main objection is that I dø not want to move units to opm-common.

this statement seems to be pretty much anchored in the ECL world? I suppose that as soon as other file formats should be supported, it would be handy have this functionality provided by a base module. (and, honestly, I can't see what could possibly be more basic than 'common'?)

alfbr:
It seems you are making the assumption that other file formats would be supported in separate parser and output repositories. I believe that would be a far greater waste through code duplication. Hence, I agree with Joakim here. The less we need to deal with units in the simulator code, the better.

akva2:
Nobody says deal with it in simulator code. There is a module downstream of opm-common needing the code, a module which is not downstream of opm-parser. Either you put in opm-common or it will have to be duped in that module. Logic dictates opm-common, keeping it in opm-parser is something else than logic.

alfbr:
> There is a module downstream of opm-common needing the code,

I know, and that actually underlines my point. I believe we are dealing with units too many places, while it should only be needed at IO.

akva2:
the only scenario that would work is if everything is non-dimensionalized on input (and dimensionalized on output). if not, it seems very sensible that diagnostics need access to conversion factors.

i would also like to point out that nothing changes in terms of dependencies if units are moved to opm-common; opm-parser already has a build-time dependency on opm-common (for buildsystem bits), and the units conversion code is a pure template. thus, opm-parser remains free of a runtime dependency on opm-common, while not introducing a dependency on the parser in the diagnostics code.

in effect you are arguing for the opposite view of what you state, although with a little different interpretation of 'dealt with at too many places' (code wise, not logic wise).

alfbr:
The diagnostics code uses conversion because it does the IO itself with raw data from libecl. If libecl provided IO in SI units, or the diagnostics code did IO through opm-output, it would not need any conversion. In effect, the diagnostics code duplicates functionality for wrapping libecl.

akva2:
right, i wasn't aware. the situation is; opm-flowdiagnostics-applications uses ert directly. it (currently) uses the conversion factors from opm-core. as you say, it could do I/O through opm-output, thus implicitly gaining the opm-parser dependency. i assume there are reasons (those reason even being as simple as legacy code and no time to change) for using ert directly.

to me it seems that putting the conversion factors in opm-parser is an acceptable solution for all parts.
opm-parser gets "first class" access to the constants. the situation in opm-flowdiagnostics-applications improve, the remaining possibly becoming a non-issue depending  the resolution to the orthogonal issue with using opm-parser. you lose the build dependency to opm-core (and opm-material), while keeping the build dependency to opm-parser. and there is (still) no runtime dependency to opm-parser. nothing changes for opm-flowdiagnostics.

atgeirr:
About the possibilities for opm-flowdiagnostics-application (recall that my wish was for that module to shed its dependency on opm-core, retaining only opm-common and ERT dependencies): having ERT provide SI numbers is probably a big job, so I believe that is a no-go. Using opm-output for reading the necessary fluxes is not possible for two reasons: opm-output does not read or consider them at all (and adding it would be a major job) and it would retain a stronger dependency on other modules (parser and output).

If the proposed solution here is acceptable, I do not know, but I have assumed not. However, if we can assume that ResInsight uses opm-parser, then it may be acceptable for opm-flowdiagnostics-applications (or rather, the library part of it containing the classes that read the fluxes and builds the graphs) to depend on it.

atgeirr:
After discussions today with @JacobStoren and others, I have changed my mind, and reached the conclusion that the current proposal *is* feasible for opm-flowdiagnostics-applications, and as @akva2 pointed out, it is an overall improvement.

So: I am open to merging this chain of PRs. Unless anyone else pipes in I'll merge it all later today.

andlaus:
seems like an agreement was reached, nice!

note that my PR series did not include the `opm-flowdiagnostics` module. (I was unaware of it, sorry.)

andlaus:
> note that my PR series did not include the opm-flowdiagnostics module. (I was unaware of it, sorry.)

ok, looks like it does not need any adaptation because of the units stuff. *phew*

atgeirr:
> ok, looks like it does not need any adaptation because of the units stuff. phew

It does, but it's a tiny fix and I'll do it. As of now, those modules are not released, and while I try to avoid things needlessly breaking, I do not expect you or others to build or otherwise test or fix those modules. So no worries!

andlaus:
strange: `make` worked flawlessly for me. maybe I disabled the unit tests or a dependency was missing...

PR #940, opened 2016-10-17
changed: rename function getType

akva2:
typeof is a semi-reserved symbol in gcc (gcc extension to get typeid)


jokva:
Merge whenever.

PR #939, opened 2016-10-14
The great shared_ptr purge

jokva:
Hi,

For a long time we've talked about getting rid of the pointless shared_ptr use in opm-parser, and now most of them are gone. This has a few implications worth mentioning:

* I've unearthed some bugs because of this
* Deck is now copyable, and EclipseState and friends will hopefully be soon.
* Copying generally now means copying
* Storage type is no longer mandated by opm-parser; the decision is left to the caller.
* Interfaces are clearer i.e. they don't imply ownership sharing anymore, which was never really the intention.

This PR does not really change any behaviour. However, downstream code WILL break from both interface changes and the removal of all the *Ptr aliases. Downstream PRs will be coming.

PR #943, opened 2016-10-19
fix `make install`

andlaus:
some CMakeLists.txt needed some adaptation after the unit
changes. sorry for not noticing this earlier...

### DIFF
--- a/opm/parser/eclipse/Units/Units.hpp
+++ b/opm/parser/eclipse/Units/Units.hpp
@@ -242,15 +242,6 @@ namespace Opm {
                 return q / unit;
             }
         } // namespace convert
-
-
-#ifndef HAS_ATTRIBUTE_UNUSED
-        namespace detail {
-            // Some units are sometimes unused, and generate a (potentially) large number of warnings
-            // Adding them here silences these warnings, and should have no side-effects
-            constexpr double __attribute__((unused)) unused_units = stb + liter + barsa + psia + darcy;
-        } // namespace detail
-#endif
     }
 
     namespace Metric {
--- a/opm/parser/eclipse/CMakeLists.txt
+++ b/opm/parser/eclipse/CMakeLists.txt
@@ -167,7 +167,7 @@ Generator/KeywordGenerator.hpp
 #
 Units/UnitSystem.hpp
 Units/Dimension.hpp
-Units/ConversionFactors.hpp
+Units/Units.hpp
 #
 EclipseState/EclipseState.hpp
 EclipseState/EclipseConfig.hpp


jokva:
I've also added these fixes to my shared_ptr purge, so if desired this can be fixed as a part of that.

andlaus:
hm: the appveyor errors seem to be caused by some GCCisms in Units.hpp: as far as I understand the error message, it does not seem to swallow the `__attribute__((unused))` statement which was added to squelch some "unused variable" GCC warnings. If I remember correctly, that one has not been required anymore after the units have been marked as `constexpr`? I pushed a patch, let's see if it helps...

> I've also added these fixes to my shared_ptr purge, so if desired this can be fixed as a part of that.

I'm fine with this. do as you prefer.

jokva:
> hm: the appveyor errors seem to be caused by some GCCisms in Units.hpp: as far as I understand the error message, it does not seem to swallow the __attribute__((unused)) statement which was added to squelch some "unused variable" GCC warnings. If I remember correctly, that one has not been required anymore after the units have been marked as constexpr? I pushed a patch, let's see if it helps...

Also fixed in my shared_ptr purge.

opm/parser/eclipse/Units/Units.hpp
```diff
@@ -242,15 +242,6 @@ namespace Opm {
                 return q / unit;
             }
         } // namespace convert
-
-
-#ifndef HAS_ATTRIBUTE_UNUSED
```
akva2: yes. @babrodtk please confirm, even though it doesn't matter any longer this block can definitely go.

akva2:
yeah, msvc does not support the attribute unused. this bit us now because this code hasn't been built on msvc before and the ifdef was wrong as you pointed out triggering the issue.

andlaus:
hm, depending on how long #939 takes until it stabilizes, we should consider merging this PR before?

atgeirr:
I'd like to have this merged if possible, to ensure Jenkins tests work properly. Is that OK?

opm/parser/eclipse/Units/Units.hpp
babrodtk: @akva2, @andlaus I agree that it should be #if

andlaus:
> Is that OK?

I suppose it is. @jokva will probably get some (very minor AFAICS) conflicts with this shared_ptr purge, though...

pgdr:
@atgeirr @andlaus no, let's wait for #939.

andlaus:
how long is this gonna take approximately?

jokva:
Probably today (finally got bloody mpi+zoltan to compile).

atgeirr:
This is *really* holding us (well, Jenkins) back right now, and I do not see the conflict with #939 at all (other than perhaps the same fix being applied there). That is a quick rebase away with no conflicts (or it should be, if not please tell me why).

Semi-related: I have asked @GitPaean to try to fix the opm-grid Zoltan issues with this PR so that we can get #939 in as quickly as possible. He has the same setup as @blattms.

atgeirr:
> Probably today (finally got bloody mpi+zoltan to compile).

Saw this after I posted my comment above: does this mean that you have already fixed opm-grid, so all we have to wait for is the final Jenkins run?

jokva:
> Saw this after I posted my comment above: does this mean that you have already fixed opm-grid, so all we have to wait for is the final Jenkins run?

Seems so, but you never know.

atgeirr:
> Seems so, but you never know.

Based on that, I am willing to wait.

blattms:
> @atgeirr @andlaus no, let's wait for #939.

I totally disagree with this approach as I consider it very bad style
to collect multiple independent fixes in PRs that change the
semantics. Every self contained fix deserves its own PR. This makes it
a lot easier to identify the origin of problems (bugs, performance
regressions, etc.) and might save a lot of time afterwards.

There please merge this independently of the shared_ptr stuff.

jokva:
A PR doesn't really help with that, although commits do. The fix is in its own commit (and a better one in my opinion, although this can certainly be changed to be just that).

It will take just as long though considering this doesn't have a confirmed jenkins build. I can also rip out my own fix and submit it separately, but again, no we probably still have to wait for jenkins.

pgdr:
>>  @atgeirr @andlaus no, let's wait for #939.

> I totally disagree [...] Every self contained fix deserves its own PR. This makes it
a lot easier to identify the origin of problems [...]

We don't use PRs to identify the origin of problems, but commits.

> There please merge this independently of the shared_ptr stuff.

Please do not.  Please understand that merging the "shared_ptr stuff" is a big job that spans every repository.  Every single new merge puts the "shared_ptr stuff" back many hours.  We do not merge until Jenkins or Travis gives the green light.

atgeirr:
> We don't use PRs to identify the origin of problems, but commits.

In practice, yes we do. Commits are often too small a unit to use in discussions about a problem.

> Please do not. Please understand that merging the "shared_ptr stuff" is a big job that spans every repository. Every single new merge puts the "shared_ptr stuff" back many hours. We do not merge until Jenkins or Travis gives the green light.

OK. But the only reason this is a problem was that #939 had a fix for the same problem added to it. Otherwise we could have merged this without causing any problem for #939.

PR #945, opened 2016-10-19
fix the PORV grid property for clang with optimizations

andlaus:
for some reason, it looks like `std::find_if()` does not work on
ubuntu 16.10 with clang and enabled optimizations. since as far I can
see this 'if' is a minor performance improvement in non-time critical
code, let's just remove the condition.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
+++ b/opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
@@ -74,17 +74,16 @@ namespace Opm {
             const auto& poro = doubleGridProperties->getKeyword("PORO");
             const auto& ntg =  doubleGridProperties->getKeyword("NTG");
 
-            if (poro.containsNaN())
-                throw std::logic_error("Do not have information for the PORV keyword - some defaulted values in PORO");
-            else {
-                const auto& poroData = poro.getData();
-                for (size_t globalIndex = 0; globalIndex < poro.getCartesianSize(); globalIndex++) {
-                    if (!std::isfinite(values[globalIndex])) {
-                        double cell_poro = poroData[globalIndex];
-                        double cell_ntg = ntg.iget(globalIndex);
-                        double cell_volume = eclipseGrid->getCellVolume(globalIndex);
-                        values[globalIndex] = cell_poro * cell_volume * cell_ntg;
-                    }
+            const auto& poroData = poro.getData();
+            for (size_t globalIndex = 0; globalIndex < poro.getCartesianSize(); globalIndex++) {
+                if (!std::isfinite(values[globalIndex])) {
+                    double cell_poro = poroData[globalIndex];
+                    if (std::isnan(cell_poro))
+                        throw std::logic_error("Some cells neither specify the PORV keyword nor PORO");
+
+                    double cell_ntg = ntg.iget(globalIndex);
+                    double cell_volume = eclipseGrid->getCellVolume(globalIndex);
+                    values[globalIndex] = cell_poro * cell_volume * cell_ntg;
                 }
             }
 
--- a/opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
+++ b/opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
@@ -71,22 +71,19 @@ namespace Opm {
                        const GridProperties<int>* intGridProperties,
                        const GridProperties<double>* doubleGridProperties)
         {
-            const auto iter = std::find_if( values.begin(), values.end(), []( double value ) { return std::isnan(value); });
-            if (iter != values.end()) {
-                const auto& poro = doubleGridProperties->getKeyword("PORO");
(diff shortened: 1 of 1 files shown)


jokva:
What do you mean "doesn't work"? Does it not compile? Does it evaluate wrong?

andlaus:
it compiles, but I get NaNs in the values of the PORV grid property. this only happens on clang and only when optimizations are enabled. maybe clang has a problem with lambdas.

anyway, just removing the that "optimization" could even make the resulting code slightly faster, because if there are no nans in the array, the loop does the same thing as what the `find_if` has done before and the version after this patch only goes over the memory once...

jokva:
I'm more worried about having to work through piles of compiler bugs than changing this, however, if there is a standards violation I'd like to know.

andlaus:
even if it was not required for clang, it would still be an improvement? (i.e., after this, we're jumping through less hoops than before!?)

blattms:
Are you using `-ffast-math` for compilation?

andlaus:
> Are you using `-ffast-math` for compilation?

yes, but with an additional `-fno-finite-math-only`. I'm not sure if I  tested without `-ffast-math`, but I'm sure that I downgraded the optimizations from `-Ofast -march=native` to `-O2` to no effect.

blattms:
You are a brave man.

Well, that might be the cause for this, see the gcc manual page
```
       -fcx-limited-range
           When enabled, this option states that a range reduction step is not
           needed when performing complex division.  Also, there is no
           checking whether the result of a complex multiplication or division
           is "NaN + I*NaN", with an attempt to rescue the situation in that
           case.  The default is -fno-cx-limited-range, but is enabled by
           -ffast-math.

           This option controls the default setting of the ISO C99
           "CX_LIMITED_RANGE" pragma.  Nevertheless, the option applies to all
           languages.
```

andlaus:
> Well, that might be the cause for this, see the gcc manual page

thanks, good to know. on my machine it currently works with GCC using the same flags as clang, but with your info considered, I suppose it could break after any GCC release.

Irrespective of whether this is strictly required, do you agree that the removal of the `find_if(); if()` statements result in (slightly) better code?

jokva:
Not necessarily. With the `find_if` still short-circuits the vector fetches somewhat more elaborate algorithm with a not-so-obvious fall-through for non-NaNs¹, and clarifies intent a bit (even though the algorithm itself is structured poorly, I might fix that some other time). Also, the right function to use would be `std::all_of`. By using find_if we can even speed it up a bit more, achieving single pass, by starting the counter at the first NaN.

¹ I chuckled typing that.

andlaus:
> Not necessarily. With the find_if it still short-circuits the vector fetches somewhat more elaborate algorithm with a not-so-obvious fall-through for non-NaNs¹,

did you check this or is this just speculation? (I.e., on one hand I'm willing to bet a beer on the fact that there is no noticeable performance difference for any real world deck, while on the other hand, the proposed version clearly is simpler and thus I'll claim that it is better from a maintainability POV.)

opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -71,22 +71,19 @@ namespace Opm {
...
-                        if (std::isnan(values[globalIndex])) {
-                            double cell_poro = poroData[globalIndex];
-                            double cell_ntg = ntg.iget(globalIndex);
-                            double cell_volume = eclipseGrid->getCellVolume(globalIndex);
-                            values[globalIndex] = cell_poro * cell_volume * cell_ntg;
-                        }
+            const auto& poro = doubleGridProperties->getKeyword("PORO");
+            const auto& ntg =  doubleGridProperties->getKeyword("NTG");
+
+            if (poro.containsNaN())
```
joakim-hove: This has slightly different semantics than the old solution. It is perfectly possible to fully initialize the `PORV` property from the deck with the `PORV` keyword, in that case the `PORO` keyword will be `NaN` all over the place - without any harm! So - the logical equivalent of the `std::find_if( ... )` - checking if there are `NaN` values in the `PORV` vector, must be performed before possibly raising an exception.

The normal path here will be that `values` vector is all `NaN` - I guess that information should influence the implementation. 

When this has been addressed I am happy to merge.

andlaus: I see: You're right, but I suppose that PORO only needs to be defined for cells where PORV is NaN? that case is not handled correctly by the current code either, right?

anyway: fix pushed, thanks.

joakim-hove:
Hmmm - this should be green om travis now?

andlaus:
> Hmmm - this should be green om travis now?

right. the great shared_ptr purge probably came into the way (the error seems to be unrelated,though: it's in opm-upscaling, and about some *ConstPtr in a PVT object.) I've rebased this PR, maybe it helps...

andlaus:
now travis is happy. seems like the reason for the failure really was some screw-up related to the shared_ptr purge (I did not change anything in this PR)...

PR #946, opened 2016-10-20
Check for separator/quote table overflow

jokva:
Just relying on the char data type is not sufficient to guard against
overflows, and several input decks would invoke undefined behaviour.
This code path is extremely hot, so bit-wise checking for overflow is
done over a two-branch check.


jokva:
Please note that this requires some more discussion @joakim-hove, so we'll hold off on merging this for now.

PR #948, opened 2016-10-21
Changes to MessageLimits:

joakim-hove:
1. Will retain already set values when only some items are set.
 2. Will parse the sections before the SCHEDULE section to create a
    correctly initialized MessageLimits object.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
@@ -88,6 +88,20 @@ namespace Opm {
         m_controlModeWHISTCTL = WellProducer::CMODE_UNDEFINED;
         addGroup( "FIELD", 0 );
 
+        /*
+          We can have the MESSAGES keyword anywhere in the deck, we
+          must therefor also scan the part of the deck prior to the
+          SCHEDULE section to initialize valid MessageLimits object.
+        */
+        for (size_t keywordIdx = 0; keywordIdx < deck.size(); ++keywordIdx) {
+            const auto& keyword = deck.getKeyword(keywordIdx);
+            if (keyword.name() == "SCHEDULE")
+                break;
+
+            if (keyword.name() == "MESSAGES")
+                handleMESSAGES(keyword, 0);
+        }
+
         if (Section::hasSCHEDULE(deck)) {
             std::shared_ptr<SCHEDULESection> scheduleSection = std::make_shared<SCHEDULESection>(deck);
             iterateScheduleSection(parseContext , *scheduleSection , grid );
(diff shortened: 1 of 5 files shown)


pgdr:
travis fails with
```
Program threw an exception: [/home/travis/build/OPM/opm-output/opm/test_util/EclFilesComparator.cpp:387] 
Keywords in first file: 17
Keywords in second file: 16
The number of keywords differ.
```

alfbr:
I guess this means that the MESSAGES keyword is now found in result files and the reference solution needs to be updated accordingly.

joakim-hove:
It is certsinly thst a new keyword has been added to one of the output files, and hence the reference needs an update, but I think it is a previous recent merge somewhere which is to blame, because this PR does not affect binary output.

jokva:
If the binary reference solutions needs to be updated essentially every time we patch anywhere I'm thinking a git repo + essentially binary diffs is a very good solution.

alfbr:
I really hope results don't change every time something is patched. I am not sure what you are proposing, but for a transitional period it may make sense to automate the binary update procedure somehow. Right now it suffices to identify which change took place, rerun the test cases and send in the updated results in a PR. I am currently on travel, so very grateful if somebody steps up :-)

atgeirr:
After some testing with some simple modified SPE1 test decks, I can confirm that this works as expected:
 - MESSAGES before schedule: seen as from first time step.
 - MESSAGES both before schedule and at first time step: the one actually at the time steps overrides the one from before the schedule.
 - MESSAGES later in schedule: not seen from first time step.

While not exhaustive, I am fairly confident that this works well enough to be a reasonably useful feature for users.

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -1163,37 +1177,32 @@ namespace Opm {
...
+                                                                        {"ERROR_PRINT_LIMIT"   , &MessageLimits::setErrorPrintLimit},
+                                                                        {"BUG_PRINT_LIMIT"     , &MessageLimits::setBugPrintLimit},
+                                                                        {"MESSAGE_STOP_LIMIT"  , &MessageLimits::setMessageStopLimit},
+                                                                        {"COMMENT_STOP_LIMIT"  , &MessageLimits::setCommentStopLimit},
+                                                                        {"WARNING_STOP_LIMIT"  , &MessageLimits::setWarningStopLimit},
+                                                                        {"PROBLEM_STOP_LIMIT"  , &MessageLimits::setProblemStopLimit},
+                                                                        {"ERROR_STOP_LIMIT"    , &MessageLimits::setErrorStopLimit},
+                                                                        {"BUG_STOP_LIMIT"      , &MessageLimits::setBugStopLimit}};
+
```
jokva: Nifty solution! A couple of points:

Just for fun, there's not *that* much of a vector (implies dynamic allocation) for this; by using auto (initializer_list) or just a raw or std::array we'd save an allocation. Finally, there's a cool thing in functional called mem_fn. Consider:

```
    void Schedule::handleMESSAGES( const DeckKeyword& keyword, size_t currentStep) {
        const auto& record = keyword.getRecord(0);
        using set_limit = decltype( std::mem_fn( &MessageLimits::setMessagePrintLimit ) );
        std::pair< std::string, set_limit > setters[] = {
            {"MESSAGE_PRINT_LIMIT", std::mem_fn( &MessageLimits::setMessagePrintLimit ) },                                                                                                                                                                                        
            {"COMMENT_PRINT_LIMIT", std::mem_fn( &MessageLimits::setCommentPrintLimit ) },
            {"WARNING_PRINT_LIMIT", std::mem_fn( &MessageLimits::setWarningPrintLimit ) },
            {"PROBLEM_PRINT_LIMIT", std::mem_fn( &MessageLimits::setProblemPrintLimit ) },
            {"ERROR_PRINT_LIMIT"  , std::mem_fn( &MessageLimits::setErrorPrintLimit ) },
            {"BUG_PRINT_LIMIT"    , std::mem_fn( &MessageLimits::setBugPrintLimit ) },
            {"MESSAGE_STOP_LIMIT" , std::mem_fn( &MessageLimits::setMessageStopLimit ) }, 
            {"COMMENT_STOP_LIMIT" , std::mem_fn( &MessageLimits::setCommentStopLimit ) }, 
            {"WARNING_STOP_LIMIT" , std::mem_fn( &MessageLimits::setWarningStopLimit ) },
            {"PROBLEM_STOP_LIMIT" , std::mem_fn( &MessageLimits::setProblemStopLimit ) },
            {"ERROR_STOP_LIMIT"   , std::mem_fn( &MessageLimits::setErrorStopLimit ) },
            {"BUG_STOP_LIMIT"     , std::mem_fn( &MessageLimits::setBugStopLimit ) }
        };    

        for (const auto& pair : setters) {
            const auto& item = record.getItem( pair.first );
            if (item.defaultApplied(0)) continue;

            int value = item.get<int>(0);
            pair.second( this->m_messageLimits, currentStep, item.get< int >( 0 ) );
        }     
    }
```

Pretty cool!

This is, however, ready whenever.

joakim-hove:
OK - have updated according to suggestion from @jokva. Will merge when travis is green.

PR #949, opened 2016-10-23
appveyor clones opm-data

jokva:


jokva:
Alright, this seems to work now.

PR #950, opened 2016-10-24
Stack allocd items

jokva:
Reimplement ParserItem and DeckItem to use a sum-type variant rather than relying on inheritance and pointer casting or tagging to determine type. All items are now value alloc'd in their respective records rather than indirected.


joakim-hove:
Hmm  - I am ready to merge this; but Travis seems unhappy?

jokva:
Ugh, yeah, Travis found some old tests in opm-core. I thought I had them patched, but maybe I didn't create the PR. I'll have a look.

jokva:
I've created a PR for downstream too.

PR #953, opened 2016-10-26
Only add present phases from WCONHIST

### LINKED ISSUE #829: Production controls added for phases that don't exist
I am trying to run a two-phase problem (oil-water) in flow (master branch), but I get the following exception:

> [<...>/opm-core/opm/core/wells/WellsManager.cpp:599] Gas phase not active and GRAT control specified.

After some investigation I found the code where the GRAT-control is added, which is [here](https://github.com/OPM/opm-parser/blob/c46b1cb9369d328056bd6a9e45513cbca5420650/opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.cpp#L52) (the case uses WCONHIST):

``` cpp
for (auto const& modeKey : {"ORAT", "WRAT", "GRAT", "LRAT", "RESV", "GRUP"}) {
    auto cmode = WellProducer::ControlModeFromString( modeKey );
    p.addProductionControl( cmode );
}
```

This code seems to add all production controls without checking what phases are active. Just removing "GRAT" from this list fixes the problem (but that will obviously break something for three-phase runs).

The code for prediction mode ([here](https://github.com/OPM/opm-parser/blob/c46b1cb9369d328056bd6a9e45513cbca5420650/opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.cpp#L95)) seems to be doing something smarter

``` cpp
for (auto const& modeKey : {"ORAT", "WRAT", "GRAT", "LRAT","RESV", "BHP" , "THP"}) {
    if (!record.getItem(modeKey).defaultApplied(0)) {
        auto cmode = WellProducer::ControlModeFromString( modeKey );
        p.addProductionControl( cmode );
    }
}
```

I imagine the issue can be fixed by something like [this](https://gist.github.com/ANerd/adc224e99890ebff75cbed4959139d82), but I don't know enough about the code to know if this breaks something else.

ANerd:
Fixes #829.

I have used this patch for a while now, but haven't made a PR before since the approach may have a few problems:

I tried not to break any interfaces, as I don't know where else they might be used. Passing by pointer may not be ideal, but that was the easiest way to stay compatible.

I don't know if the lifetime of Schedule is limited by EclipseState (as it is a shared pointer) and this approach would obviously fail if Schedule outlives the EclipseState.


ANerd:
The travis failure is due to problem described in OPM/opm-output#121.

We could handle this more gracefully, but on the other hand, I can't imagine a case where it makes sense to set the control mode to a phase that doesn't exist.

joakim-hove:
Thank you - we will merge this, or a functionally equivalent alternative. I have not looked thoroughly on it - but my initial reaction was that it should be possible to pass only the relevant phase information from the tables, and not the complete table object.

jokva:
Hi,

Thanks for bringing this up, and with a patch no less!

Something that's been on the TODO for a while is to lift the phase information from TableManager to Schedule or something similar, which I'm working on right now. I can see that this will in a sense conflict with your patch because as a consequence we might remove the phase information from TableManager altogether (the details will be worked out tomorrow). On the other hand, that should simplify your patch because you now actually have access to the phase information you need, right?

What do you think of modifying this a bit once we patch Schedule to also expose phase information?

ANerd:
It shouldn't be a problem to simplify this patch once Schedule is updated, but I may a bit busy tomorrow. If you want to integrate this into another PR or fix it in a separate PR that's fine by me. If not, I can probably update this PR on Wednesday.

jokva:
If this patch fixes your immediate problem so you can get your work done we're perfectly fine waiting.

ANerd:
I already use this patch locally, so my immediate problem is already solved. I will just continue using my existing build of flow until the conflict is resolved, so breaking this PR does not affect my work. Thanks for the consideration!

jokva:
Alright, so we've written https://github.com/OPM/opm-parser/pull/958 which addresses one glaring issue in the parser interface. Phase information is moved from the tables to its own `runspec` tree (which with that PR only has one member; phases).

This does conflict with your PR though @ANerd, so it will probably have to be rewritten. However, this *should* fix some of the design issues you ran into. The phases object is tiny and is meant to always be easy to pass and/or copy around, so you can easily amend the `Schedule` constructor or object with phase information to accomplish what you want.

An idea could for instance be to embed phase information in the production facilities altogether which internally would handle whether or not some WCONHIST information made sense (in which case adding would be a no-op, but this choice was made internally).

joakim-hove:
OK - apart from the use of a optional `Phases` argument to the `Schedule::Schedule( )` I am ready to merge this.

opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp
```diff
@@ -52,7 +53,7 @@ namespace Opm
     class Schedule {
     public:
         Schedule(const ParseContext& parseContext, const EclipseGrid& grid,
-                 const Deck& deck );
+                 const Deck& deck, const Phases &phases = Phases(true, true, true) );
```
ANerd: Just afraid of breaking downstream stuff. Removed now.

jokva:
That's ok, the `Schedule` constructor is largely to be considered internal (i.e. all downstreams trying to get a hold of a `Schedule` instance go via `EclipseState`).

ANerd:
Travis failure due to python bindings. Made a very (too?) simple fix.

As the python bindings doesn't work on my system (python 2 vs 3 problems), I'm working blind on this. If you want to expose this parameter to python, I'm probably not the right person.

opm/parser/eclipse/python/c_inter/cschedule.cc
```diff
@@ -22,7 +22,7 @@ extern "C" {
 
 
     Opm::Schedule * schedule_alloc(Opm::ParseContext * context, Opm::EclipseGrid * grid, Opm::Deck * deck) {
-        return new Opm::Schedule( *context , *grid, *deck );
+        return new Opm::Schedule( *context , *grid, *deck, Opm::Phases(true, true, true) );
```
joakim-hove: Please replace with:
```C++
return new Opm::Schedule( *context , *grid, *deck, Opm::Phases(*deck) );
```
and you are all good.

joakim-hove:
Thank you for the effort.

PR #954, opened 2016-10-28
changelog 2016-10

jokva:


pgdr:
This is a good initiative.  Will we manage to maintain it?

Do you suggest we make monthly tags/releases as well to go with the notes?  Or sha's?  It seems natural that every update/entry of this list corresponds to a fixed point in the history?

jokva:
I made it a short synopsis and monthly to make it easier to make it a routine (took me the better of four minutes to write it).

I don't think it mandates releases, that's a separate issue. Some features could probably include the sha short ref, but it's mostly intended for users to get a super quick overview of what's been going on lately.

pgdr:
I don't think it will take more than a couple of minutes to write, no, but will we manage to keep it up-to-date every month?  Or will we forget it?  Do you take the responsibility of keeping it up-to-date?

jokva:
I don't mind the responsibility, no. If it turns out we *do* forget it we'll have to devise something else.

andlaus:
> This is a good initiative. Will we manage to maintain it?

I can't answer this for you, so I'd this is just my own experience: I personally do not enjoy wading through the git logs to extract a changelog before release too much, but after the fact I'm always happy that I did it...

PR #955, opened 2016-10-28
Always applying the BHP limit from WCONINJE.

GitPaean:
This is a PR looking for suggestion. 
In our model, we need to handle some statement like the following when using group control. 

```
WCONINJE
'INJE1' 'WATER' 'OPEN' 'GRUP' /
/
```

With the current implementation, basically, it gives no control for the well `INJE1` now. And the results from Eclipse strongly suggest that they use the defaulted BHP limit value ( `6803 atma, or 1.0E5 psia, or 6895 barsa`). That means the defaulted BHP limit always work there.

Please help to suggest if the current proposal works or we should do it in some different way. We got the desired result with this PR. 

BTW: I am not sure what unit system we should use when specify default value in the keyword definition. I saw previous value is 6891, I assume it was barsa there. Please correct me if I was wrong.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
@@ -617,17 +617,15 @@ namespace Opm {
                     properties.dropInjectionControl(WellInjector::THP);
 
                 /*
-                  What a mess; there is a sensible default BHP limit
-                  defined, so the BHPLimit can be safely set
-                  unconditionally - but should we make BHP control
-                  available based on that default value - currently we
-                  do not do that.
+                  There is a sensible default BHP limit defined, so the BHPLimit can be
+                  safely set unconditionally, and we make BHP limit as a constraint based
+                  on that default value. It is not easy to infer from the manual, while the
+                  current behavoir agrees with the behovir of Eclipse when BHPLimit is not
+                  specified while employed during group control.
                 */
                 properties.BHPLimit = record.getItem("BHP").getSIDouble(0);
-                if (!record.getItem("BHP").defaultApplied(0)) {
-                    properties.addInjectionControl(WellInjector::BHP);
-                } else
-                    properties.dropInjectionControl(WellInjector::BHP);
+                // BHP control should always be there.
+                properties.addInjectionControl(WellInjector::BHP);
 
                 if (well->isAvailableForGroupControl(currentStep))
                     properties.addInjectionControl(WellInjector::GRUP);
--- a/opm/parser/share/keywords/000_Eclipse100/W/WSEGITER
+++ b/opm/parser/share/keywords/000_Eclipse100/W/WSEGITER
@@ -1,0 +1,6 @@
+     {"name" : "WSEGITER", "sections" : ["SCHEDULE"], "size" : 1, "items" : [
+         {"name" : "MAX_WELL_ITERATIONS"   , "value_type" : "INT"    , "default" : 20},
+         {"name" : "MAX_TIMES_REDUCED"     , "value_type" : "INT"    , "default" : 5},
+         {"name" : "REDUCTION_FACTOR"      , "value_type" : "DOUBLE" , "default" : 0.3},
+         {"name" : "INCREASING_FACTOR"     , "value_type" : "DOUBLE" , "default" : 2.0}
+     ]}
--- a/opm/parser/eclipse/IntegrationTests/ScheduleCreateFromDeck.cpp
+++ b/opm/parser/eclipse/IntegrationTests/ScheduleCreateFromDeck.cpp
@@ -244,13 +244,13 @@ BOOST_AUTO_TEST_CASE(WellTesting) {
             const WellInjectionProperties& prop9 = well1->getInjectionProperties(9);
             BOOST_CHECK_CLOSE(20000/Metric::Time ,  prop9.surfaceInjectionRate  , 0.001);
             BOOST_CHECK_CLOSE(200000/Metric::Time , prop9.reservoirInjectionRate, 0.001);
-            BOOST_CHECK_CLOSE(6891 * Metric::Pressure , prop9.BHPLimit, 0.001);
+            BOOST_CHECK_CLOSE(6895 * Metric::Pressure , prop9.BHPLimit, 0.001);
             BOOST_CHECK_CLOSE(0 , prop9.THPLimit , 0.001);
             BOOST_CHECK_EQUAL( WellInjector::RESV  , prop9.controlMode);
             BOOST_CHECK(  prop9.hasInjectionControl(WellInjector::RATE ));
             BOOST_CHECK(  prop9.hasInjectionControl(WellInjector::RESV ));
             BOOST_CHECK( !prop9.hasInjectionControl(WellInjector::THP));
-            BOOST_CHECK( !prop9.hasInjectionControl(WellInjector::BHP));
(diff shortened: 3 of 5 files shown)


joakim-hove:
> Dimensionfull default is tricky - I will come up with a suggestion.

OK - to my surprise(?!) it seems we actually handle this quite gracefully. The default values given in the Json configuration should be expressed in metric units; i.e. `BARSA` for pressure values. So - actually I think this should be perfectly OK :grin: 

Currently there is a test failure in Appveyor; which is *probably* a false positive due to a sought after change in behavior:
```xml
<inquisitorial tone> 
    Did you run the tests? 
</inquisitorial tone>
```

I'll merge when the tests are green.

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -603,17 +603,12 @@ namespace Opm {
                     properties.dropInjectionControl(WellInjector::THP);
 
                 /*
-                  What a mess; there is a sensible default BHP limit
-                  defined, so the BHPLimit can be safely set
-                  unconditionally - but should we make BHP control
-                  available based on that default value - currently we
-                  do not do that.
+                  There is a sensible default BHP limit defined, so the BHPLimit can be
+                  safely set unconditionally, and we make BHP control based on that default value.
```
joakim-hove: Please comment here that it is difficult to infer this form the manual, but that the (new) current behavior seems to agree with the behavior of ECLIPSE.

OPMUSER:
The default BHP will be a constraint as the well is under Group control. 
Eclipse would also issue a WARNING message stating the the default value 
of 6803 atma, or 1.0E5 psia, or 6895 barsa is being using depending on 
the units set in the RUNSPEC section.
ThunderBird Signature File

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
GitPaean: Addressed. Thanks for the comments.

GitPaean:
Thanks for the information. I can see it now.  

Do you think it is something we should avoid? 

> The default BHP will be a constraint as the well is under Group control.
Eclipse would also issue a WARNING message stating the the default value
of 6803 atma, or 1.0E5 psia, or 6895 barsa is being using depending on
the units set in the RUNSPEC section

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -603,17 +603,15 @@ namespace Opm {
...
                 */
                 properties.BHPLimit = record.getItem("BHP").getSIDouble(0);
-                if (!record.getItem("BHP").defaultApplied(0)) {
-                    properties.addInjectionControl(WellInjector::BHP);
-                } else
-                    properties.dropInjectionControl(WellInjector::BHP);
+                // BHP control should always be there.
+                properties.addInjectionControl(WellInjector::BHP);
 
                 if (well->isAvailableForGroupControl(currentStep))
```
GitPaean: @joakim-hove  Do you think here will makes the order of `WGRUPCON` and `WCONINJE` cause difference of the behavoir?
I mean, `WGRUPCON` will change the value of `well->isAvailableForGroupControl(currentStep)`. I am not sure what kind of order that opm-parser processing these keywords.

GitPaean: And, should we only add the `GRUP` control when we specify `GRUP` control for the injector? Looks like with the current code, for most of the cases, `GRUP` control is always there.

GitPaean:
Please hold this PR a little bit. More testing is going and it looks like still different from setting BHP control with defaulted BHP value. Probably something related to `GRUP` keyword needs to be addressed.

OPMUSER:
No, as it is a constraint not a target, or should be if the engineer 
knows what he is doing.  We normally set the BHP limit in injection 
wells to the fracture pressure, to avoid fracing the well (in the real 
world not the numerical world).

To be clear, we issue a warning message and continue.
ThunderBird Signature File

GitPaean:
Well control is not properly initialized when the control mode is `GRUP` in keyword `WCONINJE`. 
A fix applied to the `init` function of `WellStateFullyImplicitBlackoil`. This PR is ready for merging at will.

GitPaean:
Adding a definition of the keyword `WSEGITER` here. K-model has it while I am not sure if it is used there. 
It is not related to the original purpose of this PR. If it is a problem, I can split them to two PRs as requested.

opm/parser/share/keywords/000_Eclipse100/W/WSEGITER
```diff
@@ -0,0 +1,6 @@
+     {"name" : "WSEGITER", "sections" : ["SPECIAL", "SCHEDULE"], "size" : 1, "items" : [
```
joakim-hove: The `SPECIAL` section should be removed, the `SPECIAL` section was a mistake which was introduced at some time, and have appeared again and again after.

joakim-hove:
OK - I will merge this when Travis is green (and the `SPECIAL` section is removed from the `WSEGITER` keyword). Should the merge be coordinated with other repos?

GitPaean:
Removed `SPECIAL`.  I think this PR can just go in.

GitPaean:
@joakim-hove . I completed the definition of `GCONPROD` and also put a E300 version in the directory 
001_Eclipse300. Could you take a look to tell if it is the correct way?

joakim-hove:
> I completed the definition of GCONPROD and also put a E300 version in the directory 
001_Eclipse300. Could you take a look to tell if it is the correct way?

This looks OK.

joakim-hove:
As I pointed out already [here](https://github.com/OPM/opm-parser/pull/955#issuecomment-256961073) - this PR has a failing test; that must be fixed before it can be merged.

GitPaean:
Fixed the `runScheduleCreateFromDeck` test. 

All the test can pass now, while with `ctest -V`, the following is reported (the same was reported for the master branch.) I am not sure it is a problem. 
```
 251 27: Test command: /home/kaib/OPM-devel/debug/opm-parser-build/bin/runScheduleTests
 252 27: Test timeout computed to be: 1500
 253 27: Running 34 test cases...
 254 27: ERROR the keyword COMPLUMP is not supported used when C1 or C2 in WELOPEN have values
 255 27: ERROR the keyword COMPLUMP is not supported used when C1 or C2 in WELOPEN have values
 256 27: ERROR the keyword COMPLUMP is not supported used when C1 or C2 in WELOPEN have values
 257 27: 
 258 27: *** No errors detected
 259  27/132 Test  #27: runScheduleTests ......................   Passed    0.28 sec
```

akva2:
It is fine. It just means that somebody did good and included some negative tests in the test suite (that things fail when they are supposed to).

joakim-hove:
> It just means that somebody did good and included some negative tests in the test suite

It is actually my experience that testing the sorry-path is actually the most important; that often involves non-trivial logic added as an afterthought and it receives orders of magnitude less normal-use testing than the happy-path.

joakim-hove:
> All the test can pass now,

Good - I will merge as soon as Travis has dragged itself through the full chain.

akva2:
Fully agree i even included a compliment to hint at it ;)

joakim-hove:
> Fully agree i even included a compliment to hint at it ;)

:smile:

GitPaean:
Hi, can this PR go in now? It will make new development easier. Thanks.

PR #956, opened 2016-10-31
Support JFUNC in TableManager

pgdr:
With this PR,

* the `JFUNC` keyword is internalized in `TableManager` by the function `.useJFunc()`
* the tables that should respect `JFUNC` will not be SI converted when tables are constructed:
 * `SWFN`, `SGFN`, `SWOF`, `SGOF`, `SLGOF`

Notice that the last of the three commits only contain general maintenance.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Tables/SimpleTable.hpp
+++ b/opm/parser/eclipse/EclipseState/Tables/SimpleTable.hpp
@@ -39,7 +39,7 @@ namespace Opm {
         SimpleTable(TableSchema, const DeckItem& deckItem);
         explicit SimpleTable( TableSchema );
         void addColumns();
-        void init(const DeckItem& deckItem);
+        void init(const DeckItem& deckItem, const bool jfunc = false);
         size_t numColumns() const;
         size_t numRows() const;
         void addRow( const std::vector<double>& row);
--- a/opm/parser/eclipse/EclipseState/Tables/SimpleTable.hpp
+++ b/opm/parser/eclipse/EclipseState/Tables/SimpleTable.hpp
@@ -39,7 +39,7 @@ namespace Opm {
         SimpleTable(TableSchema, const DeckItem& deckItem);
         explicit SimpleTable( TableSchema );
         void addColumns();
-        void init(const DeckItem& deckItem, const bool jfunc = false);
+        void init(const DeckItem& deckItem );
         size_t numColumns() const;
         size_t numRows() const;
         void addRow( const std::vector<double>& row);
@@ -60,11 +60,15 @@ namespace Opm {
          */
         double evaluate(const std::string& columnName, double xPos) const;
 
+        /// throws std::invalid_argument if jf != m_jfunc
+        void assertJFuncPressure(const bool jf) const;
+
     protected:
         std::map<std::string, size_t> m_columnNames;
         std::vector<std::vector<bool> > m_valueDefaulted;
         TableSchema m_schema;
         OrderedMap<TableColumn> m_columns;
+        bool m_jfunc = false;
     };
 }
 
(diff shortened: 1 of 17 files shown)


andlaus:
> the tables that should respect JFUNC will not be SI converted when tables are constructed.

it looks to me like the respective columns of the underlying 2p tables are still accessed via `*pressureColumn()`? If this is the case, the PR is not really an improvement because by reading the code, one would still assume that these are pressures and not J(S). IMO, a "good" solution would introduce "jfuncColumn()" methods to these tables which would throw an exception if they are called if JFUNC is not featured by the deck and vice-versa for *pressureColumn().

Also, it would be cool if you would test this on "model 2.2" with flow by running it for a few time steps...

pgdr:
@andlaus I added the `getJFuncColumn`-method.  Are you sure we want to throw when used "inappropriately"?  If so, I will add that as well.

andlaus:
> Are you sure we want to throw when used "inappropriately"?

I'd say "yes" because it is better to get no result than a wrong one!? I guess @joakim-hove will have to decide this in the end..

pgdr:
@andlaus He told me that your will is my command.  Could you check the two last commits of this PR and see if they're ok by you?

We should coordinate the merging of this with the necessary changes in material as well.

andlaus:
now I'm happy with this :)

please do not merge just yet, I still need to fix up opm-material...

pgdr:
Great!  Thanks for reviewing.

Just merge this at will.

pgdr:
Note, I squashed and reordered commits.

opm/parser/eclipse/EclipseState/Tables/SimpleTable.hpp
```diff
@@ -60,11 +60,15 @@ namespace Opm {
          */
         double evaluate(const std::string& columnName, double xPos) const;
 
+        /// throws std::invalid_argument if jf != m_jfunc
+        void assertJFuncPressure(const bool jf) const;
```
andlaus: I just noticed this now while trying to adapt opm-material: I'm pretty sure that this method is specific to (some of) the saturation function tables, i.e., it may be a bit more convenient from an implementor's point of view, but it does not make too much sense IMO. do you mind moving this to `SwofTable` and friends?

pgdr: you think it's better to have this function in five different classes than in one superclass?  What about the `m_jfunc` boolean flag?  Should that go to the five classes as well?

andlaus: to all of your questions: yep, since IMO it makes about zero sense for e.g., `PvtoTable`. alternatively, you can add a `SatfuncSimpleTable` but I think adding this to all satfunc tables individually is better because this is ECL and it might be that this JFUNC handling does apply to all of them.

andlaus:
hm, another problem with this patch that I've just noticed: while it throws at constructs like `sgofTable.getPcogColumn()` if JFUNC is present, the generic variant of accessing the same data (`sgofTable.getColumn("PCOG")`) passes just fine...

pgdr:
yeah, that should be fixed.

andlaus:
hm, I hacked around to adapt opm-material for this yesterday and I changed my opinion during this exercise: with the context dependent pressure columns, the opm-material code just became more complicated, but not better (except -- maybe -- from a conceptual point of view). I guess we should thus consider to leave things as they are now...

pgdr:
Closing this as per discussion.  I will cherry-pick the internalization of the keyword (into TableManager) and the improved test at a later point, and come with a new PR on that.

pgdr:
Thanks for reviewing, Jørgen and Andreas ... in any case, we learnt something from this, so it is not entirely wasted time.

PR #961, opened 2016-11-02
[Discussion] only using WHISTCTL for non-prediction mode.

GitPaean:
I think WHISTCTL should only apply to the history matching wells, not apply to keywords like WCONPROD. 

Any comments? It throws an exception for one model due to the change of mode from WHISTCTL.


OPMUSER:
Correct WHISCTL only applies to wells declared as history matching 
wells,  should throw WARNING/ERROR not sure which.
ThunderBird Signature File

Issue #960, opened 2016-11-02
About SKIPREST


GitPaean:
Currently, OPM will throw an exception if there is no RESTART, 
```
Error in deck: Can not supply SKIPREST keyword without a preceeding RESTART
```

Looks like E100 will just ignore it without giving warning (I am not totally sure about the warning part, while the model can run through no problem.).

GitPaean:
Sorry. I did not see issue #773. closing this one now.

jokva:
https://github.com/OPM/opm-parser/pull/963

Issue #773, opened 2016-04-22
parser issues related to RESTART and SKIPREST


totto82:
The parser throws an error if SKIPREST is present in the deck without RESTART. Eclipse ignores SKIPREST if RESTART is not in the deck.

andlaus:
are the SKIPREST, SKIP100 and SKIP300 keywords supported by opm-parser at all?

joakim-hove:
Comment copied from:https://github.com/OPM/opm-core/issues/1004

Currently FLOW computes the initial state also for restart runs. This would not be a big problem ( other than some wasted calculations), if not for the fact that somehow Eclipse gives an error if EQUIL or other initialization related keywords is specified in Restart runs.

joakim-hove:
> are the SKIPREST, SKIP100 and SKIP300 keywords supported by opm-parser at all?

The `SKIP100` and `SKIP300` keywords are not supported at all; it is on the loose-way-into-future plan to implement something like `ParseIdentity` so the parser can know: "Allright - now I am parsing as e.g. ECLIPSE100 and should behave accordingly" - then we can support `SKIP100` and `SKIP300`.

`SKIPREST` is a different beast entirely; I guess it's origins comes from the fact that (I assume) ECLIPSE really works works like an interpreter stepping through the input file one keyword at a time. When RESTART is active it should not "execute" the `TSTEP` or `DATES` keywords found in the Schedule section but rather skip forward until the restart point.

joakim-hove:
OK;

I guess this will be a tradeoff between strict ECLIPSE behavior and sanity. As I see it there are typically three different paths:

1. What the Eclipse documentation says.
2. What Eclipse actually does.
3. What would be the sane thing to do.

As I see it the main problem here is that the Eclipse input format is used to both configure the case as a whole - and one particular run using the `RESTART / SKIPREST` keywords. Given a fully specified input deck it should be possible to instruct the simulator as:

```bash
bash% simulator DECK.DATA --restart=/path/to/restart_file:70
```

*Without* actually modifying the input deck. We can easily (and should?!) support this behavior, but that will be an extension of Eclipse.

**Eclipse restart semantics:** To instruct Eclipse to restart a simulation you must go into the deck and make the following modifications:

1. In the `SOLUTION` section you must insert the `RESTART` keyword:
```txt
RESTART
    /path/to/case/INITCASE 20 /
```

[...]

PR #963, opened 2016-11-02
Ignore SKIPREST

jokva:
The SKIPREST keyword is really only needed in eclipse - since we parse
through the entire deck and create a complete internal representation
with random access, we don't need to actually skip parts of the input.

The old behaviour was to mimic some possible failure modes for eclipse,
however this has caused issues for no real benefit.

refs:
* https://github.com/OPM/opm-parser/issues/773
* https://github.com/OPM/opm-parser/issues/960

### DIFF
--- a/opm/parser/eclipse/EclipseState/InitConfig/InitConfig.cpp
+++ b/opm/parser/eclipse/EclipseState/InitConfig/InitConfig.cpp
@@ -17,6 +17,8 @@
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <iostream>
+
 #include <opm/parser/eclipse/Deck/Deck.hpp>
 #include <opm/parser/eclipse/Deck/DeckItem.hpp>
 #include <opm/parser/eclipse/Deck/DeckKeyword.hpp>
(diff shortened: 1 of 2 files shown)


andlaus:
stupid question (since I do not follow opm-parser development too closely nowadays): is `SKIP100 [...] ENDSKIP` supported by the current master? I'd assume that this is often used as an equivalent to C++'s `#if 0 [...] #endif`?

joakim-hove:
> is SKIP100 [...] ENDSKIP supported by the current master? 

No

>I'd assume that this is often used as an equivalent to C++'s #if 0 [...] #endif?

Yes; or more:
```C
#ifndef ECLIPSE100
...
...
#endif
```
there is also a `SKIP300` variety (and I think a plain `SKIP` which would be `#if 0`).

andlaus:
> Yes; or more: [...]
> there is also a SKIP300 variety (and I think a plain SKIP which would be #if 0). 

also, there's an unconditional SKIP. I've been thinking about it a bit: maybe it would be worth to add a "simulator mode" to opm-parser to support stuff like this, this would also allow to support keywords that have a different definition for different simulators. I suppose this would require some updates to the JSON files and some changes in the low and medium levels of the parser, though...

joakim-hove:
> I've been thinking about it a bit: maybe it would be worth to add a "simulator mode" to opm-parser to support stuff like this,

I have also thought of that, not tomorrow though ....

andlaus:
yeah, no immediate need. we should keep it in mind, though. (as we will hopefully climb up the "deployment ladder", it may become more important.)

PR #965, opened 2016-11-02
Support externally installed cJSON.

### LINKED ISSUE #964: cJSON headers get intermixed.
I installed a current version of cJSON on my system which is found by opm-parser. The new version changed the identifiers used for the type. Unfortunately, it seems like opm/json/JSonObject.cpp still uses opm/json/cjson/cJSON.h which results in an exception when createDefaultKeywordList is run.

blattms:
Using an external cJSON installed under /usr/include was not possible before
as the cJSON headers within opm were still used due to relative paths.
With this commit we move the copied cJSON source to external/cjson and
thus prevent them to be found if an externally installed cJSON is there.

Needs OPM/opm-common#199

Closes #964


joakim-hove:
Thank you - this looks fine and I am ready to merge this, but I will let @akva2 review OPM/opm-common#199 before merging.

PR #968, opened 2016-11-09
Minor fixes unrelated (warnings etc)

pgdr:
* Implemented missing method "name" in TableColumn.cpp
* add inline to method in RestartConfig to remove warning
* Whitespace changes


jokva:
Mergen when green.

PR #969, opened 2016-11-09
Value semantics for Well

jokva:
Modifies EclipseState and its dependencies so that the Well objects have value semantics, contributing towards EclipseState being copyable.

Notable changes to achieve this:

* The `WellSet` class has been removed, and replaced by a `std::set` of strings that denote well names.
* The `GroupTreeNode` class has been removed, and `GroupTree` is implemented in terms of an implicit tree of string-pairs for group-parent.
* shared_ptr members in `EclipseState` and its sub objects are replaced with their automatic cousins, save for `TableManager`.
* The source files given to the ahead-of-time compilation of `ParserKeywords` has been trimmed to its bare necessities.


joakim-hove:
jenkins build this with downstreams opm-output=133 opm-core=1101 please

jokva:
Breaking up this into smaller PRs.

PR #970, opened 2016-11-09
Fully internalized the JFUNC keyword

pgdr:
* Fully internalized the JFUNC keyword into class `JFuncTable.hpp`
* Added JFUNC flag, `getJFuncColumn()` and `getJFuncTable()` in TableManager
* added protected member m_jfunc to SimpleTable
* Added getJFuncColumn to accompany getPcowColumn
* Throws if pressure or jfunc is accessed inappropriately
  * ... meaning if getPcowColumn is called when JFUNC is in deck, or
  * getJFuncColumn is called when JFUNC is not in deck
* added tests for throwing and for getJFuncColumn
* In the event that one tries to get "PCOW" or "PCOG" via getColumn
  * ... this will throw if JFUNC is present in the deck.
* Added tests.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Tables/JFunc.cpp
+++ b/opm/parser/eclipse/EclipseState/Tables/JFunc.cpp
@@ -32,61 +32,61 @@ namespace Opm {
         const auto& rec = kw.getRecord(0);
         const auto& kw_flag = rec.getItem("FLAG").get<std::string>(0);
         if (kw_flag == "BOTH")
-            flag = Flag::BOTH;
+            m_flag = Flag::BOTH;
         else if (kw_flag == "WATER")
-            flag = Flag::WATER;
+            m_flag = Flag::WATER;
         else if (kw_flag == "GAS")
-            flag = Flag::GAS;
+            m_flag = Flag::GAS;
         else
             throw std::invalid_argument("Illegal JFUNC FLAG, must be BOTH, WATER, or GAS.  Was \"" + kw_flag + "\".");
 
-        if (flag != Flag::WATER)
-            goSurfaceTension = rec.getItem("GO_SURFACE_TENSION").get<double>(0);
+        if (m_flag != Flag::WATER)
+            m_goSurfaceTension = rec.getItem("GO_SURFACE_TENSION").get<double>(0);
 
-        if (flag != Flag::GAS)
-            owSurfaceTension = rec.getItem("OW_SURFACE_TENSION").get<double>(0);
+        if (m_flag != Flag::GAS)
+            m_owSurfaceTension = rec.getItem("OW_SURFACE_TENSION").get<double>(0);
 
-        alphaFactor = rec.getItem("ALPHA_FACTOR").get<double>(0);
-        betaFactor = rec.getItem("BETA_FACTOR").get<double>(0);
+        m_alphaFactor = rec.getItem("ALPHA_FACTOR").get<double>(0);
+        m_betaFactor = rec.getItem("BETA_FACTOR").get<double>(0);
 
         const auto kw_dir = rec.getItem("DIRECTION").get<std::string>(0);
         if (kw_dir == "XY")
-            direction = Direction::XY;
+            m_direction = Direction::XY;
         else if (kw_dir == "X")
-            direction = Direction::X;
+            m_direction = Direction::X;
         else if (kw_dir == "Y")
-            direction = Direction::Y;
+            m_direction = Direction::Y;
         else if (kw_dir == "Z")
-            direction = Direction::Z;
+            m_direction = Direction::Z;
         else
             throw std::invalid_argument("Illegal JFUNC DIRECTION, must be XY, X, Y, or Z.  Was \"" + kw_dir + "\".");
     }
 
-    double JFunc::getAlphaFactor() const {
-        return alphaFactor;
+    double JFunc::alphaFactor() const {
+        return m_alphaFactor;
     }
 
-    double JFunc::getBetaFactor() const {
-        return betaFactor;
+    double JFunc::betaFactor() const {
+        return m_betaFactor;
     }
 
-    double JFunc::getgoSurfaceTension() const {
-        if (flag == JFunc::Flag::WATER)
+    double JFunc::goSurfaceTension() const {
+        if (m_flag == JFunc::Flag::WATER)
             throw std::invalid_argument("Cannot get gas-oil with WATER JFUNC");
-        return goSurfaceTension;
+        return m_goSurfaceTension;
     }
 
-    double JFunc::getowSurfaceTension() const {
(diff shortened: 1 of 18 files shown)


pgdr:
WIP because I will make `FLAG` (wat,gas,both) and `DIRECTION` (xy,x,y,z) enums.

pgdr:
See #956 for a prior discussions around the JFUNC keyword.

opm/parser/eclipse/EclipseState/Tables/JFuncTable.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+namespace Opm {
+
+class DeckKeyword;
+
+class JFuncTable
+{
+public:
+
+    enum Flag {
```
joakim-hove: We should start using the "new" `enum class` consistently: http://www.cprogramming.com/c++11/c++11-nullptr-strongly-typed-enum-class.html

opm/parser/eclipse/EclipseState/Tables/JFuncTable.hpp
```diff
@@ -0,0 +1,115 @@
...
+        BOTH, WATER, GAS
+    };
+    enum Direction {
+        XY, X, Y, Z
+    };
+
+    inline void init(const DeckKeyword& kw) {
+        const auto& rec = kw.getRecord(0);
+        if (rec.hasItem("FLAG")) {
```
joakim-hove: As long as the `DeckKeyword` instance indeed corresponds to `JFUNC` keyword this test will *always* return true; and if it returns false that is not because a default has been applied - but rather because the wrong keyword has been passed in - which is throw material.

opm/parser/eclipse/EclipseState/Tables/JFuncTable.hpp
```diff
@@ -0,0 +1,115 @@
...
+    inline const Flag& getJFuncFlag() const {
+        return flag;
+    }
+
+    inline const Direction& getDirection() const {
+        return direction;
+    }
+
+private:
+    Flag      flag = BOTH;             // JFUNC flag: WATER, GAS, or BOTH.  Default BOTH
```
joakim-hove: I would prefer if these default values were removed - they are already specified in the Json configuration, and if the values are requested unconditionally from the deck we will correctly get either the default or the value entered explicitly.

opm/parser/eclipse/EclipseState/Tables/JFuncTable.hpp
```diff
@@ -0,0 +1,115 @@
...
+public:
+
+    enum Flag {
+        BOTH, WATER, GAS
+    };
+    enum Direction {
+        XY, X, Y, Z
+    };
+
+    inline void init(const DeckKeyword& kw) {
```
joakim-hove: Could maybe implement `operator bool( )` in the `Jfunc` class, then the instance could be constructed as part of the `TableManager` initialization list, and the `Jfunc` member instance could just evaluate to false if the `Deck` does not have the `JFUNC` keyword.

pgdr:
Thanks for the review, Joakim.  I've fixed all of the items, except the one where you request deleting the default values.

Hold off merging anyway until we synchronize efforts with Andreas.

pgdr:
Thanks for the review, Joakim.  I've fixed all of the items.

jokva:
A quick note about commit --fixup:

If you have multiple fixups all destined to be squashed into one commit you need to fixup into the same *base* commit, not the children, for autosquash to pick up on them.

pgdr:
Yeah, I just let emacs do the job. :)

opm/parser/eclipse/EclipseState/Tables/JFunc.cpp
```diff
@@ -0,0 +1,88 @@
...
+            return;
+        const auto& kw = *deck.getKeywordList<ParserKeywords::JFUNC>()[0];
+        const auto& rec = kw.getRecord(0);
+        const auto& kw_flag = rec.getItem("FLAG").get<std::string>(0);
+        if (kw_flag == "BOTH")
+            flag = Flag::BOTH;
+        else if (kw_flag == "WATER")
+            flag = Flag::WATER;
+        else if (kw_flag == "GAS")
+            flag = Flag::GAS;
```
pgdr: in `JFUNC`, `flag` is optional and defaulted to `BOTH`.  Meaning that if it is not specified in deck, through the `JFUNC` json, it will be given as `BOTH`.

There is a potential issue here, namely if a user specifies `JFUNC GARBAGE/` in the deck.  Then we might reach a state where we read an uninitialized `flag`.

I could add a final `else throw unrecognized flag GARBAGE` if you'd like.

pgdr: I'm now throwing on garbage `FLAG` or `DIRECTION`.

joakim-hove: I guess throwing on invalid FLAG / DIRECTION is a good idea.

pgdr:
@andlaus ping

andlaus:
pong: I had no time for this this week. (have been busy with making flow_ebos merge-ready.)

pgdr:
@andlaus ping

andlaus:
pong. on it: probably next week. (if you want to speed things up, you may also do the opm-material part.)

pgdr:
@andlaus ping

(This will probably be merged pretty soon.)

andlaus:
> (This will probably be merged pretty soon.)

if `flow_ebos` (and `flow_legacy`) continue to work for Model 2.2 after the merge, I don't have any objections.

andlaus:
@pgdr: could you please split the JFUNC-keyword internalization part into a separate PR where the tables are left as they are? I think that the JFUNC internalization is a good thing to have, whereas I'm not so sure about the tables.

pgdr:
Why are you not so sure about the tables?

andlaus:
I'm not sure on whether the table stuff is worth it because adapting the opm-material side to the table semantics is quite a bit of non-obvious work (as you've already discovered yourself?), most likely it does not make the resulting code easier to understand and the result may be even a bit slower than the current hack. Finally, it makes the semantics on the opm-parser side more complicated. (a.k.a. "when am I allowed to call getPcwoColumn() again?")

anyway, the JFUNC internalization is uncontroversial, so let's do this step-by-step.

pgdr:
Third time's the charm.

We do now not throw on incorrect reading (or lack) of JFUNC column.  This will be reversed later when opm-material is updated.

pgdr:
@andlaus @jokva @joakim-hove I'm merging this tomorrow if I don't hear anything from you.

jokva:
Go for it.

andlaus:
go ahead. I'll convert opm-material to the internalized version of the keyword shortly thereafter.

PR #974, opened 2016-11-11
Only give needed source files to createKeywordList

jokva:
The ahead-of-time JSON-to-C++-object compilation step does not need all
the source files as it was passed. Reduces the set of source files
compiled into the createKeywordList binary to just the files it needs.

### DIFF
--- a/opm/parser/eclipse/CMakeLists.txt
+++ b/opm/parser/eclipse/CMakeLists.txt
@@ -50,14 +50,29 @@ set( generator_source
 Generator/KeywordGenerator.cpp
 Generator/KeywordLoader.cpp )
 
-set( build_parser_source 
-Parser/ParseContext.cpp
+set( utility_source
+Utility/Functional.cpp
+Utility/Stringview.cpp
+)
+
+set( build_parser_source
+Deck/Deck.cpp
+Deck/DeckItem.cpp
+Deck/DeckKeyword.cpp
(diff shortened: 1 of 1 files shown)


opm/parser/eclipse/CMakeLists.txt
```diff
@@ -276,18 +286,13 @@ Utility/Functional.hpp
...
 # This section manages the generation of C++ code for the default keywords.
 
 # 1. Create an executable 'createDefaultKeywordList'.
 
-add_executable( createDefaultKeywordList Parser/createDefaultKeywordList.cpp )
-target_link_libraries( createDefaultKeywordList buildParser opmjson ${Boost_LIBRARIES})
-
-
+add_executable(createDefaultKeywordList Parser/createDefaultKeywordList.cpp ${build_parser_source})
+target_link_libraries( createDefaultKeywordList opmjson ${Boost_LIBRARIES} ${ERT_LIBRARIES})
```
joakim-hove: Could probably remove the `${ERT_LIBRARIES}` from here as well?

jokva: No, it still uses a `util_call` somewhere. I tried, was disappointed, re-added and honestly figure that I think it's possible to remove it.

The entire AOT compilation could use an overhaul, this is just the lowest hanging fruit.

PR #975, opened 2016-11-11
Remove shared ptr members well

jokva:
Simply flatten the Well class by removing all internal shared_ptrs. PR also includes the following minor changes:

* A bugfix in DynamicState, making the `find` method const
* TimeMap is not a member in `Well` as only its size was used, which is what we're now storing.


jokva:
Cycle to restart appveyor.

jokva:
jenkins build this please

joakim-hove:
This is ready to be merged.

PR #976, opened 2016-11-11
Replace explicit GroupTree with ordered strings

jokva:
Replaces The GroupTree + GroupTreeNode classes building an explicit tree
of named nodes with a sorted vector of { name, parent-name } pairs that
builds an implicit tree. Provides the same semantics as the previous
group tree implementation, but with less code and copying for free.


jokva:
jenkins build this please

joakim-hove:
jenkins build this with downstreams opm-core=1101 please

PR #977, opened 2016-11-11
DeckItem.push_back overload for size = 1

jokva:
DeckItem.push_back using insert for 1 element turns out to be very
expensive and caused a performance degredation.


jokva:
jenkins build this please

joakim-hove:
This is ready to be merged.

PR #979, opened 2016-11-16
Replace WellSet pointers with set-of-names

jokva:
The WellSet class is replaced by std::set, allowing predictable copy
semantics of Group. This change has rather few consequences as accessing
a well through a wellset was hardly ever done.

Getting hold of the corresponding well instance will now have to be done
via a Schedule instance - however, this simplifies the dependency graph
by severing the edge between the Group object and Well objects.


joakim-hove:
jenkins build this with downstreams opm-output=133 please

opm/parser/eclipse/EclipseState/Schedule/tests/WellSetTests.cpp
```diff
@@ -1,101 +0,0 @@
...
-    auto timeMap = createXDaysTimeMap(10);
-
-    auto well1 = std::make_shared< Well >("WELL", 0, 0, Opm::Value<double>("REF_DEPTH"), Opm::Phase::OIL, timeMap, 0);
-    auto well2 = std::make_shared< Well >("WELL", 0, 0, Opm::Value<double>("REF_DEPTH"), Opm::Phase::OIL, timeMap, 0);
-
-    wellSet.addWell( well1.get() );
-    BOOST_CHECK_EQUAL(true , wellSet.hasWell("WELL"));
-
-    BOOST_CHECK_NO_THROW( wellSet.addWell( well1.get() ));
-    BOOST_CHECK_THROW( wellSet.addWell( well2.get() ) , std::invalid_argument );
```
jokva: Yes - multiple additions of the same name will be no-ops.

opm/parser/eclipse/EclipseState/Schedule/tests/WellSetTests.cpp
```diff
@@ -1,101 +0,0 @@
...
-    BOOST_CHECK_EQUAL(1U , wellSet.size());
-    BOOST_CHECK_EQUAL( well2.get(), wellSet.getWell("WELL2"));
-}
-
-
-BOOST_AUTO_TEST_CASE(AddWellSameName) {
-    Opm::WellSet wellSet;
-    auto timeMap = createXDaysTimeMap(10);
-
-    auto well1 = std::make_shared< Well >("WELL", 0, 0, Opm::Value<double>("REF_DEPTH"), Opm::Phase::OIL, timeMap, 0);
```
joakim-hove: Dummy comment.

PR #982, opened 2016-11-17
Simplified DynamicVector implementation

jokva:
Topic says it all, really, just a simplification of the implementation of DynamicVector. The old implementation was largely oriented around storing `shared_ptr`s, which doesn't apply as much anymore.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/DynamicVector.hpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/DynamicVector.hpp
@@ -17,14 +17,11 @@
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-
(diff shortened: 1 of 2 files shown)


opm/parser/eclipse/EclipseState/Schedule/DynamicVector.hpp
```diff
@@ -33,65 +30,37 @@ namespace Opm {
...
-            if (index >= m_data.size())
-                m_data.resize( index + 1 , m_defaultValue);
-
-            return m_data[index];
+            return this->m_data.at( index );
         }
 
         void iset(size_t index, T value) {
-            (*this)[index] = value;
+            (*this)[index] = std::move( value );
```
pgdr: this is now an API change, right?  How far away from parser is `DynamicVector.iset` used?

PR #983, opened 2016-11-17
Changes in Tabdims construction:

joakim-hove:
1. Removed Tabdims(int,int,int, ....) constructor and added
    Tabdims(Deck) constructor.

 2. Added Tabdims member to Runspec( ) object.

 3. Changed std_shared_ptr<Tabdims> to Tabdims member in TableManager.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Runspec.cpp
+++ b/opm/parser/eclipse/EclipseState/Runspec.cpp
@@ -59,16 +59,21 @@ size_t Phases::size() const noexcept {
     return this->bits.count();
 }
 
-Runspec::Runspec( const Deck& deck ) :
-    Runspec( Phases{ deck.hasKeyword( "OIL" ),
-                     deck.hasKeyword( "GAS" ),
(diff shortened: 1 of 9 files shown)


opm/parser/eclipse/EclipseState/Runspec.cpp
```diff
@@ -59,16 +59,21 @@ size_t Phases::size() const noexcept {
...
 
+
 Runspec::Runspec( const Deck& deck ) :
-    Runspec( Phases{ deck.hasKeyword( "OIL" ),
-                     deck.hasKeyword( "GAS" ),
-                     deck.hasKeyword( "WATER" ) } )
+    active_phases( Phases{ deck.hasKeyword( "OIL" ),
+                           deck.hasKeyword( "GAS" ),
+                           deck.hasKeyword( "WATER" ) } ),
+    m_tabdims( Tabdims(deck) )
```
jokva: You can actually pass deck directly to `m_tabdims` which should be sufficient.

PR #990, opened 2016-11-22
Added functionality to "fix" ZCORN problems.

joakim-hove:
Crazy crazy ...

### DIFF
--- a/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
+++ b/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
@@ -698,10 +698,16 @@ namespace Opm {
         ecl_grid_init_coord_data_double( c_ptr() , coord.data() );
     }
 
-    void EclipseGrid::exportZCORN( std::vector<double>& zcorn) const {
+    size_t EclipseGrid::exportZCORN( std::vector<double>& zcorn) const {
+        ZcornMapper mapper( getNX(), getNY(), getNZ());
+
         zcorn.resize( ecl_grid_get_zcorn_size( c_ptr() ));
         ecl_grid_init_zcorn_data_double( c_ptr() , zcorn.data() );
+
+        return mapper.fixupZCORN( zcorn );
     }
+
+
 
     const std::vector<int>& EclipseGrid::getActiveMap() const {
         if( !this->activeMap.empty() ) return this->activeMap;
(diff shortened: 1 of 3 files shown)


opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
```diff
@@ -766,6 +773,36 @@ namespace Opm {
...
+
+    void ZcornMapper::fixupZCORN( std::vector<double>& zcorn) {
+        int sign = zcorn[ this->index(0,0,0,0) ] < zcorn[this->index(0,0, this->dims[2] - 1,0)] ? 1 : -1;
+
+        for (size_t k=0; k < this->dims[2]; k++)
+            for (size_t j=0; j < this->dims[1]; j++)
+                for (size_t i=0; i < this->dims[0]; i++) {
+                    this->fixupZCORN( zcorn , sign, i , j , k , 0 );
+
+                    if (i == (this->dims[0] - 1))
```
atgeirr: I don't understand why the if() is here? Should we not do all the fixupZCORN(...., c) calls unconditionally for c = 0..3?

joakim-hove: Yes - you are right. The pillars, i.e. `COORD` are shared between cells, but not the z values.

joakim-hove:
OK - I have pushed an update to the inner loop. When it comes to return values and warnings I will do as you wish; but my suggestion is:

1. The `EclipseGrid::exportZCORN( )` method will return an integer - saying how many *cells* were modified.
2. The calling scope checks the return value and issues a warning if ` != 0`?

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
```diff
@@ -766,6 +773,36 @@ namespace Opm {
...
+    void ZcornMapper::fixupZCORN( std::vector<double>& zcorn, int sign, size_t i, size_t j, size_t k , size_t c) {
+        size_t index1 = this->index(i,j,k,c);
+        size_t index2 = this->index(i,j,k,c+4);
+
+        if ((zcorn[index2] - zcorn[index1]) * sign < 0 )
+            zcorn[index2] = zcorn[index1];
+
+    }
+
+    void ZcornMapper::fixupZCORN( std::vector<double>& zcorn) {
```
atgeirr: I think points might be the most interesting, since that would correspond to the number of "wrong" zcorn values in the input. Although I would expect it to be divisible by 4 quite often...

atgeirr:
> The calling scope checks the return value and issues a warning if != 0?

Sounds good to me!

joakim-hove:
OK - then I consider this ready.

joakim-hove:
Test failure is `test_stoppedwells` - I have problems seeing that this is related?

atgeirr:
I had a look at the failure, and before the "fix" the zcorn are 400 * 10 and then 400 * 30. After it is 800 * 10, so there is clearly a bug.

opm/parser/eclipse/EclipseState/Grid/EclipseGrid.cpp
```diff
@@ -766,6 +772,32 @@ namespace Opm {
...
+            zcorn[index2] = zcorn[index1];
+            return 1;
+        } else
+            return 0;
+
+    }
+
+
+    size_t ZcornMapper::fixupZCORN( std::vector<double>& zcorn) {
+        int sign = zcorn[ this->index(0,0,0,0) ] < zcorn[this->index(0,0, this->dims[2] - 1,0)] ? 1 : -1;
```
atgeirr: This is the bug. It looks at the top and bottom cells' top coordinates, but if there is only one layer this is the same cell, and sign becomes -1.

To fix it I tested with replacing the final 0 with a 4 (comparing to the bottom cell), that should fix all trivial cartesian cases, although for a corner-point grid there is no requirement that this should work.

However if we also replace the < with a <= I think we will probably never encounter a problematic grid, unless designed to beat this test! In summary, I suggest the line:

```
        int sign = zcorn[ this->index(0,0,0,0) ] <= zcorn[this->index(0,0, this->dims[2] - 1,4)] ? 1 : -1;
```

joakim-hove:
OK - I have updated according to your suggestion.

joakim-hove:
@atgeirr - please merge when happy. (No - you are not getting braces 😀)

atgeirr:
Now we just have to find the call sites and add some warning/logging there, but that's for another day.

PR #992, opened 2016-11-29
Update jenkins downstream list

akva2:
opm-grid moved in the dependency graph. also has a commit that removes an unused variable i now noticed. i can remove if you want.


akva2:
since this hinders me from doing updates for the things which require opm-data stuff, which is very nice to do at these quiet hours, i'm self-merging this one without ack to restore ci operation.

PR #993, opened 2016-11-30
Remagine DynamicState

jokva:
This PR fixes a couple of bugs unearthed by the main effort - simplify the `DynamicState` implementation (or rather multiple aspects of it).

The *big* change here is the last commit:
> Changes DynamicState's implementation to, instead of maintaining a
> current and initial element, rely on std::vector's capabilities. This
> introduces a small change in semantics.
> 
> The new DynamicState will up-front allocate a vector of TimeMap.size,
> and all entries will be the initial value. Lookup is then simplified to
> vector.at. DynamicState.update will assgin to the given index and *all
> consecutive elements* which will enforce the same invariant as before.
>
> The behavorial change is that DynamicState will no longer throw on
> update(x); update(y); where y < x. Instead, the resulting effect will be
> that update(x) was never called.
>
> This means that the Group.setInjectionPhase function behaves slightly
> differently, because DynamicState.size is non-sensical. If multiple
> distinct injection phases are specified for a group inside the same time
> step then the last specified phase will be used. The comment has been
> updated accordingly.

These new semantics may not be acceptable (I think they're an improvement, YMMV).

Performance wise this seems like a slight improvement on Heidrun, Norne seems unaffected.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/WellInjectionProperties.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/WellInjectionProperties.cpp
@@ -18,6 +18,7 @@
 */
 #include <opm/parser/eclipse/EclipseState/Schedule/WellInjectionProperties.hpp>
 
+#include <iostream>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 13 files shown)


joakim-hove:
Thank you for a good PR description. As you say this implies a minor change in semantics - the existing behavior was intended to mimic the 'Eclipse Interpreter State' very closely, but the extra protection in the throw has not proved important. Will think about it.

jokva:
Will Eclipse fail and bail out of these assumptions are violated? Is a non-decreasing insertion order even possible from the .DATA file?

If both answers are no it doesn't seem like it's even possible to get the ordering wrong (then again, I might have missed something and might be *very* wrong here).

joakim-hove:
No - you are right, with Eclipse it is not even possible to get this wrong. That constraint was what we were mimicking with the original implementasjon, can probably be relaxed.

opm/parser/eclipse/EclipseState/Schedule/WellInjectionProperties.cpp
```diff
@@ -54,4 +55,20 @@ namespace Opm {
     bool WellInjectionProperties::operator!=(const WellInjectionProperties& other) const {
         return !(*this == other);
     }
+
+    std::ostream& operator<<( std::ostream& stream,
```
joakim-hove: Is  `operator<<` really required? When / if we decide to support output from the parser state I would like a slightly coherent discussion on the topic, so that we agree on one way to do it.

jokva: Sort-of. It's useful for unit tests (makes BOOST_CHECK_EQUAL work) and possibly parser state output later. For BOOST_CHECK_EQUAL it's really only important that it *exists* whenever well properties are compared (e.g. the unit test file), but I figured it's just as nice to provide it generally. What it outputs really doesn't matter that much, but I have found them useful for debugging in the past.

PR #994, opened 2016-11-30
Register completion number from COMPDAT

jokva:
Connections are given a completion number starting at 1 implicitly when
they're first defined by the COMPDAT keyword. Until now this number has
been ignored.

The number itself isn't *used* for anything with this patch, just stored
and accessible.


jokva:
By the way, this is a part of the ongoing COMPLUMP effort.

joakim-hove:
This seems perfectly innocent, but be aware og the COMPORD keyword which might be relevant when supporting COMPLUMP?

joakim-hove:
Feel free to merge.

jokva:
This one should be innocent in the sense that what is provides isn't in use yet.

I'll keep COMPORD in mind.

PR #996, opened 2016-12-01
Completion number aware WELOPEN.

jokva:
No longer chokes when WELOPEN is given its last two parameters, C1 and
C2 for completion number begin/end. Instead, these now weigh in on
whether or not a well matches the criteria for opening or closing.

Includes some test fixes and some restructuring of handleWELOPEN.
Obsolete verifications of throwing on COMPLUMP have been removed.

Relies on (or supersedes, depending) https://github.com/OPM/opm-parser/pull/995

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
@@ -186,7 +186,7 @@ namespace Opm {
                 handleCOMPSEGS(keyword, currentStep);
 
             else if (keyword.name() == "WELOPEN")
-                handleWELOPEN(keyword, currentStep , section.hasKeyword("COMPLUMP"));
+                handleWELOPEN(keyword, currentStep);
 
             else if (keyword.name() == "WELTARG")
                 handleWELTARG(section, keyword, currentStep);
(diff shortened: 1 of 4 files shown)


opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -735,98 +735,95 @@ namespace Opm {
...
+
+            const auto status = WellCompletion::StateEnumFromString( status_str );
+
+            /*
+             * Construct the updated completion with the possible status change
+             * applied. Status is changed when a completion matches all the IJK
+             * criteria *and* the completion number. A defaulted field always
+             * matches the property in question.
+             */
+            auto updated = [=]( const Completion& completion ) -> Completion {
```
joakim-hove: If this function accepted a `DeckRecord& record` argument I think it should be possible to rewrite the whole function using just the `DeckItem.defaultApplied( )` method - i.e. without going temporarily through the `-1`value as default indicator.

joakim-hove: The name `updated` is in my opinion a bit misleading - the main part of the function actually implements a `matching` functionality. One alternative could even be to create the updated completion in calling scope:
```
bool matcing = []( const DeckRecord& record, const Completion& completion) {
    return true || false;
}

...

for( const auto& c : well->getCompletions( currentStep ) ) {
      if (matching(rec,c))
         completions.add( { c, new_status } );
     else  
         completions.add( c )
}
```

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -735,98 +735,95 @@ namespace Opm {
...
+                if( !defaulted( C2 ) && complnum > C2 ) return completion;
+                if( !defaulted( C1 ) && !defaulted( C2 )
+                    && ( C1 > complnum || complnum > C2 ) ) return completion;
+
+                // completion matched - update it's status
+                return { completion, status };
+            };
+
+            for( auto* well : getWells( wellname ) ) {
+                CompletionSet completions;
```
jokva: Yeah, no problem.

opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
jokva: I did that at first, but you still need to get the value (I added the `defaulted` function for readability more than anything), and I figured it was clearer what was *actually* going on this way. A bit less clunky since looking up individual items from a `DeckRecord` is somewhat clunky. Oh, and the record is reused for many completions.

jokva: That being said it can surely be changed. The function is moved and reused in https://github.com/OPM/opm-parser/pull/997 however.

jokva: Yeah, internally it's a matching, but when viewed as a black box it *always* returns a new Completion (possibly different from its input argument). In that sense what you get is a completion updated according to the record.

jokva: I see I used the name `update` in context of complump (which cares more imperative and oomph to it in my opinion). I still like the name though because that's precisely what the function does - it takes a completion and returns a fresh, updated one.

jokva: Is that an interesting distinction? (I obviously don't consider it that).

jokva:
Superseded by #997

PR #997, opened 2016-12-02
COMPLUMP support

jokva:
Support for COMPLUMP. a keyword to re-assign completion numbers that can
be used with keywords such as WELOPEN.

Some helper functions were moved out of handleWELOPEN and reused in
handleCOMPLUMP.

Built on top of https://github.com/OPM/opm-parser/pull/995 and https://github.com/OPM/opm-parser/pull/996


joakim-hove:
OK - this is mainly #996 - I had some comments to #996, which still apply here - although to a much lesser extent. I suggest #996 is closed, and you address some nit-picks in this one for good measure before everything is merged?

The travis failure is due to timeout, which is unfortunate - but can not be attributed to this PR.

jokva:
I'll apply some soft re-naming tomorrow and re-assess the other things.

PR #1000, opened 2016-12-04
ZCORN adjustments considers cell-cell relation.

joakim-hove:
In #990 we adjust the ZCORN values if a cell is twisted. With this PR we also adjust the ZCORN values if the z values of two cells come out of order.


atgeirr:
Looks good to my eyes (not compile- or run-tested).

jokva:
Approv'd.

atgeirr:
One thousand PRs/issues! That is a milestone to both fear and take pride in...!

PR #1001, opened 2016-12-05
Add some tests for the unit constants.

atgeirr:
These have been copied from opm-core, where they are scheduled to be removed (since the unit handling moved to opm-parser).


atgeirr:
The Travis failure seems to be due to timeout, further down in the chain. I'll start a Jenkins job just to be sure.

atgeirr:
jenkins build this please

joakim-hove:
> The Travis failure seems to be due to timeout, further down in the chain.

Yes  - that is very unfortunate; the efforts to keep us within the travis build limits did not last very long :disappointed:

atgeirr:
> the efforts to keep us within the travis build limits did not last very long

We'll see the usefulness of Travis decline as we add more code and tests, and hit this limit as a matter of course. We do have Jenkins, though!

alfbr:
We have even disabled opm-upscaling, and ewoms seems rather entrenched at the moment. Travis is far more expensive than Jenkins as a paid service, so this may very well mark the end of the road for Jenkins.

Issue #973, opened 2016-11-11
Problems with localization


andlaus:
there seem to be some problems in the interactions of boost and locales that sometimes break the build or cause the unit tests to fail. This is particularly annoying because it seems to happen frequently: it seems like the issue can be triggered using a freshly installed Ubuntu that was set up with a non-default locale.

Further reading: #971

jokva:
We've now merged #1002, but this issue is largely a wontfix. Here's why (for future reference):

* The problem arises because boost pre-1.57 calls `std::locale("")` which yields the user preferred locale. On linux this seems to be governed by LC_ALL and other LC_* and the LANG environment variables.
* There's no good way to override this without going platform specific, e.g. check if `std::locale("")` throws and then override user environment to a fallback encoding (C) if unix or whatever you need to do on windows. This is in my opinion a *clear* violation of what a library is supposed to do if it's permanent, and such a hassle if it's on a per-function call. It's also a work-around for the users fundamentally broken environment, and if we're to go lengths to handle all permutations of broken user envs we're in for a ride. Flow and other applications are obviously free to make this choice, which is a nice user friendly feature, but it's not for opm-parser to do.
* The issue seems to be way less of a problem with boost >= 1.57, so whenever we either require a newer boost this probably goes away.

atgeirr:
We cannot handle all permutations of broken environments, but if important cases such as a near-default setup of Ubuntu breaks we should be prepared to handle it gracefully. I do not quite understand what is "broken" with the user environment in these cases though.

Is the boost thing a bug in some sense? If so, we should require newer boost to avoid it if that solves everything.

jokva:
> I do not quite understand what is "broken" with the user environment in these cases though.
Usually language being set to some locale that isn't generated. Using a richer installer, installing US or forcing C locale always solves this problem.

> Is the boost thing a bug in some sense? If so, we should require newer boost to avoid it if that solves everything.

Yes, but RH7 ships with 1.53.

PR #1002, opened 2016-12-05
Don't crash keyword compilation on broken locales

jokva:
Work around a weakness of older boosts by forcing C locale if
std::locale throws.

The real solution is to not have broken locales, but this should make it
slightly easier to get the JSON keywords compiled.

Addresses https://github.com/OPM/opm-parser/issues/973

Please note that this does not fix unit tests, it does not fix opmi and it does not fix flow, because it's really up to the driver application to handle this.

A sufficiently new boost, properly configured locales or `LC_ALL="C" make` should also solve this problem.

### DIFF
--- a/opm/parser/eclipse/Parser/createDefaultKeywordList.cpp
+++ b/opm/parser/eclipse/Parser/createDefaultKeywordList.cpp
@@ -16,11 +16,19 @@
   You should have received a copy of the GNU General Public License
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
+
+#if !defined(WIN32)
+    #define _POSIX_C_SOURCE 200112L
+    #include <stdlib.h>
+#endif
+
 #include <iostream>
+#include <locale>
+
+#include <boost/filesystem.hpp>
 
 #include <opm/parser/eclipse/Generator/KeywordGenerator.hpp>
 #include <opm/parser/eclipse/Generator/KeywordLoader.hpp>
-
 
 
 int main(int argc , char ** argv) {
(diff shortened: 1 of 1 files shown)


opm/parser/eclipse/Parser/createDefaultKeywordList.cpp
```diff
@@ -34,6 +42,33 @@ int main(int argc , char ** argv) {
...
+             * 1.56 uses the std::locale("") constructor which respects user
+             * preferred locales, and might crash. If this is the case, and
+             * we're on a non-windows system (assuming POSIX), in case of an
+             * exception, set the environment to "C" and keep going.
+             *
+             * Can be removed once boost < 1.57 is no longer supported
+             */
+            std::locale( "" );
+        } catch( const std::runtime_error& ) {
+            #if !defined(WIN32)
```
andlaus: since the code in the catch() block is required anyway, would't it be better to always use this and throw away the `try..catch` as well as the output below?

jokva: No, because we don't want to override locales if they aren't broken.

opm/parser/eclipse/Parser/createDefaultKeywordList.cpp
```diff
@@ -34,6 +42,33 @@ int main(int argc , char ** argv) {
...
+            setenv( "LC_ALL", "C", 1 );
```
andlaus: is this required? as I interpret

http://en.cppreference.com/w/cpp/locale/locale/classic

`std::locale::classic()` does not depend on any environment variables. maybe I overinterpret it, though...

andlaus:
cool, thanks for the PR!

opm/parser/eclipse/Parser/createDefaultKeywordList.cpp
jokva: It doesn't, and in fact that's how every C++ program is started (by implicitly setting locale == classic). However, that isn't the actual problem here. The problem is that boost calls `std::locale("")` which ignores any locale setting (setting locale with `std::locale` only affects what's returned by `std::locale()`, the default constructor, and there's no (standard) way of changing what's yielded by `std::locale("")`, which on linux picks up on the environment. As long as boost keeps calling `std::locale("")` the environment must either be sane *or* we need to override it, and that is inherently unportable.

opm/parser/eclipse/Parser/createDefaultKeywordList.cpp
andlaus: why not? In this context, are there any advantages if boost tries to be Xhosa instead of English?

opm/parser/eclipse/Parser/createDefaultKeywordList.cpp
andlaus: ok, I checked: the envvar seems to be necessary, but on my broken system, the two lines below it (`boost::...::imbue(std::locale::classic())`) don't seem to have any effect...

jokva: `boost::...::imbue` only sets the locale boost prefers, but it will still call the problematic `std::locale("")`.

andlaus: yeah, my point is rather that the effect of your somewhat more complicated solution is the same as

```c++
#if !defined(WIN32)
    setenv( "LC_ALL", "C", 1 );
#endif
```

right?

PR #1003, opened 2016-12-06
TableManager.getPvtwTable

jokva:
Provide a richer interface to read the PVTW table.


jokva:
Downstream: https://github.com/OPM/opm-core/pull/1119

joakim-hove:
This does not belong with `InirConfig` - only the *implementation* resembles `EQUIL` - not the reservoir content.

jokva:
Alright. It's a table, then, I presume?

joakim-hove:
Yes - I guess table is the best match.

jokva:
Updated. I'll re-push core as well.

jokva:
Restarting CI.

PR #1005, opened 2016-12-09
Do not use ParserKeyword-derived definitions.

jokva:
These definitions are hardly used and represent a significant portion of
build times. By removing them and using their string cousins instead,
which are less safe but makes the code compile faster, we also open up
for changing the form of the auto-generated code itself.

### DIFF
--- a/opm/parser/eclipse/Deck/DeckKeyword.hpp
+++ b/opm/parser/eclipse/Deck/DeckKeyword.hpp
@@ -56,12 +56,8 @@ namespace Opm {
         const std::vector<std::string>& getStringData() const;
         size_t getDataSize() const;
 
-        template <class Keyword>
-        bool isKeyword() const {
-            if (Keyword::keywordName == m_keywordName)
-                return true;
-            else
-                return false;
+        bool isKeyword( const std::string& nm ) const {
+            return nm == this->m_keywordName;
         }
 
         const_iterator begin() const;
--- a/opm/parser/eclipse/EclipseState/EclipseState.cpp
+++ b/opm/parser/eclipse/EclipseState/EclipseState.cpp
@@ -246,14 +246,13 @@ namespace Opm {
     }
 
     void EclipseState::applyModifierDeck(const Deck& deck) {
-        using namespace ParserKeywords;
         for (const auto& keyword : deck) {
 
-            if (keyword.isKeyword<MULTFLT>()) {
+            if (keyword.isKeyword( "MULTFLT" )) {
                 for (const auto& record : keyword) {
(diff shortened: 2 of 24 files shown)


bska:
> These definitions are hardly used

That is probably true, but they are not **unused** outside of opm-parser.  We do, for instance, use `hasKeyword<kw>()` in opm-upscaling's [`RelPermUtils.cpp`](https://github.com/OPM/opm-upscaling/blob/master/opm/upscaling/RelPermUtils.cpp#L396-L400)

bska:
> Then we've found a good opportunity to port this Deck usage to `EclipseState`.

Except we can't.  Module opm-upscaling is explicitly required to be able to run on subsets of ECL keywords.  In particular, we cannot expect there to be any sectioning keywords so there is not enough information available to build the state machine.

jokva:
Will there be enough to build the subsections you care about? What's used?

bska:
> Will there be enough to build the subsections you care about?

No sure.  There is no `RUNSPEC`, for instance.

> What's used?

Off the top of my head, subsets of `GRID` (especially `COORD`/`ZCORN`, `ACTNUM`, `PORO`, `PERM*`), `REGIONS` (especially `SATNUM`) and `PROPS` (saturation functions).  I'll have to look more closely to verify.

jokva:
We can probably build all of that without exposing Deck.

bska:
> We can probably build all of that without exposing `Deck`.

Right.  Can we also do that for subsets of cells?  We need that ability for the [`CornerpointChopper`](https://github.com/OPM/opm-upscaling/blob/master/opm/upscaling/CornerpointChopper.hpp#L259)

jokva:
Don't know yet, maybe. Create an issue and we'll have a look.

joakim-hove:
This in it's own right is of course OK. However the loss of the compiled in default values I think is a no-go.

jokva:
Even with its only 5 occurences?

(If yes, I can try to figure out a way to keep some default values without incurring massive compilation times).

joakim-hove:
> Even with its only 5 occurences?

Well - I agree that five occurences is not much; but I consider those uses as very valuable - and if anything would like increase them. That being said the defaults could be made available in a completely different way; e.g. the code for a lookup ~singleton resembling `std::map<std::string,<std::map<std::string,T>>` could be assembled during build?

jokva:
Something like that, yeah, I'd have to try a bit to figure out something that's both performant and ergonomic, but it's the gist of it for sure.

I *was* pretty bummed seeing that changing from a type-register scheme would mean that the default values went away, but I think we can bring them back.

(this entire PR should be considered a proof-of-concept).

joakim-hove:

PR #1007, opened 2016-12-12
Integrate Deck into Parser directory.

jokva:
This patch accomplishes two things:

1. Merges all Deck header and source files into Deck.cpp and Deck.hpp.
   This gives us a few fewer files to keep track of and include, and
   reduces the number of compiling objects accordingly. The cost is
   larger files (~300 header, ~700 source). Moves the Deck into Parser
   directory, effectively reducing the number of parallel source trees
   by one.
2. Merges all Deck tests into one DeckTests.cpp under Parser/tests,
   further reducing temporary object files.

All other tests and includes are updated accordingly, and headers are
fixed if they broke (as the new Deck includes less).

This is intended as an experiment and seeing if this is a good idea or not. For this reason discussion is particularly welcome. I'll upload measurements w.r.t. compile times soon.


jokva:
Compile times:
* master: 16m8s
* pr: 15m30s

This is obviously unscientific (time make -j8; read user)

So approx. 38s for these changes - I assume we can expect just as much returned for every similar set of changes.

jokva:
> Are those compile times for all opm modules?

No, just for parser.

> Does this PR change encapsulation of its own headers in a good way? (I.e. meaning fewer recompiles downstream, just relinking)?

Working on it (want it for ourselves too, obviously), but it's a partial goal.

jokva:
Just to make it clear: this PR doesn't really affect any encapsulation, but I expect that if it gets approved (and I get to work on similar merges) it lays the groundwork for slimmer public headers and moving significant portions into private headers and better encapsulation.

jokva:
I'd also like to add that the only interesting changes in the patch itself are in Parser/Deck.cpp, Parser/Deck.hpp (the final, merged files) and CMakeLists.txt. Other than that this PR should be about the concept of reducing granularity in our modules with broader, wider-reaching header and fewer compilation targets.

joakim-hove:
> Does this PR change encapsulation of its own headers in a good way?

I though that had improved quite a lot already.

atgeirr:
> I though that had improved quite a lot already.

I think so, too! That does not remove it from the list of things I care about though...

joakim-hove:
> For this reason discussion is particularly welcome

Well - I am in general skeptical to all changes which hurt readability, I fear this is moving in that direction.

1. How will compilation time be affected if we retain more files and use a simple `#include "Deckxxx.cpp"` scheme?

2. I would prefer to keep the deck in a separate directory; the `Opm::Parser` class is very much public - whereas we would like to make the `Opm::Deck` module private.

jokva:
> How will compilation time be affected if we retain more files and use a simple #include "Deckxxx.cpp" scheme?

I'd assume largely the same as with this patch (but we're now moving build system work to the preprocessor)

> I would prefer to keep the deck in a separate directory; the Opm::Parser class is very much public - whereas we would like to make the Opm::Deck module private.

Fair. I don't really feel it's warranted to give the one file its own directory though, but I can see that Parser isn't the right place for it.

atgeirr:
Collecting many classes in a single file is hard on us old-timers who tend to think one file == one class.

Also, given that the time reduction was only on the order of a few percentage points, I don't think that warrants the collecting (also tends to make make -j 8 less efficient as it serializes a little). (Also I am really grateful for my SSD that lets me compile all modules in about 10-12 minutes, combined...)

I *do* however appreciate moving towards better encapsulation of headers, but that does not necessarily require combining them in huge cpp files.

jokva:
> Collecting many classes in a single file is hard on us old-timers who tend to think one file == one class.

You can think of it as one file = one module (where multiple included files is an implementation detail). I obviously don't consider that a problem, and think the 1-to-1 mapping of source file - class name is a Java disease.

> Also, given that the time reduction was only on the order of a few percentage points

This patch only addresses one of multiple places we can improve. If we're doing this I expect to see some 50% reduction.

> (also tends to make make -j 8 less efficient as it serializes a little).

True, except a lot of these small files seem to be dominated by includes and setup, not useful work. They'd also disappear under -j8.

andlaus:
> You can think of it as one file = one module (where multiple included files is an implementation detail). I obviously don't consider that a problem, and think the 1-to-1 mapping of source file - class name is a Java disease.

I agree with @atgeirr: I'm easily getting confused (read: unproductive) if files get too long and do not only do one thing.

also in agreement with @atgeirr, I think that the savings in compile time are not worth it: if you want to accelerate the build, calling cmake with `-DBUILD_TESTING=OFF` helps *a lot* more (3m16 instead of 13m33 user time on my machine).

jokva:
> I agree with @atgeirr: I'm easily getting confused (read: unproductive) if files get too long and do not only do one thing.

I'm not proposing files doing more than one thing. The proposal is rethinking slightly what "one thing" is. That's why I'm using Deck as the example - it's one tree and can be considered as such. The five types that make up that "one thing" doesn't take that away.

> also in agreement with @atgeirr, I think that the savings in compile time are not worth it: if you want to accelerate the build, calling cmake with -DBUILD_TESTING=OFF helps a lot more (3m16 instead of 13m33 user time on my machine).

It's unacceptable for parser developers to turn off testing, but that's certainly what you should do downstream.

andlaus:
> It's unacceptable for parser developers to turn off testing, but that's certainly what you should do downstream.

yeah, I know. but if you're interested in developing the parser itself, you can change this to `-DBUILD_TESTING=ON` and you will probably not mind the extra minute of waiting time too much. Nothe that I always disable the unit tests for all modules which I'm not interested in, even those of opm-material and ewoms (if I'm working on something else). this allows me to do a rebuild of opm-simulators from scratch in about 5 minutes of real time. (I pretty sure that this is unachievable with enabled tests regardless of how much the build times are optimized.)

jokva:
> yeah, I know. but if you're interested in developing the parser itself, you can change this to -DBUILD_TESTING=ON and you will probably not mind the extra minute of waiting time too much.

You're wrong about that.

andlaus:
> You're wrong about that.

I don't mind to wait for a few extra seconds if opm-parser needs some changes. (actually, what's the *real time* improvement of you patch? I do not care about user time at all because I've got a multi-core machine.) what I *do* mind is when I don't find the location of the implementation of the classes anymore because they have been mingled into a single source file and I'll thus have to `grep` my way...

jokva:
`Deck` related types -> Deck.hpp

bska:
## Without patch ##
```
  opt/seq/dune-2.4.1/opm-parser
  --> cmake -DGCC_WARN_SHADOW=ON /home/bska/work/opm/src/opm/opm-parser
      Time = 5.68 [s]
  --> cmake --build . --target all -- -k -j20
      Time = 93.9 [s]: Success
  --> cmake --build . --target install -- -j1
      Time = 9.96 [s]: Success
```

## With patch (i.e., including #1007), rebased onto `master` ##
```
  opt/seq/dune-2.4.1/opm-parser
  --> cmake -DGCC_WARN_SHADOW=ON /home/bska/work/opm/src/opm/opm-parser
      Time = 5.9 [s]
  --> cmake --build . --target all -- -k -j20
      Time = 94.9 [s]: Success
  --> cmake --build . --target install -- -j1
      Time = 9 [s]: Success
```

In other words, no difference in wall-time between the two approaches (on my machine, at this instance at `-j20`).

andlaus:
> Deck related types -> Deck.hpp

I lost quite a few working hours on the fact that `GeoProps.[ch]pp` in opm-simulators does not contain a `GeoProps` class but all kinds of other stuff. you suggestion would have the same effect on opm-parser.

to answer my own question: on my machine, the build time in real time without this PR is 1.5 seconds longer ( 0m28.827s vs. 0m27.367s) without the tests and with tests it is 2 seconds faster with this PR (1m20.367s vs 1m22.685). do you really think it is worth to make such a fuzz about 2 seconds?

jokva:
> I lost quite a few working hours on the fact that GeoProps.[ch]pp in opm-simulators does not contain a GeoProps class but all kinds of other stuff. you suggestion would have the same effect on opm-parser.

This is a straw man, I'm not suggesting that. I presume this means you're confused since `memory` does not include a class `memory`, but rather `shared_ptr` and `unique_ptr`.

>  do you really think it is worth to make such a fuzz about 2 seconds?

Had this been the only change, then no, but this is just a topic for discussion. If we're going through with it we will do it for the rest of opm-parser as well, having a more central concept -> header model and drastically reduce the number of (trivial) source files.

You're seeing less gains with a faster machine, much because there's a *huge* barrier for the last few processes, since the tests won't build until the main library is built. Which is why looking at wall time is deceiving. Also, improvements in this area means faster CI which I think is a good thing. Travis already times out, and jenkins too takes forever.

I'd  like to add that this *is* a problem for me, I'm spending way too much time waiting for opm-parser. I don't have a 20 core machine, I'm dealing with slow disks and I frequently touch upon plenty of other places in parser. This is also an opportunity to reduce the number of public headers and what they contain which is *long* overdue. We'd still obviously like to keep the changes minimal and as little disruptive as possible, and that means getting experience and estimate the benefit before doing anything major. This is just that, but I do expect the payoff to be much more than two seconds.

andlaus:
> I'd like to add that this is a problem for me, I'm spending way too much time waiting for opm-parser. I don't have a 20 core machine

still, you have yet to answer how much time you gain in wall time with your 8 cores...

jokva:
I'll measure tomorrow. It's not the only interesting metric.

jokva:
make -j8:
* master: 2m45
* pr: 2m29

PR #1008, opened 2016-12-12
Remove pessimizing move() in return statements.

atgeirr:
This removes occurences of the anti-pattern of moving the return value. This prevents compile warnings on new compilers (at least clang), is simpler to read and understand, and can be faster.


atgeirr:
jenkins build this please

jokva:
This *really* shouldn't change anything, and I don't think we need to wait for jenkins for it.

Issue #999, opened 2016-12-03
NOSIM Issues


OPMUSER:
Trying to get OPM to work using NOSIM so I can see if this particular model can run under OPM, the model runs under Eclipse 100 and tNavigator, but NOSIM option keeps on failing. By failing I mean the software finds an unrecognizable keyword and then terminates. It should instead print a warning message and continue. The list of offending keywords with comments is attached.

Note the WELOPEN and COMPLUMP failure has already been addressed.

[OPM Testing Comments Rev-0-1.pdf](https://github.com/OPM/opm-parser/files/628704/OPM.Testing.Comments.Rev-0-1.pdf)

OPMUSER:
Updated attachment with Eclipse keywords that cause the problems for easier verification.
[OPM Testing Comments Rev-2.pdf](https://github.com/OPM/opm-parser/files/628710/OPM.Testing.Comments.Rev-2.pdf)

PR #1010, opened 2016-12-14
Parse new set of keywords

jokva:
This is largely addressing https://github.com/OPM/opm-parser/issues/999

This PR does not add support for any new keywords¹ except that we no longer crash on the keywords reported as problematic in #999. The keywords aren't exposed via EclipseState nor will they affect flow's behaviour, but at least we won't crash and burn on them. This is roughly equivalent to using PARSE_RANDOM_KEYWORD option in a Deck that uses these keywords, but more robust as opm-parser now has a concept of their structure.

¹ We already support several of the keywords GMWSET and FMWSET expand to, and this subset is obviously supported.


joakim-hove:
I think this should just be merged right away!?

jokva:
Happy to.

Issue #795, opened 2016-05-08
Difficult api: EclipseDeck?


joakim-hove:
The seemingly quite simple PR: https://github.com/OPM/opm-upscaling/pull/188 is an example of a recurring problem:
1. The `Deck` datatype is essentially a quite simple container of `DeckKeyword` instances, the `Deck` container gives no upfront guarantees of the content of the container.
2. The `EclipseState` constructor requires a valid[1] deck as input, and will raise an exception if e.g. the `GRID` section is missing.
3. The `EclipseGrid( Deck )`constructor is by construction accepts a non valid grid - in particular a grid without a `GRID` section.

We therefor have a situation where the two constructors:

``` C++
EclipseGrid::EclipseGrid( Deck ) 
...
EclipseState::EclipseState( Deck, )
```

have different semantic requirement to the input argument - that is quite difficult to see!

**EclipseDeck**

As a possible improvement I suggest that we create a class e.g. `EclipseDeck` which wraps/extends an existing `Deck` instance and enforces deck validity in the constructor:

``` C++
EclipseDeck::EclipseDeck( const Deck& deck) {
    if (!deck.hasKeyword("GRID"))
      throw std::invalid_argument("Missing GRID section");
   ...
}
```

By using the `EclipseDeck` or `Deck` type in the api we can clearly communicate whether we want a semantically correct Eclipse deck, or just a random collection of keywords.

[1]: Valid is quite poorly defined in this context, but it at least requires some sections. Going along the suggested route with `EclipseDeck` we could also give a much better definition to _valid deck_.

andlaus:
couldn't these "correctness requirements" be simply folded into the parseMode with EclipseState doing the parser creation and deck parsing internally?

also, IMO it would be beneficial if `EclipseState` would provide a `getDeck()` method: it seems to be a common pattern in C++ that higher-level classes provide a "I-know-what-I'm-doing,-give-me-the-low-level-stuff-in-all-its-gory" method; e.g. std::sting::c_str() and std::sting::data(). why should `EclipseState` stay different in that respect?

joakim-hove:
> couldn't these "correctness requirements" be simply folded into the parseMode with EclipseState doing the parser creation and deck parsing internally?

[...]

PR #1011, opened 2016-12-14
Split parser interface into public/private components

jokva:
This is the first PR in a series to encapsulate the internals of opm-parser better, and have a clear split on public/private interfaces. The endgame is to be able to move faster internally, provide a more stable programming interface for downstream and eventually mature and version the public interface.

The PR itself is huge, but is largely renames. A brief overview:

* A new directory, `eclipse/bits` is introduced. Internal headers will go here, and if one of these headers are included we *do not guarantee* anything w.r.t. stability, availability or breakage. Use at your own risk.
* EclipseState, Parser, and Units, are moved to `eclipse`. Files directly under `eclipse` are meant to be included by users, and when we mature it should be sufficient to include these headers in your project to fully leverage opm-parser. This also means that if you include these headers (or a subset of them) we will maintain includes for new features.
* The `Deck` class is moved into `eclipse/bits` and is considered internal. There will be a grace period, but querying `Deck` instances directly will from now on be considered unsupported. Once downstream is properly ported we might even stop *installing* the Deck headers on user systems, making them a build-time detail.
* `Deck` is no longer a part of `Parser`'s interface.

The final point is the only behavioural change in this PR. Including `Parser.hpp` will no longer provide the `Deck` declaration. This also means the static methods `Parser::parse, Parser::parseFile` etc. are gone. They have been replaced (essentially renamed) by the function `ecl::parse`, declared in EclipseState.hpp. The final interface is still up for discussion, and I'd love feedback on the solution.

The moving of files was done doing roughly this:
```
1. mv {Deck, RawDeck, Utility, Generator} -> bits/
2. find -name "*.[ch]pp" -exec sed -i 's,/Deck/,/bits/Deck/,' {} \;
3. mv {EclipseState/EclipseState.hpp} -> EclipseState.hpp
4. mv {Parser/Parser.hpp} -> Parser.hpp
5. find -name "*.[ch]pp" -exec sed -i 's,EclipseState/EclipseState.hpp,EclipseState.hpp,' {} \;

```

edit: 
PS. Notice that `bits/Parsers.hpp` isn't installed on user systems. This is by design.
PPS. I intend to later extend `bits/Parsers.hpp` with other input -> public object functions to hide the deck implementation detail.
PPS. First PS is wrong, `bits/Parsers.hpp` is installed.


andlaus:
First of all, I think this PR changes things for the better! while at it, what are the plans to reorganize the module in order to *remove* some directory levels? (IMO places like `opm/parser/eclipse/EclipseState/SummaryConfig/tests` are not exactly intuitive.)

jokva:
Still under discussion, but there will most likely be EclipseState/Schedule/, EclipseState/Grid and the rest just under EclipseState. No guarantees though.

andlaus:
can `opm/parser/eclipse` shortened to `opm/parser`? (the entropy `eclipse` is zero if everything is in that directory.)

jokva:
Probably not, we're still toying with the idea of multiple frontends.

jokva:
> PS. Notice that bits/Parsers.hpp isn't installed on user systems. This is by design.

I have to take this back, I have to install it for now because opm-material uses Deck so much.

pgdr:
This PR perfectly addresses issues raised in #795 — _Difficult api: EclipseDeck?_

atgeirr:
> This PR perfectly addresses issues raised in #795 — Difficult api: EclipseDeck?

A bit along the way, yes. I do think that we should *not* make Deck private. That (by requiring EclipseState) will eventually break lots of code that currently works well on imperfect/invalid/sectionless decks.

I don't understand why we need the Parser class to be public though, other than as a gateway to Deck and EclipseState?

jokva:
> A bit along the way, yes. I do think that we should not make Deck private. That (by requiring EclipseState) will eventually break lots of code that currently works well on imperfect/invalid/sectionless decks.

Deck should still be private, but a part of that process is being able to instantiate sub objects of eclipsestate (which is largely an aggregate of these) with relaxed requirements to deck completeness.

> I don't understand why we need the Parser class to be public though, other than as a gateway to Deck and EclipseState?

User-modifyable behaviour of the parsing process. This is mostly uninteresting from flow, but there are other applications where this can be useful. For flow and similar applications doing `ecl::parse( inputfile )` should be sufficient.

atgeirr:
> Deck should still be private, but a part of that process is being able to instantiate sub objects of eclipsestate (which is largely an aggregate of these) with relaxed requirements to deck completeness.

If we call the Deck interface "level 1" and the EclipseGrid etc. interface "level 2" (and the EclipseState being a validated set of level 2 things): I take this to mean that you intend to provide level 2 interfaces for absolutely every single possible keyword/feature? I think that sounds like a *lot* of work, especially as we move to the clients that are not Flow. Are you sure this is worth giving such a high priority?

If it is, then by all means: I like how for example the EclipseState takes away the need to understand many gory details, and would appreciate all the level 2 interfaces.

jokva:
If EclipseState or similar facilities aren't providing it then it will be re-implemented by the downstream that need it, often poorly and without proper context. I'd rather see that be a contribution to parser. So in a sense, yes.

pgdr:
@jokva this must probably be rebased after the jfunc; but then I think we want to merge this as soon as possible.  @joakim-hove do you have any comments before a potential merge?

jokva:
@pgdr Rebasing isn't a problem. I have a couple of other PRs that should go first though, https://github.com/OPM/opm-parser/pull/1016 https://github.com/OPM/opm-parser/pull/1015 and respective downstreams (and the additions they will spawn).

joakim-hove:

PR #1013, opened 2016-12-15
Provide DENSITY & ROCK table through EclipseState

jokva:
~~Title really does say it all.~~

This has changed a bit since it was first uploaded. Short synopsis of new scope and feature set:

1. Introducing `FlatTable` for structurally similar list-of-records such as DENSITY and PVTW. This is an internal parser detail and really isn't exposed to user  code. It hooks into `std::vector` for its implementation and carries vector semantics.
2. `DENSITY` and `PVTW` implemented with `FlatTable`
3. `ROCK` implemented with FlatTable as well.
4. RS+RV grid properties exposed, similar to PRESSURE and SWAT, SGAS.


joakim-hove:
jenkins build this with downstreams opm-core=1120 please

opm/parser/eclipse/EclipseState/Tables/Tables.cpp
```diff
@@ -67,6 +67,7 @@
 #include <opm/parser/eclipse/EclipseState/Tables/SwofTable.hpp>
 #include <opm/parser/eclipse/EclipseState/Tables/TableContainer.hpp>
 #include <opm/parser/eclipse/EclipseState/Tables/WatvisctTable.hpp>
+#include <opm/parser/eclipse/EclipseState/Tables/DensityTable.hpp>
```
joakim-hove: The file `DensityTable.hpp` is not installed.

jokva:
Good catch, I'll patch right away.

opm/parser/eclipse/EclipseState/Tables/Tables.cpp
andlaus: is DENSITY a table keyword? As far as I can remember, there's only one entry per record and nothing is interpolated?

joakim-hove: > is DENSITY a table keyword?

Good question; I guess the `DENSITY` keyword goes like a thermodynamic property - which is what you will find in the tables directory. It is a border-line case.

jokva: > is DENSITY a table keyword? As far as I can remember, there's only one entry per record and nothing is interpolated?

One (per phase) entry per record, which corresponds to regions. No interpolation as far as I'm aware.

It's as @joakim-hove points out an edge case - it's very similar in structure to a lot of the other tables, and there might be more properties soon that are more similar to DENSITY than anything else. We'll see, maybe it'll soon be time for `properties` in addition to `tables`, but there's plenty of time for that later.

I'm happy to move it though if we find a better home.

joakim-hove:
jenkins build this with downstreams opm-core=1120 please

opm/parser/eclipse/EclipseState/Tables/Tables.cpp
andlaus: well, the confusion is probably in my side: when I originally wrote these classes I intended them to represent continuous (1D or 2D) functions. since DENSITY represents just single data points (?), this keyword does not fall into this definition. (my original mental model may have become obsolete since then, though.) Anyway, I don't think this is a big deal.

jokva:
Maybe if I include the header in the commit too.

jokva:
jenkins build this with downstreams opm-core=1120 please

jokva:
jenkins build this with downstreams opm-core=1120 opm-simulators=984 please

OPMUSER:
Density keyword format is:

--  OIL      WAT GAS
--  DENSITY  DENSITY DENSITY
--  -------  ------- -------
DENSITY
       39.0    62.37     0.04519                   / DENSITY  PVT DATA 
REGION 1

With one line per PVT region, so it is not really a table but could be 
treated as.
ThunderBird Signature File

jokva:
jenkins build this with downstreams opm-core=1120 opm-simulators=984 please

jokva:
jenkins build this with downstreams opm-core=1120 opm-simulators=984 please

PR #1016, opened 2016-12-22
Understand PVCDO, VISCREF, WATDENT

jokva:
Used in opm-material.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Tables/FlatTable.hpp
+++ b/opm/parser/eclipse/EclipseState/Tables/FlatTable.hpp
@@ -48,6 +48,43 @@ struct RockTable : public FlatTable< ROCKRecord > {
     using FlatTable< ROCKRecord >::FlatTable;
 };
 
+struct PVCDORecord {
+    static constexpr std::size_t size = 5;
+
+    double reference_pressure;
+    double volume_factor;
+    double compressibility;
+    double viscosity;
+    double viscosibility;
+};
+
+struct PvcdoTable : public FlatTable< PVCDORecord > {
+    using FlatTable< PVCDORecord >::FlatTable;
+};
+
+struct VISCREFRecord {
+    static constexpr std::size_t size = 2;
+
+    double reference_pressure;
+    double reference_rs;
+};
+
+struct ViscrefTable : public FlatTable< VISCREFRecord > {
+    using FlatTable< VISCREFRecord >::FlatTable;
+};
+
+struct WATDENTRecord {
+    static constexpr std::size_t size = 3;
+
+    double reference_temperature;
+    double first_coefficient;
+    double second_coefficient;
+};
+
+struct WatdentTable : public FlatTable< WATDENTRecord > {
+    using FlatTable< WATDENTRecord >::FlatTable;
+};
+
 }
 
 #endif //OPM_FLAT_TABLE_HPP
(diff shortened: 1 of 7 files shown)


opm/parser/eclipse/EclipseState/Tables/FlatTable.hpp
```diff
@@ -48,6 +48,43 @@ struct RockTable : public FlatTable< ROCKRecord > {
...
+    static constexpr std::size_t size = 5;
+
+    double reference_pressure;
+    double volume_factor;
+    double compressibility;
+    double viscosity;
+    double viscosibility;
+};
+
+struct PvcdoTable : public FlatTable< PVCDORecord > {
```
andlaus: I suppose here the philosophical question what is considered to be a "table" kicks in again.

opm/parser/eclipse/EclipseState/Tables/Tables.cpp
```diff
@@ -1085,6 +1085,22 @@ struct flat_props< PVTWRecord, 0 > {
     }
 };
 
+constexpr const char* pvcdo_err[] = {
```
andlaus: that way to do error messages is pretty intransparent IMO.

jokva: Ok.

Please note that these keywords having error messages (defaulted value where it is illegal) is the *exception*. I valued making the common case simpler more. PVCDO is also special because *none* of the fields can be defaulted, in which case it was easily worked around. Typically only the first item cannot be defaulted, if any at all, in which case you only implement that.

andlaus: I just mean that something like

```c++
throw std::run_error("PVCDO reference pressure cannot be defaulted");
```

is much easier to comprehend than

```c++
throw std::run_error(pvcdo_err[0]);
```

and that a few duplicated strings for error messages are not a large maintenance burden. That said, if you think the second alternative is better, that's okay with me as well.

pgdr: I agree somewhat with Andreas here, but I want the most readable solution, and I don't immediately see the consequences of Andreas' solution.

Chose the one that you think will be the most readable (read: maintainable).

opm/parser/eclipse/EclipseState/Tables/tests/TableManagerTests.cpp
```diff
@@ -1101,6 +1101,26 @@ BOOST_AUTO_TEST_CASE( TestParseROCK ) {
...
+    auto deck = parser.parseString(data, Opm::ParseContext());
+    Opm::TableManager tables( deck );
+    const auto& pvcdo = tables.getPvcdoTable();
+
+    BOOST_CHECK_CLOSE( 3600.00, pvcdo[ 0 ].reference_pressure / 1e5, 1e-5 );
+    BOOST_CHECK_CLOSE( 1.12,    pvcdo[ 0 ].volume_factor, 1e-5 );
+    BOOST_CHECK_CLOSE( 1.6e-5,  pvcdo[ 0 ].compressibility * 1e5, 1e-5 );
+    BOOST_CHECK_CLOSE( 0.88,    pvcdo[ 0 ].viscosity * 1e3, 1e-5 );
+    BOOST_CHECK_CLOSE( 0.0,     pvcdo[ 0 ].viscosibility * 1e5, 1e-5 );
+}
```
pgdr: I would like to see a test that throws one of the aforementioned errors.

PR #1019, opened 2017-01-03
REFs hack for opm-material

jokva:
This is to facilitate the privatisation of the Deck object and
redefinition of the opm-parser interfaces. The opm-material module reads
the deck directly to extract temperature information, but this is
largely unsupported in opm-parser.

In order to not break downstream functionality we take ownership of
the opm-material implemented parsing of THERMX1 and friends and expose
their raw values via the REFs and get_refs_material function. This is
ONLY meant for compatibility and SHOULD NOT BE CALLED in new code.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Tables/TableManager.cpp
+++ b/opm/parser/eclipse/EclipseState/Tables/TableManager.cpp
@@ -105,6 +105,54 @@ namespace Opm {
         if( deck.hasKeyword( "WATDENT" ) )
             this->m_watdentTable = WatdentTable( deck.getKeyword( "WATDENT" ) );
 
+        if( deck.hasKeyword( "THERMEX1" ) ) {
+            const auto& thermex1 = deck.getKeyword( "THERMEX1" )
+                                       .getRecord( 0 )
+                                       .getItem( 0 );
+
+                this->refs_material.THERMEX1 = thermex1.getSIDoubleData();
+        }
+
+        if( deck.hasKeyword( "TREF" ) ) {
+            const auto& thermex1 = deck.getKeyword( "TREF" )
+                                       .getRecord( 0 )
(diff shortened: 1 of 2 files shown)


andlaus:
doesn't this require downstream changes?

> This is ONLY meant for compatibility and SHOULD NOT BE CALLED in new code.

In principle I don't mind the goal of making `Deck` an implementation detail of `EclipseState` (as long as I don't need to do the work associated with this), but do you intend to expose _all_ keywords from `opm/parser/share/keywords/` via `EclipseState`? If yes: First, I'm not sure if this is really worth the giant amount of work which this requires, and second, how can custom keywords such as the one used by `opm-upscaling` still be supported?

jokva:
> doesn't this require downstream changes?

Not to work, but the downstream changes are the goal.

> In principle I don't mind the goal of making Deck an implementation detail of EclipseState (as long as I don't need to do the work associated with this), but do you intend to expose all keywords from opm/parser/share/keywords/ via EclipseState?

Yes. All the work done downstream to read info from `Deck` objects should've been in opm-parser to begin with.

> First, I'm not sure if this is really worth the giant amount of work which this requires, and second, how can custom keywords such as the one used by opm-upscaling still be supported?

Firstly, it's not that big of a job really (rather, most of it's been done, except it's in the wrong module). As for the second point. That depends. If they're completely new keywords we'll support them as if they were native. If they're completely new semantics overriding eclipse keywords we'll have to take a closer look.

jokva:
Regarding `REFs` - I suppose the goal over time is to properly support temperature and all that jazz from opm-parser, but we're far away from that now, and we still want to close the interface. Hopefully having the interface more restricted means we can in a smoother fashion add support for temperature too.

joakim-hove:

PR #1020, opened 2017-01-03
Endpoint scaling support

jokva:
Understand and expose the ENDSCALE keyword.

### DIFF
--- a/opm/parser/eclipse/EclipseState/EndpointScaling.cpp
+++ b/opm/parser/eclipse/EclipseState/EndpointScaling.cpp
@@ -1,0 +1,113 @@
+#include <opm/parser/eclipse/Deck/Deck.hpp>
+#include <opm/parser/eclipse/EclipseState/Runspec.hpp>
+#include <opm/parser/eclipse/Utility/String.hpp>
+
+namespace Opm {
+
+EndpointScaling::operator bool() const noexcept {
+    return this->options[ static_cast< ue >( option::any ) ];
+}
+
+bool EndpointScaling::directional() const noexcept {
+    return this->options[ static_cast< ue >( option::directional ) ];
+}
+
+bool EndpointScaling::nondirectional() const noexcept {
+    return bool( *this ) && !this->directional();
+}
+
+bool EndpointScaling::reversible() const noexcept {
+    return this->options[ static_cast< ue >( option::reversible ) ];
+}
+
+bool EndpointScaling::irreversible() const noexcept {
+    return bool( *this ) && !this->reversible();
+}
+
+bool EndpointScaling::twopoint() const noexcept {
+    return bool( *this ) && !this->threepoint();
+}
+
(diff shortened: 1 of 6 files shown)


opm/parser/eclipse/EclipseState/Runspec.cpp
```diff
@@ -62,23 +62,128 @@ size_t Phases::size() const noexcept {
...
+
+bool endscale_nodir( const DeckKeyword& kw ) {
+    if( kw.getRecord( 0 ).getItem( 0 ).defaultApplied( 0 ) )
+        return true;
+
+    const auto& value = kw.getRecord( 0 ).getItem( 0 ).get< std::string >( 0 );
+
+    if( value != "DIRECT" && value != "NODIR" )
+        throw std::invalid_argument(
+            "ENDSCALE argument 1 must be defaulted', 'DIR' or 'NODIR', was "
```
pgdr: `DIR` or `DIRECT`

opm/parser/eclipse/EclipseState/Runspec.cpp
```diff
@@ -62,23 +62,128 @@ size_t Phases::size() const noexcept {
...
+        throw std::invalid_argument( "SCALECRS takes 'YES' or 'NO'" );
+
+    return value == 'Y';
+}
+
+bool endscale_nodir( const DeckKeyword& kw ) {
+    if( kw.getRecord( 0 ).getItem( 0 ).defaultApplied( 0 ) )
+        return true;
+
+    const auto& value = kw.getRecord( 0 ).getItem( 0 ).get< std::string >( 0 );
```
joakim-hove: I guess `value`should be uppercased before the comparison.

joakim-hove:
I would like to have the `EndScale`implementation in separate files - apart from that I think this can be merged.

opm/parser/eclipse/EclipseState/Runspec.cpp
jokva: Absolutely right.

opm/parser/eclipse/EclipseState/EndpointScaling.cpp
```diff
@@ -0,0 +1,113 @@
...
+    return bool( *this ) && !this->threepoint();
+}
+
+bool EndpointScaling::threepoint() const noexcept {
+    return this->options[ static_cast< ue >( option::threepoint ) ];
+}
+
+namespace {
+
+bool threepoint_scaling( const Deck& deck ) {
```
andlaus: @joakim-hove: what's your plan for the naming convention? the lowercase_underscores() convention looks quite strange when mixed with camelCase() for "related" objects, IMO...

joakim-hove: Thank you for asking - I think this slightly tricky: See #1021 for my view of the situation.

PR #1021, opened 2017-01-03
Endscale rename

joakim-hove:
In liue of the [comment]( https://github.com/OPM/opm-parser/pull/1020#pullrequestreview-14974521) I have concluded that I merged #1020 a bit prematurely, as @andlaus commented the methodname `Runspec::endpoint_scaling()` is quite different from the main pattern in opm-parser.

Personally I find this a quite difficult subject to weigh down on.

1. A consistent codebase is obviously easier for everyone to navigate, and the information implicitly conveyed in a consistent codebase will catch bugs before they happen. 

2. Being a developer is a form of creative work - we are nearly artists; and I guess the freedom to make such slightly arbitrary choices like `Runspec::endpoint_scaling()` versus `Runspec::endpointScaling()` is valued by many of us. As developers we also change and mature; and it would be counterproductive to stick to bad decisions in the name of consistency. As @atgeirr once commented I feel there is something *fascistoid* to be said about insisting on a very uniform and consistent codebase; a *slightly* inconsistent codebase feels more alive.

But I do feel that the public name `Runspec::endpoint_scaling()` is *quite different* from the main pattern used in opm-parser, and combined with comments from @jokva which gave me the impression that he might be planning a wholesale rename, made me call for a timeout with this PR. 

Since opm-parser is (nearly) the most upstream of all modules the conventions applied here (at least in the public api) is relevant for all developers in the opm community, I would therefor be grateful if other developers mainly working downstream voiced their opinion and general thought on the matter. Observe that I do *not* hold any absolute views on whether `Runspec::endpoint_scaling()` or `Runspec::endpointScaling()` is "better". The question is mainly as @andlaus pointed out - is this a move *towards* the rest of the opm codebase, or *away* from it - and does it matter?

I have previously been slightly hostile - or at least quite uninterested - in things like common opm style guides; I am warming to the idea now. But if we are to go in that direction I think we must:

1. Have a broad discussion involving many developers; quite few of us feel any sort of ownership to the style guide found on the opm-wiki.
2. More concious separation of internal details (e.g. `m_xxx` versus `xxxx_` for private members) and the public api.
3. **Implement a solution with e.g. `cpplint` to [enforce a C++ style](http://www.ostricher.com/2014/11/use-cpplint-to-check-your-c-code-against-googles-style-guide/) as part of build testing.**

### DIFF
--- a/opm/parser/eclipse/CMakeLists.txt
+++ b/opm/parser/eclipse/CMakeLists.txt
@@ -175,6 +175,7 @@ EclipseState/EclipseState.hpp
 EclipseState/EclipseConfig.hpp
 EclipseState/Eclipse3DProperties.hpp
 EclipseState/Runspec.hpp
+EclipseState/EndpointScaling.hpp
 #
 EclipseState/checkDeck.hpp
 #
--- a/opm/parser/eclipse/EclipseState/EndpointScaling.cpp
+++ b/opm/parser/eclipse/EclipseState/EndpointScaling.cpp
@@ -1,5 +1,5 @@
 #include <opm/parser/eclipse/Deck/Deck.hpp>
-#include <opm/parser/eclipse/EclipseState/Runspec.hpp>
+#include <opm/parser/eclipse/EclipseState/EndpointScaling.hpp>
 #include <opm/parser/eclipse/Utility/String.hpp>
 
 namespace Opm {
--- a/opm/parser/eclipse/EclipseState/EndpointScaling.hpp
+++ b/opm/parser/eclipse/EclipseState/EndpointScaling.hpp
@@ -1,0 +1,56 @@
+/*
+  Copyright 2017  Statoil ASA.
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify it under the terms
+  of the GNU General Public License as published by the Free Software
+  Foundation, either version 3 of the License, or (at your option) any later
+  version.
+
+  OPM is distributed in the hope that it will be useful, but WITHOUT ANY
+  WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+  A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License along with
+  OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef OPM_ENDPOINTSCALING_HPP
+#define OPM_ENDPOINTSCALING_HPP
+
+#include <bitset>
+
+namespace Opm {
+class Deck;
+
+
+class EndpointScaling {
+    public:
+        EndpointScaling() noexcept = default;
+        explicit EndpointScaling( const Deck& );
+
+        /* true if endpoint scaling is enabled, otherwise false */
+        operator bool() const noexcept;
+
+        bool directional() const noexcept;
+        bool nondirectional() const noexcept;
+        bool reversible() const noexcept;
+        bool irreversible() const noexcept;
+        bool twopoint() const noexcept;
+        bool threepoint() const noexcept;
+
+    private:
+        enum class option {
+            any         = 0,
+            directional = 1,
+            reversible  = 2,
+            threepoint  = 3,
+        };
+
+        using ue = std::underlying_type< option >::type;
+        std::bitset< 4 > options;
+};
+}
+
+#endif
--- a/opm/parser/eclipse/EclipseState/Runspec.cpp
+++ b/opm/parser/eclipse/EclipseState/Runspec.cpp
@@ -80,7 +80,7 @@ const Tabdims& Runspec::tabdims() const noexcept {
     return this->m_tabdims;
 }
 
-const EndpointScaling& Runspec::endpoint_scaling() const noexcept {
+const EndpointScaling& Runspec::endpointScaling() const noexcept {
     return this->endscale;
 }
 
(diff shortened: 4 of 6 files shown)


joakim-hove:
OK - the main point about this PR was to initiate a style discussion; if we decide against merging the proposed rename `Runspec::endpoint_scaling() -> Runspec::endpointScaling()` I would still like to merge the first commit.

jokva:
A generally good idea in my opinion is to align with the language's standard library.

andlaus:
> Observe that I do *not* hold any absolute views on whether `Runspec::endpoint_scaling()` or `Runspec::endpointScaling()` is "better". 

me too, actually. I just like to keep stuff in **public** interfaces consistent in the sense that the code "feels" like it belongs to the same project. 

> The question is mainly as @andlaus pointed out - is this a move *towards* the rest of the opm codebase, or *away* from it - and does it matter?

I'm also a bit indifferent here. The current OPM conventions seems to lean more towards camelCase. (as do other large C++ projects like Dune, Qt or KDE.) on the other hand, the STL uses_underscores as do most C projects. I think both conventions are fine, but I prefer not mixing them in the same project, so if there is a general consensus that it is worthwhile to switch to underscores, I'll budge. (but we should convert to it in bulk and in a coordinated manner. I also don't think this is worth the effort.)

tl;dr: I think it's less work to stick with camelCase.

> I have previously been slightly hostile - or at least quite uninterested - in things like common opm style guides; I am warming to the idea now.

I agree. these guidelines should probably only be specific with regard to public APIs and the rest should stay in the discretion of the maintainers. (e.g., I couldn't care less whether private attributes in a module are called `m_foo` or `foo_`.) 

@jokva 
> A generally good idea in my opinion is to align with the language's standard library, at least if you're to force a style.

IMO that's only true if you're starting a project from scratch and/or most of your major dependencies also use this convention. since OPM uses Dune quite a lot, camelCase was chosen for the C++ parts (I guess).

jokva:
This is all for external interfaces, right? m_foo or foo_ or foo shouldn't matter.

joakim-hove:
> This is all for external interfaces, right? m_foo or foo_ or foo shouldn't matter.

That I fully agree - and my point is that a style guide should separate clearly between the external interfaces and internal implementation details; it should be considerable less strict - if not completely ignorant - when it comes to `m_foo` versus `foo_`.

jokva:
> as do other large C++ projects like Dune, Qt or KDE

Two of these projects predate STL and a (reasonable) C++ standard library.

andlaus:
> Two of these projects predate STL and a (reasonable) C++ standard library.

which are the two? according to wikipedia the STL was started in the 1980s while Qt was first released in 1995 and KDE after it.

jokva:
No, Stepanov programmed generically in the 80s. STL was formalised in C++98, proposed in 93 or 94 or so.

> This work undoubtedly would have continued for some time being just a research project or at best would have resulted in an HP proprietary library, if Andrew Koenig of Bell Labs had not become aware of the work and asked Stepanov to present the main ideas at a November 1993 meeting of the ANSI/ISO committee for C++ standardization. The committee's response was overwhelmingly favorable and led to a request from Koenig for a formal proposal in time for the March 1994 meeting. Despite the tremendous time pressure, Alex and Meng were able to produce a draft proposal that received preliminary approval at that meeting.

I recommend we stop here, I just found the examples interesting since there really was no standard library at the time.

blattms:
I am not helping but I just can help myself saying: "Why not use a new
naming scheme camel_Case and rule all future projects!" ;)

I honestly do not care what kind of naming scheme a project/module
uses. opm-parser in particular as I rarely touch it.

What I usually appreciate is some kind of consistency (well sometimes
there is some historical baggage), not changing the scheme on the
run. If you do change it for existing stuff, please be aware that this
poses additional work on users for no obvious gain. There should be a
_really_ good reason.

atgeirr:
First, my opinions:
 - The coding standard on the wiki (http://wiki.opm-project.org/index.php?title=Suggested_coding_standard) still contain my suggestions for a good standard. The most important point is that we should use the *C++ coding standards* book as a starting point, however I suspect that many of you do not have this book? The style issues are less important, but not unimportant. I do think that the wiki page has organized these issues pretty well, so perhaps any continued style discussion could use that as starting point, referring to its guidelines/parts when challenging or supporting it?
 - We should not hold public and private code parts to different standards.
 - We must be willing to adapt out personal styles to the style already used in the class/file/module we are adding to. As an example: when I add a member to a class where members have trailing underscores, I add one too. When I add a member to a class where they have a `m_` prefix I add one too. When making a new class I should follow the conventions of the module I am in, especially if asked nicely :-)
 -  I think we should not compromise on indentation in C++ files but ensure consistency.

Second, here follows an attempt to make a summary of the discussion. While I cannot see a true consensus, there seems to be some clear majorities for the following points:
 - that we *should not* change styles for existing modules, consistency is valuable
 - we *should* keep the ClassName, functionName pattern for consistency, 
 - apart from above, private implementation details *should* be given more stylistic freedom (really folks?!? Well I guess I'll have to give in...)

(The above attempts to be a statement of fact, not opinion, so please correct me if I am wrong.)

Finally, on this PR: I agree that we should change the name to the commonly used pattern.

joakim-hove:
OK; thank you for the input. As maintainer in opm-parser I will now enforce the following naming conventions:

1. We will stick with `camelCase` in the public api. 
2. I will be positive to shorter symbols names i.e. omitting the `get` in `getXXX()`.
3. No style enforcing for internal symbols, although internal consistency with one compilation unit should be a goal.

PR #1022, opened 2017-01-09
Skip AITS and ROCKNUM

pgdr:
Commit 1.  Implements `operator==` and `operator!=` for `Well`.

Commit 2.  Parser now handles `AITS` (but ignores it).

Commit 2.  Internalize `ROCKNUM` and added test.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Well.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Well.cpp
@@ -88,6 +88,17 @@ namespace Opm {
     }
 
 
+    bool Well::operator==(const Well& other) const {
+        return this->m_creationTimeStep == other.m_creationTimeStep
+            && this->m_name == other.m_name
+            && this->m_preferredPhase == other.m_preferredPhase
+            && this->timesteps == other.timesteps;
+    }
+
+    bool Well::operator!=(const Well& other) const {
+        return !(*this == other);
+    }
+
     double Well::production_rate( Phase phase, size_t timestep ) const {
         if( !this->isProducer( timestep ) ) return 0.0;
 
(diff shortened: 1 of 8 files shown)


opm/parser/eclipse/EclipseState/Schedule/Well.cpp
```diff
@@ -88,6 +88,14 @@ namespace Opm {
...
 
 
+    bool Well::operator==(const Well& other) {
+        const auto & me = *this;
+        return me.m_creationTimeStep == other.m_creationTimeStep
+            && me.m_name == other.m_name
+            && me.m_preferredPhase == other.m_preferredPhase
+            && me.timesteps == other.timesteps;
+    }
+
```
joakim-hove: Maybe also add `Well::operator!=` - or is that automatic?

jokva: It's not¹

¹http://en.cppreference.com/w/cpp/utility/rel_ops/operator_cmp

pgdr: thanks for the feedback, it's been implemented, together with a test.

opm/parser/share/keywords/000_Eclipse100/R/ROCKNUM
```diff
@@ -0,0 +1 @@
+{"name" : "ROCKNUM" , "sections" : ["REGIONS"], "data" : {"value_type" : "INT"}}
```
joakim-hove: Would be very simple to add "full" support for `ROCKNUM` by adding an entry [here](https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp#L188)

andlaus: also please add an integer grid property (at least if you don't want downstream to resort to using the deck directly [using the deck here would be a worse kludge than usual because it ignore the modifier keywords, but it is a thing one does if one is desperate enough ;)])

andlaus: @joakim-hove: sorry, I did not notice that you suggested the same...

pgdr:
ready for review

PR #1023, opened 2017-01-09
WIP Scaled saturation function support

jokva:
The goal is to support (scaled) saturation functions from opm-parser, rather than downstream having to interpret tables.

This is highly experimental and still in its *very* early stages, but I'm publishing it now for discussion, thoughts, observations, wishes and warm fuzzy feelings. I encourage downstream to chime in and let us know what exactly you need from the saturation functions, how you intend to use them and represent them, in order for the interface to be nice and simple.

I intend for this effort to also give us some nice documentation on edge cases and subtle interactions, and more importantly, good tests.

If everything goes as I hope it will, downstream won't have to touch 3d properties or tables again.

### DIFF
--- a/opm/parser/eclipse/CMakeLists.txt
+++ b/opm/parser/eclipse/CMakeLists.txt
@@ -81,6 +81,7 @@ EclipseState/EclipseConfig.cpp
 EclipseState/Eclipse3DProperties.cpp
 EclipseState/Runspec.cpp
 EclipseState/EndpointScaling.cpp
+EclipseState/saturation.cpp
 #
 EclipseState/checkDeck.cpp
 #
--- a/opm/parser/eclipse/EclipseState/Grid/tests/CMakeLists.txt
+++ b/opm/parser/eclipse/EclipseState/Grid/tests/CMakeLists.txt
@@ -2,7 +2,8 @@ foreach(tapp EclipseGridTests MULTREGTScannerTests GridPropertyTests
              FaceDirTests GridPropertiesTests BoxTests PORVTests
              BoxManagerTests TransMultTests FaultTests
              EqualRegTests MultiRegTests ADDREGTests CopyRegTests
-             SatfuncPropertyInitializersTests)
+             SatfuncPropertyInitializersTests
+             SaturationTests)
   opm_add_test(run${tapp} SOURCES ${tapp}.cpp
                           LIBRARIES opmparser ${Boost_LIBRARIES})
 endforeach()
--- a/opm/parser/eclipse/EclipseState/Grid/tests/SaturationTests.cpp
+++ b/opm/parser/eclipse/EclipseState/Grid/tests/SaturationTests.cpp
@@ -1,0 +1,163 @@
+#define BOOST_TEST_MODULE EclipseGridTests
+
+#include <boost/test/unit_test.hpp>
+
+#include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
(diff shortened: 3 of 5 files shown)


atgeirr:
Major comment:
 - Is this necessary? Wanted? It seems to me it will be a big effort, and will replace work now in opm-material. Is not that module a better location for interpreting the details of the multiphase flow functions? What is the benefit we can get from this? What does the proposed interface look like?

Minor comment:
 - Please do not start a naming revolution, but keep the Class:methodName() pattern that up to now has been pretty much the only code style thing we have managed to keep more or less consistent across OPM.

joakim-hove:
> Please do not start a naming revolution, but keep the Class:methodName() pattern that up to now has been pretty much the only code style thing we have managed to keep more or less consistent across OPM.

This #1021 was an attempt to open a discussion on that topic; happy to hear your opinion there.

joakim-hove:
This code seems to completely bypass / reinvent the existing table code in opm-parser?

atgeirr:
> This #1021 was an attempt to open a discussion on that topic; happy to hear your opinion there.

Others said things I agreed with, and I weighed against a "me too" post, I got the impression that the consensus was staying with the existing convention? I'll try to say something there as well, then.

jokva:
> This code seems to completely bypass / reinvent the existing table code in opm-parser?

By design for the experiment, yes.

andlaus:
(disclaimer: I haven't looked at the actual code yet.) 

just my 2c: In principle, I don't really care where this code is as long as it works without functional- and performance regressions. Though, as far as I can see, it is only possible do do half of the job in the opm-parser module (at most) because I'm pretty sure that the parser is not the right place to deal with stuff like automatic differentiation and if you agree with this, the code which implements the actual satfuncs must be aware of most of the nitty-gritty details (2-point or 3-point enpoint scaling, hysteresis, JFUNC, threephase model, etc.) anyway. so, from a global perspective, what do we gain with this?

tl;dr: I don't see how this could improve the current solution substantially, but I'm happy to be proven wrong.

jokva:
It will probably be killed - it was based on the (wrong) assumption that scaling was largely limited to a cell-by-cell static modification of functions, based on whichever modifiers are present in the deck, i.e. construct N functions `fn :: saturation -> ( relperm, capillary pressure )`. I've just had a chat with @alfbr and that seems to not be the case - the functions aren't as static as I assumed they'd be) which means the underlying idea for this approach is flawed.

jokva:
> Is this necessary? Wanted? It seems to me it will be a big effort, and will replace work now in opm-material. Is not that module a better location for interpreting the details of the multiphase flow functions? What is the benefit we can get from this? What does the proposed interface look like?

In that process (interpreting whatever would be the output of this effort) material does a lot of parsing work.

alfbr:
Saturation functions are complex beasts. While the functionality present in opm-material today can be envisioned in the parser code, I am not sure it is a viable strategy for the future. Basically, scaling of saturation functions are even history dependent through hysterisis. Hence, the simulator will need to do scaling on the fly regardless of what has been done at parse time. Today, the parser handles reading in the tables, and making them available to the simulator. This is a very straight forward interface, that can be made available in the EclipseState. If we start to include scaling, it is very difficult (at least for me) to see where we draw the line, i.e., it will then be very difficult to come up with a good API between the parser and the simulator.

The problem @jokva points out is important though. The current handling of J-functions in opm-material clearly demonstrates that we need to do re-factoring to get a cleaner interface between the parser and the simulator.

joakim-hove:

PR #1024, opened 2017-01-09
Added JSON for FIPOWG

pgdr:
ps., are anyone familiar with `FIPREG`?

### DIFF
--- a/opm/parser/share/keywords/000_Eclipse100/O/OPERNUM
+++ b/opm/parser/share/keywords/000_Eclipse100/O/OPERNUM
@@ -1,0 +1,5 @@
+{
+ "name" : "OPERNUM",
+ "sections" : [ "GRID" ],
+ "data" : {"value_type" : "INT" }
+}
--- a/opm/parser/share/keywords/000_Eclipse100/F/FIPOWG
+++ b/opm/parser/share/keywords/000_Eclipse100/F/FIPOWG
@@ -1,0 +1,1 @@
+{"name" : "FIPOWG", "sections" : ["REGIONS"]}
--- a/opm/parser/share/keywords/000_Eclipse100/W/WELL_PROBE
+++ b/opm/parser/share/keywords/000_Eclipse100/W/WELL_PROBE
@@ -142,6 +142,7 @@
         "WSTAT",
         "WMCTL",
         "WMCON",
+        "WOVPR",
         "WEPR",
         "WEPT",
         "WEFF",


joakim-hove:
> ps., are anyone familiar with FIPREG?

See the documentation of `FIP` - there is an autocreation functionality for regions keywords; i.e. 
```txt
FIPREG
    -- nx * ny * nz integers /

FIPXYZ
   -- nx * ny * nz integers /
```
will autocreate to region properties `FIPREG` and `FIPXYZ`- which can/will then subsequently be used for reporting purposes; we do not support this auto creation property (I think ....).

opm/parser/share/keywords/000_Eclipse100/O/OPERNUM
```diff
@@ -0,0 +1,5 @@
+{
+ "name" : "OPERNUM",
```
andlaus: please add this to the integer grid properties. (it can be manipulated with modifier keywords, right?)

OPMUSER:
Yes.
ThunderBird Signature File

opm/parser/share/keywords/000_Eclipse100/O/OPERNUM
pgdr: @andlaus we don't support the `OPERNUM` keyword, but I added it to the `REGIONS` section in  #1031

andlaus: I'm confused: doesn't the presence of this JSON file mean that the OPERNUM *keyword* is supported. (though the corresponding *grid property* is still missing, as far as I can see.)

pgdr: I'm not quite sure what you mean.  If you read the Eclipse manual, it will become obvious that we don't _support_ `OPERNUM`.  The purpose of this PR was to not _die_ when we encountered this keyword in a very special model I have on my desk.

If someone comes with a model using `OPERNUM` properly, we must consider actually implementing the proper support, but that's a non-trivial job.

andlaus: > If you read the Eclipse manual, it will become obvious that we don't support OPERNUM.

the *simulator* does not support OPERNUM, but so far this has never prevented the *parser* to provide proper support (at least if it was not much effort)?

> The purpose of this PR was to not die when we encountered this keyword in a very special model I have on my desk.

yeah, but adding that integer grid property is a one-liner. (as of now the parser probably still dies if it sees something like

```
COPY
"FLUXNUM" "OPERNUM" /
/
```

?)

pgdr: Again, I'm not sure what you mean.  The parser doesn't support `OPERNUM` either (read the manual).

Do you mean this?
https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp#L197

andlaus: > The parser doesn't support OPERNUM either (read the manual).

as far as I interpret this, opm-parser just does does not support/enforce the quirks/documented E100 bugs?

> Do you mean this?

yep. that's what I meant. sorry for the noise.

PR #1025, opened 2017-01-09
added BUILD_SHARED_LIBS to option in cmake

pgdr:
This adds `BUILD_SHARED_LIBS` as a `CMake` option so it is possible to toggle in `ccmake`.


pgdr:
But, as far as I know, it is not exposed in ccmake in any way?

joakim-hove:
> But, as far as I know, it is not exposed in ccmake in any way?

No you are right - cmake *understands* it, but you must still expose it manually they way you have done in this PR.

PR #1026, opened 2017-01-10
Understand STONE1EX.

jokva:
Understand and expose the STONE1EX table. It has a known bug when values
are defaulted - it won't copy the previous table's entry, but rather
insert 1.0.

### DIFF
--- a/opm/parser/eclipse/CMakeLists.txt
+++ b/opm/parser/eclipse/CMakeLists.txt
@@ -242,6 +242,7 @@ EclipseState/Tables/PvdsTable.hpp
 EclipseState/Tables/SimpleTable.hpp
 EclipseState/Tables/PlymaxTable.hpp
 EclipseState/Tables/PlyrockTable.hpp
+EclipseState/Tables/Stone1exTable.hpp
 EclipseState/Tables/SwofTable.hpp
 EclipseState/Tables/SgwfnTable.hpp
 EclipseState/Tables/SwfnTable.hpp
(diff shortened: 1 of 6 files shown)


andlaus:
the patch seems fine. (besides the philosophical issue that STONE1EX is not a table keyword IMO: columns which always exhibit exactly one entry should be called entries, but I can live with that)

jokva:
It's also slightly buggy because it behaves differently than some other tables (again, what are they) w.r.t. defaulted behaviour. I suppose it's due for refactoring soon, which hopefully addresses some of these problems.

andlaus:
> (again, what are they)

IMO tables represent continuous functions, i.e., it should make sense to interpolate between values.

joakim-hove:
No - this is not a table; it should rather be implemented e.g. as the `EQUIL` keyword.

pgdr:
> IMO tables represent continuous functions, i.e., it should make sense to interpolate between values.

Piecewise linear, no?

alfbr:
How to interpolate is up to the simulator, with piecewise linear being common. As @joakim-hove says, this is not a table in that respect. It is a set of exponents, one for each saturation function region.

andlaus:
> Piecewise linear, no?

STONE1EX is one value per saturation region (for those, adjacent region numbers cannot really be considered to be related nor is there any other number which can be used for the x-Axis), so it doesn't even much sense to consider it a piecewise constant function?

andlaus:
> How to interpolate is up to the simulator, with piecewise linear being common.

it is only "kind of" up to the simulator: if you want to get a good match with E100, you better use piecewise linear interpolation ;). (in opm-material, switching to monotonic cubic splines is pretty much a no-brainer, though.)

alfbr:
> it is only "kind of" up to the simulator: if you want to get a good match with E100, you better use piecewise linear interpolation ;). (in opm-material, switching to monotonic cubic splines is pretty much a no-brainer, though.)

Yes, but I hope we can get past that. Smoother interpolation functions may improve numerics, and it is more physical.

PR #1027, opened 2017-01-10
Expose preferred relative permeability model

jokva:
The currently supported models are default, STONE1 and STONE2.

### DIFF
--- a/opm/parser/eclipse/EclipseState/SimulationConfig/HysteresisOptions.cpp
+++ b/opm/parser/eclipse/EclipseState/SimulationConfig/HysteresisOptions.cpp
@@ -1,0 +1,230 @@
+#include <opm/parser/eclipse/EclipseState/SimulationConfig/HysteresisOptions.hpp>
+#include <opm/parser/eclipse/Utility/String.hpp>
+
+namespace Opm {
+
+namespace {
+
+HysteresisOptions::Model readHysteresisModel( int x ) {
+    using Model = HysteresisOptions::Model;
+    switch( x ) {
+        case -1: return Model::Equilibration;
+        case  0: return Model::CarlsonHysteresisNonWettingDrainageWetting;
+        case  1: return Model::CarlsonHysteresisNonWettingImbibitionNonWetting;
+        case  2: return Model::KilloughHysteresisNonWettingDrainageWetting;
+        case  3: return Model::KilloughHysteresisNonWettingImbibitionWetting;
+        case  4: return Model::KilloughHysteresisNonWettingKilloughHysteresisWetting;
+        case  5: return Model::CarlsonNonWettingGasWaterDrainageOil;
+        case  6: return Model::KilloughNonWettingGasWaterDrainageOil;
+        case  7: return Model::KilloughHysteresisNonWettingGasWaterWettingOil;
+        case  8: return Model::JargonHysteresisNonWettingDrainageWetting;
+        case  9: return Model::JargonHysteresisNonWettingImbibitionWetting;
+        default:
+            throw std::invalid_argument( "Hysteresis model must be an integer [-1, 9]" );
+    }         
+}
+
+int readHysteresisApply( const DeckItem& item ) {
+    if( item.defaultApplied( 0 ) ) return 0;
+
+    const auto& key = uppercase( item.getTrimmedString( 0 ) );
+
+    if( key == "BOTH" ) return 0;
+    if( key == "PC" ) return 1;
+    if( key == "KR" ) return 2;
+
+    throw std::invalid_argument( "Limiting hysteresis flag must be 'PC', 'KR' or 'BOTH'" );
+}
+
+HysteresisOptions::CapillaryPressureScanningCurveShape readHysteresisCurveShape( const DeckItem& item ) {
+    if( item.defaultApplied( 0 ) )
+        return HysteresisOptions::CapillaryPressureScanningCurveShape::RETR;
+
+    const auto& key = uppercase( item.getTrimmedString( 0 ) );
+    if( key == "RETR" )
+        return HysteresisOptions::CapillaryPressureScanningCurveShape::RETR;
+
+    if( key == "NEW" )
+        return HysteresisOptions::CapillaryPressureScanningCurveShape::RETR;
+
+    throw std::invalid_argument( "Flag controlling the shape of capillary pressure curves on retraversal must be 'RETR' or 'NEW'" );
+}
+
+bool readHysteresisCorrectImbibition( const DeckItem& item ) {
+    if( item.defaultApplied( 0 ) ) return false;
+
+    const auto& key = uppercase( item.getTrimmedString( 0 ) );
+    if( key == "DRAIN" ) return false;
+    if( key == "BOTH" ) return true;
+
+    throw std::invalid_argument( "Flag controlling fluid mobility correction application on imbibition curves must be 'DRAIN' or 'BOTH'" );
+}
+
+Phase readHysteresisWettingPhase( const DeckItem& item ) {
+    if( item.defaultApplied( 0 ) ) return Phase::OIL;
+
+    const auto& key = uppercase( item.getTrimmedString( 0 ) );
+    if( key == "OIL" ) return Phase::OIL;
+    if( key == "GAS" ) return Phase::GAS;
+
+    throw std::invalid_argument( "Hysteresis Wetting phase must be 'OIL' or 'GAS'" );
+}
+
+bool yesno( const DeckItem& item ) {
+    if( item.defaultApplied( 0 ) ) return false;
+
+    const auto& key = item.get< std::string >( 0 );
+    if( key.front() == 'Y' || key.front() == 'y' ) return true;
+    if( key.front() == 'N' || key.front() == 'n' ) return false;
+
+    throw std::invalid_argument( "Hysteresis boolean flag must be 'YES' or 'NO'" );
+}
+
+}
+
+HysteresisOptions::HysteresisOptions( const Deck& deck ) {
+    if( !deck.hasKeyword( "EHYSTR" ) ) return;
+
+    const auto& record = deck.getKeyword( "EHYSTR" ).getRecord( 0 );
+
+    this->enabled = true;
+    this->curvature = record.getItem( 0 ).get< double >( 0 );
+    this->hys_model = readHysteresisModel( record.getItem( 1 ).get< int >( 0 ) );
+    this->curvature_killough = record.getItem( 2 ).get< double >( 0 );
+    this->modification_killough = record.getItem( 3 ).get< double >( 0 );
+
+    const int apply_to = readHysteresisApply( record.getItem( 4 ) );
+    this->apply_capillary_pressure = apply_to == 0 || apply_to == 1;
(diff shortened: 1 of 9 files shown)


opm/parser/eclipse/EclipseState/SimulationConfig/SimulationConfig.hpp
```diff
@@ -31,6 +31,15 @@ namespace Opm {
 
     public:
 
+        enum Kr_model {
```
joakim-hove: Call it e.g. `RelpermModel`; the symbol `Kr` is not obvious.

jokva: It'll be extended with relperm options (from the SATOPTS keyword), and that's a part of the change.

jokva:
Sqash'd

opm/parser/eclipse/EclipseState/SimulationConfig/SimulationConfig.hpp
```diff
@@ -20,13 +20,51 @@
...
+            RelpermOptions( const Deck& );
+
+            bool directional() const noexcept;
+            bool nondirectional() const noexcept;
+            bool reversible() const noexcept;
+            bool irreversible() const noexcept;
+            bool hysteresis() const noexcept;
+            bool surftens() const noexcept;
+
+            enum relperm_model {
```
joakim-hove: `RelpermModel`

opm/parser/eclipse/EclipseState/SimulationConfig/HysteresisOptions.cpp
```diff
@@ -0,0 +1,230 @@
+#include <opm/parser/eclipse/EclipseState/SimulationConfig/HysteresisOptions.hpp>
+#include <opm/parser/eclipse/Utility/String.hpp>
+
+namespace Opm {
+
+namespace {
+
+HysteresisOptions::Model readHysteresisModel( int x ) {
+    using Model = HysteresisOptions::Model;
```
andlaus: Is the `Model::` part mandatory (i.e., a enum class)? IMO this will look weird in downstream code where a model usually is the code which discretizes the PDEs...

opm/parser/eclipse/EclipseState/SimulationConfig/HysteresisOptions.cpp
```diff
@@ -0,0 +1,230 @@
...
+        case  0: return Model::CarlsonHysteresisNonWettingDrainageWetting;
+        case  1: return Model::CarlsonHysteresisNonWettingImbibitionNonWetting;
+        case  2: return Model::KilloughHysteresisNonWettingDrainageWetting;
+        case  3: return Model::KilloughHysteresisNonWettingImbibitionWetting;
+        case  4: return Model::KilloughHysteresisNonWettingKilloughHysteresisWetting;
+        case  5: return Model::CarlsonNonWettingGasWaterDrainageOil;
+        case  6: return Model::KilloughNonWettingGasWaterDrainageOil;
+        case  7: return Model::KilloughHysteresisNonWettingGasWaterWettingOil;
+        case  8: return Model::JargonHysteresisNonWettingDrainageWetting;
+        case  9: return Model::JargonHysteresisNonWettingImbibitionWetting;
```
andlaus: I guess there's quite some work left for opm-material to implement all of these ;)

opm/parser/eclipse/EclipseState/SimulationConfig/HysteresisOptions.cpp
```diff
@@ -0,0 +1,230 @@
...
+        default:
+            throw std::invalid_argument( "Hysteresis model must be an integer [-1, 9]" );
+    }         
+}
+
+int readHysteresisApply( const DeckItem& item ) {
```
andlaus: shouldn't these helper functions be members of `HysteresisOptions`?

opm/parser/eclipse/EclipseState/SimulationConfig/HysteresisOptions.cpp
```diff
@@ -0,0 +1,230 @@
...
+    if( item.defaultApplied( 0 ) ) return Phase::OIL;
+
+    const auto& key = uppercase( item.getTrimmedString( 0 ) );
+    if( key == "OIL" ) return Phase::OIL;
+    if( key == "GAS" ) return Phase::GAS;
+
+    throw std::invalid_argument( "Hysteresis Wetting phase must be 'OIL' or 'GAS'" );
+}
+
+bool yesno( const DeckItem& item ) {
```
andlaus: funny function name. maybe `isYesOrNo()`. it is also not specific to the hysteresis stuff, so possibly it makes sense to have this directly in the `DeckItem` class...

Also is it correct to only consider the first character? it currently returns true for strings like `yabba daba doo` or `njet`!

opm/parser/eclipse/EclipseState/SimulationConfig/HysteresisOptions.cpp
```diff
@@ -0,0 +1,230 @@
...
+HysteresisOptions::HysteresisOptions( const Deck& deck ) {
+    if( !deck.hasKeyword( "EHYSTR" ) ) return;
+
+    const auto& record = deck.getKeyword( "EHYSTR" ).getRecord( 0 );
+
+    this->enabled = true;
+    this->curvature = record.getItem( 0 ).get< double >( 0 );
+    this->hys_model = readHysteresisModel( record.getItem( 1 ).get< int >( 0 ) );
+    this->curvature_killough = record.getItem( 2 ).get< double >( 0 );
+    this->modification_killough = record.getItem( 3 ).get< double >( 0 );
```
andlaus: I suppose that the `this->` is superfluous. Also, isn't the opm-parser convention for private attributes `m_camelCase`?

opm/parser/eclipse/EclipseState/SimulationConfig/HysteresisOptions.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+    const int apply_to = readHysteresisApply( record.getItem( 4 ) );
+    this->apply_capillary_pressure = apply_to == 0 || apply_to == 1;
```
andlaus: brackets please

opm/parser/eclipse/EclipseState/SimulationConfig/HysteresisOptions.cpp
```diff
@@ -0,0 +1,230 @@
...
+    this->apply_relative_permeability = apply_to == 0 || apply_to == 2;
+
+    this->curve_shape = readHysteresisCurveShape( record.getItem( 5 ) );
+    this->apply_imbibition = readHysteresisCorrectImbibition( record.getItem( 6 ) );
+    this->wetting_phase = readHysteresisWettingPhase( record.getItem( 7 ) );
+    this->baker_oil = yesno( record.getItem( 8 ) );
```
andlaus: This may be wrong, because `yesno("n")` returns true.

opm/parser/eclipse/EclipseState/SimulationConfig/HysteresisOptions.cpp
```diff
@@ -0,0 +1,230 @@
...
+    this->curvature_killough = record.getItem( 2 ).get< double >( 0 );
+    this->modification_killough = record.getItem( 3 ).get< double >( 0 );
+
+    const int apply_to = readHysteresisApply( record.getItem( 4 ) );
+    this->apply_capillary_pressure = apply_to == 0 || apply_to == 1;
+    this->apply_relative_permeability = apply_to == 0 || apply_to == 2;
+
+    this->curve_shape = readHysteresisCurveShape( record.getItem( 5 ) );
+    this->apply_imbibition = readHysteresisCorrectImbibition( record.getItem( 6 ) );
+    this->wetting_phase = readHysteresisWettingPhase( record.getItem( 7 ) );
```
andlaus: why do you use the item indices instead of the item names?

opm/parser/eclipse/EclipseState/SimulationConfig/HysteresisOptions.cpp
```diff
@@ -0,0 +1,230 @@
...
+    this->threshold_saturation = record.getItem( 11 ).get< double >( 0 );
+    this->modify_wetting_relperm = record.getItem( 12 ).get< int >( 0 );
+
+    if( this->curvature < 0 )
+        throw std::invalid_argument(
+                "EHYSTR curvature parameter for capillary pressure curves "
+                "must be larger than zero" );
+}
+
+HysteresisOptions::operator bool() const noexcept {
```
andlaus: IMO it is better to avoid that operator because it occludes things (i.e., an `enabled()` method is better in my opinion) and my experience with the jfunc code was that the container class threw an exception at me once I tried to access the object if it was not enabled, so I could not use it anyway.

opm/parser/eclipse/EclipseState/SimulationConfig/HysteresisOptions.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+bool HysteresisOptions::applyToBothCapillaryPressureAndRelativePermeability() const noexcept {
+    return this->applyToCapillaryPressure() && this->applyToRelativePermeability();
+}
+
+HysteresisOptions::CapillaryPressureScanningCurveShape HysteresisOptions::curveShapesOnSecondaryReversal() const noexcept {
+    return this->curve_shape;
+}
+
+bool HysteresisOptions::initialMobilityCorrectionAppliesToDrainageCurve() const noexcept {
```
andlaus: stupid question: what is the mobility correction for the drainage curve?

opm/parser/eclipse/EclipseState/SimulationConfig/HysteresisOptions.hpp
```diff
@@ -0,0 +1,93 @@
...
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <iosfwd>
```
andlaus: this should be moved to after the guard macro.

atgeirr:
I just looked briefly at this: is the purpose to internalize (EclipseState-ify...) these parameters to avoid direct Deck access in clients? If so, where is the code it "replaces" (or makes simple(r) to implement)?

opm/parser/eclipse/EclipseState/SimulationConfig/HysteresisOptions.hpp
```diff
@@ -0,0 +1,93 @@
+#ifndef OPM_HYSTERESIS_OPTIONS_HPP
+#define OPM_HYSTERESIS_OPTIONS_HPP
```
andlaus: so far, guard macros have always been located after the copyright comment...

jokva:
@atgeirr In material.

joakim-hove:

PR #1028, opened 2017-01-10
Added FIP_PROBE, we now accept all FIPxxxx keywords

pgdr:
* Deleted definition for `FIPNUM`
* Added generic probe keyword `FIPxxxx`.

A keyword is a `FIPxxxx` if it has prefix `FIP` and it is not `FIPOWG`.

_Motivation: Support "keyword" `FIPREG` from some internal model.._


opm/parser/eclipse/EclipseState/Grid/GridProperties.cpp
```diff
@@ -36,6 +36,17 @@ namespace Opm {
...
+    template<typename> bool isFipxxx( const std::string& ) { return false; }
+
+    template<>
+    bool isFipxxx< int >(const std::string& keyword) {
+        // FIPxxxx can be any keyword, e.g. FIPREG or FIPXYZ that has the pattern "FIP.+"
+        // However, it can not be FIPOWG as that is an actual keyword.
+        if (keyword.size() < 4 || keyword == "FIPOWG") {
+            return false;
+        }
+        return keyword[0] == 'F' && keyword[1] == 'I' && keyword[2] == 'P';
```
andlaus: can uppercase be assumed here?

jokva: Yes. All keywords are uppercased on construction.

PR #1029, opened 2017-01-11
Added method Runspec::eclPhaseMask( )

joakim-hove:
Small utility


opm/parser/eclipse/EclipseState/Runspec.hpp
```diff
@@ -59,8 +58,9 @@ class Runspec {
         const Phases& phases() const noexcept;
         const Tabdims&  tabdims() const noexcept;
         const EndpointScaling& endpointScaling() const noexcept;
+        int eclPhaseMask( ) const noexcept;
 
-    private:
+   private:
```
pgdr: why deleted space?

PR #1030, opened 2017-01-11
Made getRegion() in e3d private

pgdr:
As far as I can tell (except in ResInsight), the function `Eclipse3DProperties.getRegion(deckitem)` is not used _outside `Eclipse3DProperties`._

Made it private.


pgdr:
(fails because of timeout not because (necessarily) it is wrong)

PR #1031, opened 2017-01-12
added REGIONS to opernum kw

pgdr:


opm/parser/eclipse/EclipseState/Eclipse3DProperties.cpp
```diff
@@ -194,7 +194,7 @@ namespace Opm {
             GridProperties< int >::SupportedKeywordInfo( "IMBNUM" , 1, "1" ),
             GridProperties< int >::SupportedKeywordInfo( "MISCNUM", 1, "1" ),
             GridProperties< int >::SupportedKeywordInfo( "MULTNUM", 1, "1" ),
-            GridProperties< int >::SupportedKeywordInfo( "OPERNUM", 1, "1" ),
```
joakim-hove: Would be good to have a comment about the special behavior of `OPERNUM` where there can be different values in `GRID` and `REGIONS` - which we do not support in OPM.

PR #1032, opened 2017-01-13
(Discuss) new way to handle WCONHIST

GitPaean:
Wells under WCONHIST control should only have only one active control mode, and also a lower BHP limit when not under BHP control. Switching to BHP control is not very desirable, while we need a BHP limit there. 

It is different from the WCONPROD, which has one active control, while there are many other limits that can potentially become a control. 

It fixes OPM/opm-simulators#1011, while needs detailed testing.

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
@@ -433,8 +433,8 @@ namespace Opm {
                     properties = WellProductionProperties::prediction( record, addGrupProductionControl );
                 } else {
                     const WellProductionProperties& prev_properties = well->getProductionProperties(currentStep);
-                    double BHPLimit = prev_properties.BHPLimit;
-                    properties = WellProductionProperties::history( BHPLimit , record, m_phases);
+                    const double BHPLimit = prev_properties.BHPLimit;
+                    properties = WellProductionProperties::history(BHPLimit, record);
                 }
 
                 if (status != WellCommon::SHUT) {
--- a/opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.cpp
@@ -41,49 +41,34 @@ namespace Opm {
     {}
 
 
-    WellProductionProperties WellProductionProperties::history(double BHPLimit, const DeckRecord& record, const Phases &phases)
+    WellProductionProperties WellProductionProperties::history(const double BHPLimit, const DeckRecord& record)
     {
-        // Modes supported in WCONHIST just from {O,W,G}RAT values
-        //
-        // Note: The default value of observed {O,W,G}RAT is zero
-        // (numerically) whence the following control modes are
-        // unconditionally supported.
         WellProductionProperties p(record);
         p.predictionMode = false;
 
         namespace wp = WellProducer;
-        if(phases.active(Phase::OIL))
-            p.addProductionControl( wp::ORAT );
-
-        if(phases.active(Phase::WATER))
-            p.addProductionControl( wp::WRAT );
-
-        if(phases.active(Phase::GAS))
-            p.addProductionControl( wp::GRAT );
-
-        for( auto cmode : { wp::LRAT, wp::RESV, wp::GRUP } ) {
-            p.addProductionControl( cmode );
-        }
-
-        /*
-          We do not update the BHPLIMIT based on the BHP value given
-          in WCONHIST, that is purely a historical value; instead we
-          copy the old value of the BHP limit from the previous
-          timestep.
-
-          To actually set the BHPLIMIT in historical mode you must
-          use the WELTARG keyword.
-        */
-        p.BHPLimit = BHPLimit;
 
         const auto& cmodeItem = record.getItem("CMODE");
         if (!cmodeItem.defaultApplied(0)) {
-            const auto cmode = WellProducer::ControlModeFromString( cmodeItem.getTrimmedString( 0 ) );
+            const std::string cmode_string = cmodeItem.getTrimmedString( 0 );
+            const auto cmode = WellProducer::ControlModeFromString(cmode_string);
 
-            if (p.hasProductionControl( cmode ))
+            if (cmode == wp::LRAT || cmode == wp::RESV || cmode == wp::ORAT ||
+                cmode == wp::WRAT || cmode == wp::GRAT || cmode == wp::BHP) {
+                p.addProductionControl( cmode );
                 p.controlMode = cmode;
-            else
-                throw std::invalid_argument("Setting CMODE to unspecified control");
+            } else {
+                const std::string msg = "unsupported control mode " + cmode_string + " for WCONHIST";
+                throw std::invalid_argument(msg);
+            }
+
+            if (cmode == wp::BHP) {
+                p.BHPLimit = record.getItem( "BHP" ).getSIDouble( 0 );
+            } else {
+                // When the control mode is not BHP, we should add a BHP limit as 1 atm.
+                p.addProductionControl( wp::BHP );
+                p.BHPLimit = BHPLimit;
+            }
         }
 
         return p;
--- a/opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.hpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.hpp
@@ -49,7 +49,7 @@ namespace Opm {
         bool operator!=(const WellProductionProperties& other) const;
         WellProductionProperties();
 
-        static WellProductionProperties history(double BHPLimit, const DeckRecord& record, const Phases &phases = Phases(true, true, true) );
+        static WellProductionProperties history(const double BHPLimit, const DeckRecord& record);
         static WellProductionProperties prediction( const DeckRecord& record, bool addGroupProductionControl );
 
         bool hasProductionControl(WellProducer::ControlModeEnum controlModeArg) const {
--- a/opm/parser/eclipse/EclipseState/Schedule/tests/ScheduleTests.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/tests/ScheduleTests.cpp
@@ -984,9 +984,8 @@ BOOST_AUTO_TEST_CASE(changeBhpLimitInHistoryModeWithWeltarg) {
     Schedule schedule(parseContext , grid, deck, Phases(true, true, true) );
     auto* well_p = schedule.getWell("P");
 
-    BOOST_CHECK_EQUAL(well_p->getProductionProperties(0).BHPLimit, 0); //start
+    BOOST_CHECK_EQUAL(well_p->getProductionProperties(0).BHPLimit, 0.); //start
     BOOST_CHECK_EQUAL(well_p->getProductionProperties(1).BHPLimit, 50 * 1e5); // 1
-    // The BHP limit should not be effected by WCONHIST
     BOOST_CHECK_EQUAL(well_p->getProductionProperties(2).BHPLimit, 50 * 1e5); // 2
 
     auto* well_i = schedule.getWell("I");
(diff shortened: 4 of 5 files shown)


pgdr:
Just a comment, 3 tests in opm-parser fail under this PR.
```
	 28 - runScheduleTests (Failed)
	 30 - runWellPropertiesTests (Failed)
	 95 - runScheduleCreateFromDeck (Failed)
```
Log:

```
95/150 Test #95: runScheduleCreateFromDeck ................***Failed 0.39 sec
Running 24 test cases...
28/150 Test  #28: runScheduleTests .........................***Failed    0.68 sec
Running 38 test cases...

/opm-parser/opm/parser/eclipse/EclipseState/Schedule/tests/ScheduleTests.cpp(990):
error in "changeBhpLimitInHistoryModeWithWeltarg": check
well_p->getProductionProperties(2).BHPLimit == 50 * 1e5 failed [101325 !=
5000000]

unknown location(0): fatal error in "changeModeWithWHISTCTL":
std::invalid_argument: Tried to set invalid control: RESV for well: P1

/opm-parser/opm/parser/eclipse/EclipseState/Schedule/tests/ScheduleTests.cpp(1004):
last checkpoint

*** 2 failures detected in test suite "ScheduleTests"



30/150 Test #30: runWellPropertiesTests ...................***Failed 0.08 sec

Running 7 test cases...

/opm-parser/opm/parser/eclipse/EclipseState/Schedule/tests/WellPropertiesTests.cpp(134):
error in "WCH_All_Specified_BHP_Defaulted": check
p.hasProductionControl(Opm::WellProducer::WRAT) failed

/opm-parser/opm/parser/eclipse/EclipseState/Schedule/tests/WellPropertiesTests.cpp(135):
error in "WCH_All_Specified_BHP_Defaulted": check
p.hasProductionControl(Opm::WellProducer::GRAT) failed
...
...
...
/opm-parser/opm/parser/eclipse/EclipseState/Schedule/tests/WellPropertiesTests.cpp(208):
error in "WCH_Rates_Defaulted_BHP_Specified": check
p.hasProductionControl(Opm::WellProducer::GRAT) failed

/opm-parser/opm/parser/eclipse/EclipseState/Schedule/tests/WellPropertiesTests.cpp(209):
error in "WCH_Rates_Defaulted_BHP_Specified": check
p.hasProductionControl(Opm::WellProducer::LRAT) failed

/opm-parser/opm/parser/eclipse/EclipseState/Schedule/tests/WellPropertiesTests.cpp(218):
error in "WCH_Rates_Defaulted_BHP_Specified": check false ==
p.hasProductionControl(Opm::WellProducer::BHP) failed [false != true]

/opm-parser/opm/parser/eclipse/EclipseState/Schedule/tests/WellPropertiesTests.cpp(224):
error in "BHP_CMODE": exception std::invalid_argument is expected

*** 26 failures detected in test suite "WellPropertiesTest"


/opm-parser/opm/parser/eclipse/IntegrationTests/ScheduleCreateFromDeck.cpp(175):
error in "WellTesting": check prop3.hasProductionControl(WellProducer::GRAT)
failed

/opm-parser/opm/parser/eclipse/IntegrationTests/ScheduleCreateFromDeck.cpp(176):
error in "WellTesting": check prop3.hasProductionControl(WellProducer::WRAT)
failed

*** 2 failures detected in test suite "ScheduleIntegrationTests"
```

GitPaean:
Thanks for pointing out. The tests need to change to adapt to the new way. I will update them on Monday.

joakim-hove:
> Just a comment, 3 tests in opm-parser fail under this PR.

I realize I will not be able to subscribe all the members of the OPM community to the church of TDD, but failing to even run the tests before submitting a PR is not good enough :angry:. It might be unfair - but when reviewing a PR with many failing tests I am quite skeptical from the onset. 

Rant aside: The exact semantics of the `WCONHIST` keyword is not very clear (to me at least), we have been going through this earlier and I fear the current PR is ~reverting previous changes.  The `WCONHIST` keyword serves two purposes:

1. It is used to enter a control mode and a corresponding rate for that control mode.
2. It is optionally - although quite common - used to enter historical rates for non-controlmode phases, that is how we can get the `xxxH` keywords in the output.
3. A historical `BHP` can be set with the `WCONHIST` keyword, that is passed to `WBHPH` in output, and not at all used to enforce a BHP limit to the simulator. If the well is BHP limited that must be enforced with `WELTARG`.

So although I certainly acknowledge that there can be problems with the `WCONHIST` implementation the current patch will not be merged, at the least it will destroy the output of history keywords. I think maybe the problem is more in the interaction between `struct wells` and `Well`. Maybe `struct wells` needs to be aware of history mode - or even better just remove `struct wells` and `WellsManager` completely?

OPMUSER:
Joakim,

Points (1) to (3) are correct.

Normally a fluid rate is set as the control phase and the BHP and THP 
(and the other phases) give us the "quality" of the history match.  So 
for an oilfield for example, the control phase on the WCONHIST would be 
oil rate, but water rate, gas rate, BHP and THP are entered so that one 
can plot historical data versus simulated data.

There is also an RESV option, in which all three phases are converted to 
reservoir volumes and the well produces that volume. This is used to 
validate the pressure profile and in-place volumes from a material 
balance prospective. So even if the fluid rates are incorrect the 
pressure profile should be in reasonable agreement with the measured data.
ThunderBird Signature File

GitPaean:
Sorry for the testing issue. I have not addressed them when the PR is still in the discussion mode, since the tests need to be changed based on the understanding. 

The following is a few understanding from my understanding for this PR. 
1. There is clearly a BHP control mode for WCONHIST. When specified with BHP control mode, the wells are supposed to produce under observed BHP values. 
2. When not specified under BHP control, there will be a default BHP limit (1 atm.). Under this kind of circumstance, switching to BHP control is a bad sign, while it still possible. It should serve as a limit. 
3. We will always have observed oil rate, water rate, and gas rate. The rates are used to calculate the LRAT, RESV targets, or other targets based on the specification. while they are not always limits. For example, if a well is under ORAT control mode, it will not switch to WRAT control at all no matter what water rate is produced and what the observed water rate is. As a result, the water rate should not enter well_control as a limit at all, there should be no WRAT control mode when parsing WCONHIST with ORAT specification.

OPMUSER:
Kai,

Not true, WRAT can be a control mode as well, although very seldom used.:



WGRA and NGL are Eclipse 300 options.

Eclipse 100 CRAT, is new too me, now idea what this for :-)

ThunderBird Signature File

GitPaean:
I mean, when we specify with ORAT within WCONHIST, the observed water rate will not work as a limit. 

If it work as a limit, then potentially, the well can switch to WRAT control.  It is the case for WCONPROD, while not for WCONHIST.

joakim-hove:
> Points (1) to (3) are correct.

Thank you for contributing to this.

GitPaean:
> I think maybe the problem is more in the interaction between struct wells and Well. Maybe struct wells needs to be aware of history mode - or even better just remove struct wells and WellsManager completely?

We can handle it through the opm-core (like closed OPM/opm-core#1135) and opm-simulators (changing the function `resvWells`).  But I think it can be better to correct the understanding from parser. 

And it is also kind of significant. With the current implementation, if we specifies ORAT for WCONHIST, while we did not provide correct water rate and gas rate, our results will be wrong. Since we handle it with RESV way, while the other two rates are wrong.

GitPaean:
> 3. A historical BHP can be set with the WCONHIST keyword, that is passed to WBHPH in output, and not at all used to enforce a BHP limit to the simulator. If the well is BHP limited that must be enforced with WELTARG.

I think when we specify BHP control for WCONHIST, the observed BHP value will be set as BHP target. 
Anyway, we can discuss this later.  I am not sure when specified under BHP control, how it will interact with WELTARG. 

The other first two points are correct to me. 

The current implementation of BHP part is wrong.  It should be 1 atm. at the first time, can be alternated by WELTARG, later WCONHIST will not change it if not specified under BHP control.

GitPaean:
The tests part are not fixed yet. Tests will still be failed. Sorry about that.

joakim-hove:
OK, this is my understanding of the WCONHIST keyword - I am not confident enough to tattoo these claims on my back.

**BHP:** Wells which are run in historical mode can have a BHP limit, however that limit must be set with the `WELTARG`keyword, the BHP value entered as part of the `WCONHIST`keyword is only to record *historical* values of BHP.

**Rates:** As part of the `WCONHIST` keyword we (typically) specify a control mode, i.e. `ORAT`, and historical rates for all the phases. In the case of control `ORAT` the oil rate specified as part of the `WCONHIST` keyword is used to drive the simulator, the water and gas rates (if specified) are just passed through to the output layer to provide values for the `WPRH` and `GPRH` keywords, however as part of the implementation in opm-parser `GRAT` and `WRAT` become available as control modes because values are entered for these phases.

**Problem?** Since the implementation of the `WCONHIST` keyword makes e.g. `WRAT` and `GRAT`available as control modes the simulator is free to change to one of these modes. When in in historical mode we do *not* want the simulator to be allowed to change control mode. 

**Iff** my understanding of the problem is correct I think we need to increase the semantic awareness of *historical* mode in the well-control modelling; I guess one alternative would be that the parser internalized the non-control-mode rates like `GRAT` and `WRAT` in the example above as *historical rates* - and not rates which can be used to control the simulator.

joakim-hove:
>I think when we specify BHP control for WCONHIST, the observed BHP value will be set as BHP target.
Anyway, we can discuss this later. I am not sure when specified under BHP control, how it will interact with WELTARG.

We will have to read the *very fine print* of the manual, you are probably right. I think the main issue here is that the even in historical mode, where changes in control mode are generally not allowed, the well can change to `BHP`control - in that case the `BHP` limit comes from the `WELTARG` keyword. However *iff* `BHP` control is specified in the `WCONHIST` keyword, the `BHP`control limit is taken from the `WCONHIST`keyword? What a mess!

GitPaean:
I agree almost everything in the above two comments except that the BHP limit is always there in my opinion (either defaulted 1 atm.  or specified by WELTARG or the observed value when under BHP control). I totally agree that BHP limit related is a little messy, some part needs a few tests to tell. 

Let us continue the discussion next week.  Thanks for the comments.

OPMUSER:
The "mess" is because WCONHIST was added as an additional feature 
(history match control mode, as oppose to WCONPROD  control mode) and so 
was WELTARG (this was added so that one didn't have to use the WCONPROD 
keyword all the time just to change rates) . Now you also have a 
WHISTCTL keyword which operates on WCONHIST control wells (like 
WCONPROD/WELTARG), but it just changes the control mode (not the 
value,). This is why you constantly see these type of inconsistencies. 
The Eclipse developers have to to maintain backward compatibility, while 
at the same time add additional functionality with without breaking the 
existing decks.

BHP has to be defined as it is used to calculate well potentials, with 
the well potentials used to allocate rates for wells under group 
control. So if a well is under WCONHIST control the BHP defaults to 
values that are non-restrictive, and the way to change this is to reset 
the BHP using WELTARG. The WELTARG can only change the BHP for WCONHIST 
controlled wells. Hindsight would suggest it would have been better to 
place the BHP value on the WHISTCTL keyword instead of the WELTARG 
keyword, but may be WELTARG was available at the time and WHISTCTL came 
later.

This may appear confusing, but the defaults is what engineers want to 
happen. They don't want want the wells switching to BHP control when 
history matching well performance until the history match is "close", 
the final set of iterations (after the fluid and general pressure match) 
is to "tune" the well BHP. by calibrating the well's productivity index.

The recommended approach for switching between historical and prediction 
controls is by applying the last flowing bottom hole pressure at the 
instant the transition takes place as the production target while the 
latest flow rates are retained as phase limits. This ensures a smooth 
transition between history and prediction without having to resort to 
unreasonable changes to the model. In ECLIPSE this is achieved by 
specifying the below at the beginning of the restart run (forecast)

WELTARG
* BHP /
/

A second alternative that achieves the same effect is
WELCNTL
* BHP /
/
ThunderBird Signature File

joakim-hove:
So - anything happening here?

GitPaean:
Nope yet. We are putting something else in much higher priority at the moment. It is not easy to put efforts here yet.  The BHP limit for this one can be a little troublesome in the parser side, while it should not be huge. while the major challenge will be testing through all the history matching cases we have and we were successful in.

joakim-hove:
I'll close this now - as it seems quite stale; please repoen if still relevant.

GitPaean:
Reopening it to work on it.

GitPaean:
Will open a new one to make the discussion more clear, while referring to  this one.

PR #1035, opened 2017-01-26
Special case unity unit transformations.

joakim-hove:
This a minor preparation to facilitate bitwise correct restarts.

### DIFF
--- a/opm/parser/eclipse/Units/UnitSystem.cpp
+++ b/opm/parser/eclipse/Units/UnitSystem.cpp
@@ -457,24 +457,38 @@ namespace {
     }
 
     double UnitSystem::from_si( measure m, double val ) const {
-        return this->measure_table_from_si[ static_cast< int >( m ) ] * val;
+        if (m == measure::identity)
+            return val;
+        else
+            return this->measure_table_from_si[ static_cast< int >( m ) ] * val;
     }
 
     double UnitSystem::to_si( measure m, double val ) const {
-        return this->measure_table_to_si[ static_cast< int >( m ) ] * val;
+        if (m == measure::identity)
+            return val;
+        else
+            return this->measure_table_to_si[ static_cast< int >( m ) ] * val;
     }
 
     void UnitSystem::from_si( measure m, std::vector<double>& data ) const {
-        double factor = this->measure_table_from_si[ static_cast< int >( m ) ];
-        auto scale = [=](double x) { return x * factor; };
-        std::transform( data.begin() , data.end() , data.begin() , scale);
+        if (m == measure::identity)
+            return;
+        {
+            double factor = this->measure_table_from_si[ static_cast< int >( m ) ];
+            auto scale = [=](double x) { return x * factor; };
+            std::transform( data.begin() , data.end() , data.begin() , scale);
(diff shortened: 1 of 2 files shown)


joakim-hove:
jenkins build this with downstreams ert=1406 please

jokva:
Isn't identity already == 1 in the unit table?

joakim-hove:
> Isn't identity already == 1 in the unit table?

Absolutely - and this whole PR might be unnecessary - the reason was that I wanted a guarantee that 
```
units::from_si( measure::identity , x ) == x
```
bitwise. I don't know if the IEEE representation is guaranteed to represent 1.0 exactly?

jokva:
You can assume that.

>  Despite these examples, there are useful optimizations that can be done on floating-point code. First of all, there are algebraic identities that are valid for floating-point numbers. Some examples in IEEE arithmetic are x + y = y + x, 2 ×  x = x + x, 1 × x = x, and 0.5× x = x/2. However, even these simple identities can fail on a few machines such as CDC and Cray supercomputers. Instruction scheduling and in-line procedure substitution are two other potentially useful optimizations.

https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html

A question though: why is bitwise equality important?

joakim-hove:
> A question though: why is bitwise equality important?

That was a minor preparation for bitwise correct restart.

joakim-hove:
OK - @jokva: feel free to close this if you feel that is appropriate; or merge it of course. As long as I have a reasonable IEEE guarantee that:
```txt
1.0 *x == x
```
I am satisfied.

jokva:
It doesn't seem like it's *mandated* by the standard (because they very rarely mandate these kinds of properties), but as far as potential portability issues go this seems like a non-problem. If proper portability and bitwise-representation-on-disk actually becomes an issue we can use `numeric_limits` to statically determine whether or not some particular representation is used, but regardless we need to do something to make restart files (or others) portable between platforms, since as of right now everything just dumps its native-ish float format.

akva2:
IEEE guarantees exact representation of +- 2^0, but all bets are off after a flop.

joakim-hove:
> IEEE guarantees exact representation of +- 2^0, but all bets are off after a flop.

So @akva2: what is your opinion - does this PR make sense, or is it just over something? I am happy to see it closed if it does not make sense.

akva2:
i do not think this will help much in practice. it would just aid a synthetic test of what you wrote above ^^. all these numbers will have calculations performed on them "downstream". the only case where this guarantee is useful is that you can do default values like 1.0, and if no operations are done you are guaranteed that you can use if (a == 1.0) type checks.

joakim-hove:
> all these numbers will have calculations performed on them "downstream"

Well - in my opinion it should be possible to formulate the restart code so that the bytes lifted from disk are passed all the way to the simulator with a single flop.

akva2:
probably, but then we would need to some derived data as well, which is regenerated in the simulator upon restart.

joakim-hove:
>but then we would need to store some derived data as well, which is regenerated in the simulator upon restart.

Closing this for now.

Issue #1033, opened 2017-01-18
compatibility of windows and linux directory separators


hnil:
Given an input deck with include statement
INCLUDE
'..\MYDIR\MYFILE.INC'

which should work on windows to not work on linux where the normal way of writing would be

INCLUDE
'../MYDIR/MYFILE.INC'

Handling this case would make transfer of files between windows and linux simpler...

joakim-hove:
> Handling this case would make transfer of files between windows and linux simpler...

Hmmm - I must admit I do not fully understand your question. My understanding of the situation regarding path separators is as follows:

1. opm-parser is using `boost::filesystem` for path operations; I assume that works transparently on Windows and Linux.

2. I *thought* Windows actually handled both `/` and `\` as path separator? 
 
> Handling this case would make transfer of files between windows and linux simpler...

Do you really mean *transfer of files* - or do you rather mean running the same deck on both windows and linux?

atgeirr:
> Do you really mean transfer of files - or do you rather mean running the same deck on both windows and linux?

The second, running a deck with backslashes in it on Linux.

joakim-hove:
NO - on Linux the path separator is `/` - on the other I think maybe Windows also will handle `/`. If windows can not handle `/` as path separator your best bet would be to use s script to convert between the two.

Sorry - closing.

atgeirr:
I think, that if Eclipse on Linux supports backslash separators, perhaps we should support it too? Do you know if it does?

OPMUSER:
Yes, it does but it does not handle spaces in directory or file names on 
Linux.
ThunderBird Signature File

Regards, David Baxendale Email: David.Baxendale@gmail.com

------------------------------------------------------------------------

atgeirr:
> Yes, it does

Would it be a big effort to add this to opm-parser, @joakim-hove?

> but it does not handle spaces in directory or file names on
> Linux.

That is a strange difference, but not one we need to copy I think...

joakim-hove:
> I think, that if Eclipse on Linux supports backslash separators, perhaps we should support it too? 

I guess I spoke my mind quite clearly by closing the case - but of course if the community as a whole argues that we should indeed support `\` as path separator on Linux we will of course look into it. I have no experience with this - but I feel quite uncomfortable implementing a file system semantics which is actually in conflict with the underlying platform.

[...]

PR #1039, opened 2017-02-09
Replace backslash in INCLUDE paths

atgeirr:
The following addresses #1033 by converting backslashes in include path strings to slashes. This is done after all other processing such as expanding mapped paths (of the `$PATH_KEY` kind), just before creating the boost::filesystem::path object.

I believe it is a better solution than using a conversion script since:
 - requiring such a script is likely to reduce usage by casual users (it doubles the number of commands to remember...)
 - writing such a script in such a way that it only does what it is supposed to do might be hard, the current solution can only affect INCLUDE paths
 - we can (and do) give a warning about this, so that users can choose to make the paths universal (using slashes) but do not require it

@hnil can you verify that this PR solves the observed problem?

### DIFF
--- a/opm/parser/eclipse/Parser/tests/ParserTests.cpp
+++ b/opm/parser/eclipse/Parser/tests/ParserTests.cpp
@@ -288,8 +288,8 @@ BOOST_AUTO_TEST_CASE( PATHS_has_global_scope ) {
     ParseContext parseContext;
 
     parseContext.update( ParseContext::PARSE_MISSING_INCLUDE , Opm::InputError::THROW_EXCEPTION);
-    parser.parseFile( "testdata/parser/PATHSInInclude.data", parseContext );
-    BOOST_CHECK(parser.hasKeyword("OIL"));
+    const auto deck = parser.parseFile( "testdata/parser/PATHSInInclude.data", parseContext );
+    BOOST_CHECK(deck.hasKeyword("OIL"));
     BOOST_CHECK_THROW( parser.parseFile( "testdata/parser/PATHSInIncludeInvalid.data", ParseContext() ), std::invalid_argument );
 }
 
@@ -298,8 +298,8 @@ BOOST_AUTO_TEST_CASE( PATHS_with_backslashes ) {
     ParseContext parseContext;
 
     parseContext.update( ParseContext::PARSE_MISSING_INCLUDE , Opm::InputError::THROW_EXCEPTION);
-    parser.parseFile( "testdata/parser/PATHSWithBackslashes.data", parseContext );
-    BOOST_CHECK(parser.hasKeyword("OIL"));
+    const auto deck = parser.parseFile( "testdata/parser/PATHSWithBackslashes.data", parseContext );
+    BOOST_CHECK(deck.hasKeyword("OIL"));
 }
 
 BOOST_AUTO_TEST_CASE( handle_empty_title ) {
--- a/opm/parser/eclipse/Parser/tests/ParserTests.cpp
+++ b/opm/parser/eclipse/Parser/tests/ParserTests.cpp
@@ -289,6 +289,7 @@ BOOST_AUTO_TEST_CASE( PATHS_has_global_scope ) {
 
     parseContext.update( ParseContext::PARSE_MISSING_INCLUDE , Opm::InputError::THROW_EXCEPTION);
     parser.parseFile( "testdata/parser/PATHSInInclude.data", parseContext );
+    BOOST_CHECK(parser.hasKeyword("OIL"));
     BOOST_CHECK_THROW( parser.parseFile( "testdata/parser/PATHSInIncludeInvalid.data", ParseContext() ), std::invalid_argument );
 }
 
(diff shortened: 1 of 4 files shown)


andlaus:
on a slightly related topic: are there serious -- preferably technical -- objections to getting rid of Boost::filesystem? If no, I offer to have a look what it takes.

joakim-hove:
Thank you for adding test. 'll hold my nose and merge this but first:

1. I guess the `foo_target.inc` file has been forgotten from the PR?

2. Please add some keywords in the `foo_target.inc` file and check that they are present in the resulting `Deck` object.

joakim-hove:
I am at the top of my head not aware of serious reasons to hold on to Boost FS.

andlaus:
Ok, then I'll have a look soon.

atgeirr:
The file is already there: `testdata/parser/include/foopath/foo_target.inc` (it's already used by a different test that I copied, replacing slashes with backslashes in other files).  I have added a check that the one keyword in it is actually found (also added this to the existing INCLUDE test).

joakim-hove:
> Ok, then I'll have a look soon.

Mind you - my answer was: "I am at the top of my head not aware of serious reasons to hold on to Boost FS." - i.e. there might still be good reasons to keep Boost FS around, which I had forgotten.

opm/parser/eclipse/Parser/tests/ParserTests.cpp
```diff
@@ -298,6 +299,7 @@ BOOST_AUTO_TEST_CASE( PATHS_with_backslashes ) {
 
     parseContext.update( ParseContext::PARSE_MISSING_INCLUDE , Opm::InputError::THROW_EXCEPTION);
     parser.parseFile( "testdata/parser/PATHSWithBackslashes.data", parseContext );
+    BOOST_CHECK(parser.hasKeyword("OIL"));
```
joakim-hove: Ehhh - this just tests that the parser instance recognizes the `OIL` keyword, not that it has actually loaded that keyword into a `Deck` object. The test you need is:

```C++
const auto deck = parser.parseFile( ... );
BOOST_CHECK( deck.hasKeyword("OIL"));
```

joakim-hove:
Instead of special casing the `INCLUDE` handling I would prefer if you added a
```C++
ParserState::ParserState( const ParseContext& context, const std::string& filename) { ... }
```
overload [here](https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/Parser/Parser.cpp#L279) then we could (module shell escaping?) actually run `flow` as:

```txt
bash% flow \path\to\deck/with\slash/confusion/ECL.DATA
```
ugly - but to some extent consistent.

opm/parser/eclipse/Parser/tests/ParserTests.cpp
atgeirr: Thanks, I did not realize this distinction.

atgeirr:
> Instead of special casing the INCLUDE handling I would prefer if you added a
> 
> ParserState::ParserState( const ParseContext& context, const std::string& filename) { ... }
> overload here then we could (modulo shell escaping?) actually run flow as:
> 
> bash% flow \path\to\deck/with\slash/confusion/ECL.DATA
> ugly - but to some extent consistent.

I assume then the backslash detection/replacement should happen in that constructor. For the warning, I assume using the ParseState's deck member's message container would be correct?

joakim-hove:
> I assume then the backslash detection/replacement should happen in that constructor. 

Yes.

> For the warning, I assume using the ParseState's deck member's message container would be correct?

Yes.

To go completely overboard we could make a `ParseContext` behavior out of this - so that the behavior could be regulated - although this will be somewhat oddball there.

atgeirr:
I have tried and failed to make a ParseState::ParseState() constructor overload work for this. The reason is that in the end it is the ParseState::loadFile() method that is used to actually open an INCLUDEd file, not the constructor.

Adding the logic to loadFile() instead, and changing the constructor(s) to call loadFile() would be possible, but there is a subtlety there: loadFile takes a boost::filesystem::path and not a string, so to do the required manipulation there we'd need to extract the string, replace backslashes, then make it a path again.

While doable, it seems to me that the original proposal is far simpler, and less brittle. It does not give the ability to use backslashes in arbitrary paths (only in INCLUDE), but I don't believe that is a big drawback.

I have pushed an update to the test issue as requested. For the other issue discussed above, please advise.

joakim-hove:
> [...] The reason is that in the end it is the ParseState::loadFile() method that is used to actually open an INCLUDEd file, not the constructor.

Well - yes; so the constructor would not the intercept in the `std::string -> boost::filesystem::path` transition, and then continue processing with a valid `boost::filesystem::path` instance. Let it be if it is to much hassle.

atgeirr:
> Let it be if it is to much hassle.

Appreciated. In that case I think there are no more changes from me here unless you have further requests.

andlaus:
> Mind you - my answer was: "I am at the top of my head not aware of serious reasons to hold on to Boost FS." - i.e. there might still be good reasons to keep Boost FS around, which I had forgotten.

Ok, I had a look: there seems to be a reason to keep it, mainly having a way iterate over directories without resorting to operating system specific low-level functions. That said it looks like about 90% of the users of boost::filesystem in opm-parsers do not really seem to need...

andlaus:
note that boost::filesystem recently transformed into std::filesystem. unfortunately its only available from C++-2017 on...

atgeirr:
> std::filesystem

Hooray! What really matters most for us is not necessarily the C++ version, but which gcc version that includes it (and when that version is the default or available on our platforms). Do you know if gcc 5 has this?

andlaus:
I googled a bit for it and it seems like GCC 5 does not include it at all while GCC 6 has `std::experimental::filesystem`. looks like we have to wait a few years ;)

PR #1040, opened 2017-02-13
Store saturation tableId for completions

totto82:


joakim-hove:
Well - this is good with tests and all; but I would like a lightly different approach - sorry. I would like you to pass the `SATNUM` array - as `GridProperty<int>` reference - then the `Compeltion` object can immediately internalize the correct `STANUM` id value if it is defaulted in the input deck, then we can use a plain `int` as id value - and we will not need to keep track of the "Has been assigned / use default value" question.

totto82:
> I would like you to pass the SATNUM array - as a GridProperty<int> reference - then the Completion object can immediately internalize the correct SATNUM id value if it is defaulted in the input deck

I thought about this, but it seems to me that  SATNUM array is not easily accessible from the Completion. So I moved the default handling to opm-material. Is there a better way than passing the SATNUM array all the way from EclipseState?

joakim-hove:
> Is there a better way than passing the SATNUM array all the way from EclipseState?

No unfortunately - that is what you have to do.

totto82:
I hope this was what you wanted. The interface change of Schedule() would also break downstream code. I have not made any PR to other repos yet. So this should not be merged before the downstream code is ready.

joakim-hove:
> I hope this was what you wanted. 

Yes - this looks good; I had initially though to only pass the `GridProperty<int>& satnum` property, and not not the full property collection; but that is OK.

opm/parser/eclipse/EclipseState/Schedule/Completion.hpp
```diff
@@ -89,10 +93,12 @@ namespace Opm {
         Value<double> m_connectionTransmissibilityFactor;
         double m_wellPi;
         Value<double> m_skinFactor;
+        Value<int> m_satTableId;
```
joakim-hove: Should not be necessary to use the `Value<int>` type now, since we are guaranteed that the value will be correctly initialized in the constructor.

testdata/integration_tests/SCHEDULE/SCHEDULE_COMPDAT1
```diff
@@ -55,7 +55,7 @@ COMPDAT
 COMPDAT 
 -- WELL        I    J    K1  K2            Sat.        CF       DIAM        KH SKIN ND        DIR   Ro 
      'W_1'     3    7    1    3      'OPEN'  1*     32.948      0.311   3047.839  2*         'X'     22.100 /
-     'W_1'     3    7    2    2      'OPEN'  1*     23487.0     0.311   4332.346  2*         'X'     22.123 /
+     'W_1'     3    7    2    2      'OPEN'  2      23487.0     0.311   4332.346  2*         'X'     22.123 /
```
joakim-hove: Add a nont-trivial `SATNUM` keyword to this deck - so tat you can check the default lookup code.

joakim-hove:
> The interface change of Schedule() would also break downstream code.

Yes - but does downstream instantiate `Schedule` directly?

joakim-hove:
Please rebase this on current master before pushing again.

joakim-hove:
Very good - this will go in when Jenkins says so.

joakim-hove:
jenkins build this downstreams please

joakim-hove:

PR #1043, opened 2017-02-16
Assert grid property

joakim-hove:
The main purpose of the is PR is to support forced auto generation of keywords in the `GridProperties<T>` container from the output layer. In the opm-output we want to unconditionally write the `NTG`property to the INIT file; to ensure that the container has the keyword `NTG`- irrespective of the deck I tried:

```C++
properties.getKeyword( "NTG" );
```

but it seems the compiler optimized the whole thing away? Have in this PR added a void method:

```C++
properties.assertKeyword("NTG");
```

which does the trick.


joakim-hove:
jenkins build this with downstreams opm-parser=1043 please

andlaus:
that's weird: the simulators use quite a few other properties (e.g. PORV) this way and it seems to work fine. maybe there are subtle differences, though.

joakim-hove:
> that's weird: the simulators use quite a few other properties (e.g. PORV) this way and it seems to work fine. 

I don't know - but in the normal use of `getKeyword( )` we actually handle the return value; maybe the whole call can be optimized out when we do not assign the return value?

andlaus:
> maybe the whole call can be optimized out when we do not assign the return value?

I'd consider this to be a compiler bug. can you test if it is around with a different compiler?

anyway, I think that this PR is an improvement regardless of the NTG issue which motivated it.

joakim-hove:
I would like to merge this - and opm-output 167, objections?

pgdr:
@joakim-hove No objections.  Merge if green.

PR #1048, opened 2017-03-10
Adding 1 atma deafult BHP for keyword WCONPROD

GitPaean:
And always employing a BHP constraint there. 

Please suggest if this is the good way to do that.

The side effect is that the BHP constraint will always be the last one, potentially affects some situation that we switching well controls when some constraint is violated. I mean, for the situation that the order of the constraints in well_controls struct matter.

### DIFF
--- a/opm/parser/share/keywords/000_Eclipse100/W/WCONPROD
+++ b/opm/parser/share/keywords/000_Eclipse100/W/WCONPROD
@@ -7,7 +7,7 @@
        {"name" : "GRAT"         , "value_type" : "DOUBLE" , "default" : 0.0 , "dimension" : "GasSurfaceVolume/Time"},
        {"name" : "LRAT"         , "value_type" : "DOUBLE" , "default" : 0.0 , "dimension" : "LiquidSurfaceVolume/Time"},
        {"name" : "RESV"         , "value_type" : "DOUBLE" , "default" : 0.0 , "dimension" : "LiquidSurfaceVolume/Time"},
-       {"name" : "BHP"          , "value_type" : "DOUBLE" , "default" : 0.0 , "dimension" : "Pressure"},
+       {"name" : "BHP"          , "value_type" : "DOUBLE" , "default" : 1.01325 , "dimension" : "Pressure"},
        {"name" : "THP"          , "value_type" : "DOUBLE" , "default" : 0.0 , "dimension" : "Pressure"},
        {"name" : "VFP_TABLE"    , "value_type" : "INT"    , "default" :   0 },
        {"name" : "ALQ"          , "value_type" : "DOUBLE" , "default" : 0.0 },
--- a/opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/WellProductionProperties.cpp
@@ -107,14 +107,16 @@ namespace Opm {
         using mode = std::pair< const char*, wp::ControlModeEnum >;
         static const mode modes[] = {
             { "ORAT", wp::ORAT }, { "WRAT", wp::WRAT }, { "GRAT", wp::GRAT },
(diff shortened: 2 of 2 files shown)


opm/parser/share/keywords/000_Eclipse100/W/WCONPROD
```diff
@@ -7,7 +7,7 @@
        {"name" : "GRAT"         , "value_type" : "DOUBLE" , "default" : 0.0 , "dimension" : "GasSurfaceVolume/Time"},
        {"name" : "LRAT"         , "value_type" : "DOUBLE" , "default" : 0.0 , "dimension" : "LiquidSurfaceVolume/Time"},
        {"name" : "RESV"         , "value_type" : "DOUBLE" , "default" : 0.0 , "dimension" : "LiquidSurfaceVolume/Time"},
-       {"name" : "BHP"          , "value_type" : "DOUBLE" , "default" : 0.0 , "dimension" : "Pressure"},
+       {"name" : "BHP"          , "value_type" : "DOUBLE" , "default" : 1.01325 , "dimension" : "Pressure"},
```
atgeirr: Three questions here for @joakim-hove mostly:
 - What units should default values use in the keyword?
 - Is there a way to use the unit constants from the C++ code here somehow? (I guess not.)
 - Is there a way to put comments in these files? JSON seems to not support it, so I guess it's a no, too.

joakim-hove: > What units should default values use in the keyword?

The defaults should use metric units - when resolving defaults two different unit system are in action, the unit system of the deck, and the default unit system (which is metric).

> Is there a way to use the unit constants from the C++ code here somehow? (I guess not.)

Yes there is - see eg: https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Tables/Tabdims.hpp#L42

> Is there a way to put comments in these files?

Well - as you JSON itself does not support comments, so there is no first class support for comments. However - no one is stopping you from something like:
```json
{"name" : "BHP", 
 "value_type" : "DOUBLE" , 
 "default" : 1.01325 , 
 "dimension" : "Pressure",
 "comment"  : "The default corresponds to ...."}
```
however that will not be internalized in any way.

andlaus:
>> Is there a way to put comments in these files?
>
> Well - as you JSON itself does not support comments, so there is no first class support for comments. However - no one is stopping you from something like: [...]

I consider the comment-as-string-item hack really ugly and verbose. Since the keyword definitions are supposed to be humanly generated instead of being just used for data transfer, we should really think about switching to a JSON parser that supports comments, IMO. (e.g., [jsonCPP](https://github.com/open-source-parsers/jsoncpp).) While this would prevent our JSON files from being 100% [standards conformant](https://www.json.org), it would be much easier to document them doxygen style and it would IMO also not violate the *spirit* of JSON.

atgeirr:
> I consider the comment-as-string-item hack really ugly and verbose. Since the keyword definitions are supposed to be humanly generated instead of being just used for data transfer, we should really think about switching to a JSON parser that supports comments, IMO. (e.g., jsonCPP.) While this would prevent our JSON files from being 100% standards conformant, it would be much easier to document them doxygen style and it would IMO also not violate the spirit of JSON.

I don't think comment-as-string looks that bad, but will abide by the approach chosen for opm-parser. Both seem workable, although the comment-as-string approach has the immediate advantage of no extra work.

joakim-hove:
It is not really clear to me if you want this merged, or if more is coming? Is the suggested default value from the manual, or is it something you are suggesting - what is then the purpose?

GitPaean:
The default value is from the manual. And I believe this is the correct thing to do in the parser side. If you agree, please merge.   If anything comes up from the simulator side, we should fix from the simulator side.

GitPaean:
The purpose is to avoid the hard-coded constants in the PR OPM/opm-simulators#1088 .  It possibly leaves out the history matching wells, which should be able to get fixed in #1032 when we work on that PR.

PR #1051, opened 2017-03-15
EndpointScaling: will create default if SWATINIT

joakim-hove:


joakim-hove:
@totto82 : Will you merge this when/if you are happy.

totto82:
Thanks. Is testing this now.

joakim-hove:
Merging after communication with @totto82

akva2:
this broke the 'equil' test in opm-core.

pgdr:
Yes, we saw now that Travis actually timed out _before_ finishing opm-core.

PR #1055, opened 2017-03-29
Do not build opm-common on Windows.

joakim-hove:
opm-common now has a `#include <unistd.h>` which breaks the build on windows.


atgeirr:
Googled a little, and to restore Windows building of opm-common I think it would be sufficient to include `<io.h>` and call `_isatty` instead of `<unistd.h>` and `isatty`. Would you be able to try this? If it works we could make a small patch to fix this.

joakim-hove:
Sure - we can try that.

PR #1056, opened 2017-04-05
Well events

joakim-hove:
@atgeirr ??

### DIFF
--- a/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
+++ b/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
@@ -327,6 +327,7 @@ namespace Opm {
             const auto& record = keyword.getRecord(recordNr);
             const std::string& wellName = record.getItem("WELL").getTrimmedString(0);
             const std::string& groupName = record.getItem("GROUP").getTrimmedString(0);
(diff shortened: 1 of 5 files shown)


opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -467,7 +471,7 @@ namespace Opm {
                 updateWellStatus( *well , currentStep , status );
                 if (well->setProductionProperties(currentStep, properties))
                     m_events.addEvent( ScheduleEvents::PRODUCTION_UPDATE , currentStep);
```
atgeirr: Should not this (and injection, polymer) update also insert a well production update event?

joakim-hove: Yes - update coming.

joakim-hove:
The Polymer event might be a little too detailed?

atgeirr:
> The Polymer event might be a little too detailed?

Perhaps, I just looked for some kind of symmetry between field and well events.

joakim-hove:
OK - I assume this is *Good Enough*&trade; - we can improve it later if necessary.

PR #1057, opened 2017-04-06
Default to build static libraries.

joakim-hove:


andlaus:
I like this becaus I recently had to do this manually to get self-contained binaries. during that excercuse I also had to instruct ERT to be static by default (and thus I had to disable its python support). is a similar change ERT in the pipeline?

joakim-hove:
> during that excercise I also had to instruct ERT to be static (and thus I had to disable its python support). is a similar change of defaults for ERT in the pipeline?

The default will still be to build shared libraries for ert, but with this: https://github.com/Ensembles/ert/pull/1485 it should be sufficient to pass `-DBUILD_SHARED_LIBS=OFF` and the python wrappers will automatically be disabled.

andlaus:
ok, what I actually wanted to know is if ERT in default mode works fine with OPM in its defaults mode. I got a problem due to missing `-fPIC`, but that may be because I might have tried it the other way round. (i.e., ERT static, opm-parser dynamic.)

joakim-hove:
> ok, what I actually wanted to know is if ERT in default mode works fine with OPM in its defaults mode.

Yes - the all-default route should *just work* - otherwise I would consider it a bug.

PR #1058, opened 2017-04-18
explicitly specify /usr as a prefix path in packaging

akva2:
needed for python-cwrap find rule.

Backport from release branch.


joakim-hove:
Thank you!

PR #1059, opened 2017-04-26
changed: use the common functionality for optimization flags

akva2:
while packaging 2017.04, i realized opm builds with -mtune=native by default. this is not a good idea in generic binaries, as you end up tuning to whatever hardware happens to be available on the build slave.

in general, the option is attached to a cmake option (which defaults to true). that means i could disable it.
in opm-parser build optimization flags are more rigid, so the release is built with native tuning.

this enables the generic opm-common hosted rule to allow proper builds in the future. it should not alter anything by default (maybe order and such but the actual flags being enabled should be the same). please help me verify the last point - it's the same on my box but that's no proof.


joakim-hove:
> please help me verify the last point 

Looks good; tests pass (until timeout). In she goes.

PR #1060, opened 2017-05-04
Added 2 new message types to OpmLog (Probleminfo and Warninginfo)

nairr:
Adds new message types which can be used to provide additional information about a problem or warning without being explicitly logged as a problem or warning


nairr:
Not required anymore...closing

PR #1062, opened 2017-05-15
Downstream ecl#54

jokva:
Downstream of https://github.com/Statoil/libecl/pull/54


alfbr:
I have build breakage on opm-parser after this one. I get complaints about libecl requiring cmake 3.0 or newer when running cmake.

jokva:
You need >= 2.8.12

PR #1063, opened 2017-05-18
Basic support for cylindrical grids

joakim-hove:
With this PR the EclipseGrid implementation has limited support for radial grids. The grid must be specified using the keywords `DRV, DTHETAV, DZV`and `TOPS`. When the radial grid is specified this way everything - including simulation and ouptut migth actually *Just Work*&trade;

### DIFF
--- a/lib/eclipse/EclipseState/Grid/EclipseGrid.cpp
+++ b/lib/eclipse/EclipseState/Grid/EclipseGrid.cpp
@@ -575,7 +575,7 @@ namespace Opm {
 
     void EclipseGrid::assertVectorSize(const std::vector<double>& vector , size_t expectedSize , const std::string& vectorName) {
         if (vector.size() != expectedSize)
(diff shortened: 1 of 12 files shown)


joakim-hove:
Does anyone care? I'll just merge this tomorrow.

atgeirr:
I care, but have not tested this. I assume that it just works... About simulation just working, I would not expect that. At least we must ensure the grid has connections across the periodic boundary of the input grid.

joakim-hove:
> At least we must ensure the grid has connections across the periodic boundary of the input grid.

Yes - the actual connection must be established in opm-grid, but the `EclipseGrid` has got the new method `circle( )` which return `true` iff such a periodic connection should be added in the simulation grid.

atgeirr:
> Yes - the actual connection must be established in opm-grid, but the EclipseGrid has got the new method circle( ) which return true iff such a periodic connection should be added in the simulation grid.

Sounds good!

atgeirr:

PR #1065, opened 2017-05-24
Redesign makefiles

jokva:
Creating early PR for travis.


andlaus:
while I don't exactly consider this module to be my glass of beer, I'm still wondering what the motivation for this PR is, in particular given the fact that it is quite massive.

Also, it might be worth to consider to stop running the code generator during the build process and simply ship a copy of the results:

- In some cases the fact that the `generated_source` directory is not available from the start leads to GCC warnings
- I consider having source code in the build directory to be very bad style because it adds quite a bit of complexity to the build system. In particular,  this makes it much harder to experiment with alternative build systems.
- The code generator might not work properly during the build. in particular, my build broke on several occasions because of an illegal locale, incorrect libstdc++ versions or due to the boost libraries not being in the LD_LIBRARY_PATH
- The code generator could be more easily switched to fancier methods like python+jinja2 without causing problems for normal users
- the build of opm-parser would become faster and less complex because no separate "core parser" library would be needed just for the code generator anymore

it looks like this is not included in this PR, though?

jokva:
> while I don't exactly consider this module to be my glass of beer, I'm still wondering what the motivation for this PR is, in particular given the fact that it is quite massive.

Faster, easier-to-understand, more straight forward & modern target-oriented. It would've been nicer had we been able to use a 3.x series cmake, but we're still a few years from that.

> Also, it might be worth to consider to stop running the code generator during the build process and simply ship a copy of the results:

No, because that makes it more work and higher threshold of tuning the generated output (which still has plenty room for improvement).

> In some cases the fact that the generated_source directory is not available from the start leads to GCC warnings

What warnings would that even be? gcc doesn't care.

> I consider having source code in the build directory to be very bad style because it adds quite a bit of complexity to the build system. In particular, this makes it much harder to experiment with alternative build systems.

You're obviously free to draw that conclusion, but the complexity of generated code doesn't go away by generating it ahead-of-time, and other build systems will still need to account for it.

> The code generator might not work properly during the build. in particular, my build broke on several occasions because of an illegal locale, incorrect libstdc++ versions or due to the boost libraries not being in the LD_LIBRARY_PATH

All of which have nothing to do with the code being generated.

> The code generator could be more easily switched to fancier methods like python+jinja2 without causing problems for normal users

There's no problem changing the generator itself today, it would be a drop-in in the custom command. It's certainly not ideal that it's written in C++, but that's and old and peculiar decision. But there's nothing with this approach nor code generation in general that hinder that.

> the build of opm-parser would become faster and less complex because no separate "core parser" library would be needed just for the code generator anymore

This entire PR is about making the build less complex, but code generators are nothing new.

joakim-hove:
Lo and behold; I actually get a compilation error on this. Using gcc 4.9 and Boost 1.58:

```txt
[  2%] Building CXX object lib/json/CMakeFiles/opmjson.dir/JsonObject.cpp.o
In file included from /private/joaho/work/OPM/opm-parser/lib/json/JsonObject.cpp:29:0:
/private/joaho/work/OPM/opm-parser/lib/json/include/opm/json/JsonObject.hpp:29:15: error: using typedef-name ?boost::filesystem::path? after ?class?
         class path;
               ^
In file included from /private/joaho/work/OPM/opm-parser/lib/json/JsonObject.cpp:26:0:
/usr/include/boost/filesystem/path.hpp:49:52: note: ?boost::filesystem::path? has a previous declaration here
     typedef basic_path< std::string, path_traits > path;                                                    ^

make[2]: *** [lib/json/CMakeFiles/opmjson.dir/JsonObject.cpp.o] Error 1
make[1]: *** [lib/json/CMakeFiles/opmjson.dir/all] Error 2
make: *** [all] Error 2
```

is it me?

jokva:
No, that looks like a bug (on our part). Looks like boost.filesystem.path will be a typedef instead of a class from 1.58, which means forward declaring it like this breaks.

Two solutions: 
1. re-introduce the boost filesystem header
2. remove the public boost filesystem dependency of JsonObject (i.e. kill the boost-path constructor).

andlaus:
let me first reiterate that I don't call the shots in this module and my comment above was in no way meant as a criticism. I just think that using a build system that is significantly different from that of the rest of the project creates a significant additional amount of additional work for you. If you continue to be willing to maintain that wheel and if you are responsive to issues, then I have no problem with the church of NIH either.

With that said, I'd like to answer your reply:

>> Also, it might be worth to consider to stop running the code generator during the build process and simply ship a copy of the results:
>
> No, because that makes it more work and higher threshold of tuning the generated output (which still has plenty room for improvement).

why do you think so? I'm not proposing to get rid of the code generator entirely, just to not run  it automatically by the build system. In this scenario, the updated auto-generated files are just a `make genfiles` (or so) in the source directory away. As far as I can see, the main difference is that maintainers of opm-parser should make sure that this is done when some keyword is changed or added before merging the respective PR.

Actually, I think that switching to a fancier approach to code generation like that of OPM/opm-material#220 *lowers* the threshold of improving the code generator because c++ is not really a good match for this job.

>>  In some cases the fact that the generated_source directory is not available from the start leads to GCC warnings
>
> What warnings would that even be? gcc doesn't care.

mine does:

```
erne@muddipc:~/src/opm-parser|master > make
[  1%] Built target opmjson
[  2%] Built target runjsonTests
[  2%] Building CXX object opm/parser/eclipse/CMakeFiles/createDefaultKeywordList.dir/Deck/Deck.cpp.o
cc1plus: warning: /home/erne/src/opm-parser/generated-source/include: No such file or directory [-Wmissing-include-dirs]
```

>> I consider having source code in the build directory to be very bad style because it adds quite a bit of complexity to the build system. In particular, this makes it much harder to experiment with alternative build systems.
>
> You're obviously free to draw that conclusion, but the complexity of generated code doesn't go away by generating it ahead-of-time, and other build systems will still need to account for it.

what makes you think so? I'll bet a barrel of beer that pre-generated files are just like any other source files when it comes to the build system, i.e., they can reside in the source directory instead of the build directory and no particular weirdness of the code generator needs to be taken care of during the build.

>>   The code generator might not work properly during the build. in particular, my build broke on several occasions because of an illegal locale, incorrect libstdc++ versions or due to the boost libraries not being in the LD_LIBRARY_PATH
>
> All of which have nothing to do with the code being generated.

it does in the sense that no source gets generated at all and my build breaks prematurely ;) (the binaries will still not run, but in this scenario I can change the LD_LIBRARY_PATH without loosing ten minutes for the rebuild.)

>> The code generator could be more easily switched to fancier methods like python+jinja2 without causing problems for normal users
>
> There's no problem changing the generator itself today, it would be a drop-in in the custom command.

the point is that this would force *everyone* who wants to build the module to install somewhat exotic dependencies, not just people who modify keyword definitions or the code generator itself.

jokva:
> mine does:

That's a bug, then, include dirs should be provided.

andlaus:
> That's a bug, then, include dirs should be provided.

doesn't that lead back to the original point that source code in the build directory should be considered very bad style?

jokva:
No, it just means buggy code is buggy.

andlaus:
yeah and solutions tend to be more buggy than simple ones ;)

jokva:
Which is why we're moving towards a simpler, vanilla-cmake solution.

joakim-hove:
Some comments:

1. No system is perfect - and that certainly includes the build system in opm-parser, however this PR is in my opinion a clear improvement.

2. @andlaus seems to argue that keywords should be pre generated and embedded in the source, and not automatically rebuilt. I am willing to bet that e.g. @andlaus himself would complain about the crappy build system on the very first occasion someone forgot to rerun the code generator.¨

3. I agree that using C++ for the code generator is a somewhat weird choice; at the time we thought that dynamically adding keywords to the parser *at runtime* would be an important feature, then it would necessarily need to be implemented with C++ and to abide by the eat-your-own principle C++ was used for all code generation.

joakim-hove:
> Two solutions: [...]

Your call.

andlaus:
> I am willing to bet that e.g. @andlaus himself would complain about the crappy build system on the very first occasion someone forgot to rerun the code generator.

I won't. Also, this can easily be caught by a CI system by just running the code generator "manually" and making it fail if anything changed. as far as I can see, that would be basically a two-liner, right @akva2?

also, let me repeat: if *you* fix the quirks quickly enough *I* do not really care. (but I care about features like `dunecontrol` support and support for newer Dune versions and I will definitely complain if they break in the future.)

lib/eclipse/Deck/DeckKeyword.cpp
```diff
@@ -17,9 +17,9 @@
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "DeckKeyword.hpp"
```
andlaus: Since this results in almost all source files of the module being touched, don't you think the directory reorganization and such stylistic non-buildsystem changes merit a separate PR? (or is this due to the PR's WIP status?)

jokva: Easy enough to split out when things are ready, and easier for me to work with one PR. Making it public has mostly to do with travis.

jokva: I take it back - the moving should coincide with the makefile change because it's a reorganisation as well. It's touching the files, sure, but only by `git mv`, so it shouldn't be that big of a deal.

jokva:
jenkins build this with downstreams opm-common=248 opm-output=188 please

jokva:
jenkins build this with downstreams opm-common=248 opm-output=188 please

jokva:
jenkins build this with downstreams opm-common=248 opm-output=188 please

jokva:
jenkins build this with downstreams opm-common=248 opm-output=188 please

jokva:
jenkins build this with downstreams opm-common=248 opm-output=188 please

jokva:
jenkins build this with downstreams opm-common=248 opm-output=188 please

jokva:
Alright, so this goes through on travis now. I'll do some cleanup and further testing Monday, and we'll move onto possible merging.

jokva:
jenkins build this with downstreams opm-common=248 please

jokva:
jenkins build this with downstreams opm-common=248 please

jokva:
jenkins build this with downstreams opm-common=248 please

jokva:
jenkins build this with downstreams opm-common=248 please

jokva:
Looks like jenkins has aggressively cached some old garbage - can the cache be evicted?

akva2:
there is no cache on jenkins.

jokva:
Then why does it immediately start the parser build without rebuilding ecl?

akva2:
? it builds opm-common, then libecl, then opm-parser. as it should.

jokva:
Yeah, except when I try to restart it it gets to parser (on the new job) in about five seconds, and it *should* take longer, which implies that it isn't a fresh build.

akva2:
that's ccache doing its job.

jokva:
And if I want to evict that, how do I proceed?

akva2:
jenkins uses installed builds. sibling builds work, but the include paths get messed up if libecl was installed.

akva2:
okay, it's the static build that causes issues. 
with a shared build:
```cmake
set_target_properties(ecl PROPERTIES
  INTERFACE_COMPILE_OPTIONS "-pthread"
  INTERFACE_INCLUDE_DIRECTORIES "/home/akva/kode/opm/opm-parser/deps/libecl/lib/include;/home/akva/kode/opm/opm-parser/serial/build-libecl/lib/include"
  INTERFACE_LINK_LIBRARIES "m;dl;-pthread;/usr/lib/libblas.so;/usr/lib/liblapack.so;/usr/lib/libblas.so;/usr/lib/x86_64-linux-gnu/libz.so"
)
```
with a static build:
```cmake
set_target_properties(ecl PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "/var/lib/jenkins/workspace/opm-data-PR-builder/deps/libecl/libecl/include"
  INTERFACE_LINK_LIBRARIES "ert_geometry;ert_util"
)
```

the link libraries have their own properties which set subpaths, but the problem is that we are missing  the include dir from the build directory.

jokva:
```
set_target_properties(ecl PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "/var/lib/jenkins/workspace/opm-data-PR-builder/deps/libecl/libecl/include"
  INTERFACE_LINK_LIBRARIES "ert_geometry;ert_util"
)
```

These libraries don't exist anymore. Is it not possible to create a proper new, clean build? Reset some state?

akva2:
the jenkins job starts with a git clean -xffd - that wipes absolutely everything.

jokva:
Then why is ert_util still there?

jokva:
Besides, the missing header jenkins reports is not from the build directory, it's from the source directory.

akva2:
if you are gonna be snarky, you should at least make sure you interpret information correctly.
the missing directory (or well, equivalent) is
```
/home/akva/kode/opm/opm-parser/serial/build-libecl/lib/include
```
that is the build directory.

jokva:
Who's being snarky?

Jenkins reports:

> /var/lib/jenkins/workspace/opm-parser-PR-builder/lib/eclipse/include/opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.hpp:27:30: fatal error: ert/ecl/Smspec.hpp: No such file or directory
>  #include <ert/ecl/Smspec.hpp>

libecl reports that it's installed, and it's not generated during build, it's copied from libecl's source directory. What am I missing?

akva2:
finally found the damn cause. now to figure out how it could possibly happen.
for reasons i do not yet fathom, we get
```
ecl_DIR:PATH=/var/lib/jenkins/workspace/opm-data-PR-builder/mpi/build-libecl
```

cmake has picked it from a completely different builder directory. i have NO IDEA how that could possibly happen, it shouldn't be seeing that shit at all.

jokva:
Anything you think I can do to fix it?

akva2:
got it. it's the cmake registry that causes issues. export() registers it in a windows-type registry and whichever happened to be put there first "wins" it seems. icky concept. i need to use CMAKE_FIND_PACKAGE_NO_SYSTEM_PACKAGE from toolchain to fix it. easy enough here but this might bite anyone using installed builds at random :/

you can now kick jenkins again.

jokva:
Isn't that registry supposed to be invalidated by nearly anything? At least that's what the documentation indicates. I'll restart the job now.

jokva:
jenkins build this with downstreams opm-common=248 please

akva2:
apparently the documentation is the sunshine version cause it did not overwrite with the new one in this case.

akva2:
wtf.

akva2:
doh. i confused myself. the toolchain option isn't working, i had manually added to flag to the find_package call in my test. investigating..

akva2:
right. the option was added in 3.1 :/ for 3.0.2 there is no choice but to use the find_package flag. i will obtain a new cmake .. somehow.

jokva:
Am I to understand `install`'d packages persist across builds? And across parser/data/output etc?

akva2:
yes, that's the function served by export() in CMakeLists.txt - it registers the package in a registry and find_package just magically grabs it from there. but this cause an issue on the build server, where we build and install many many copies of the same library (to different paths).

cmake 3.6.2 is now in place (thank you jessie-backports). i have tested again by hand and it worked. and there shall be no edited files now so it should be good(tm).

jokva:
You mean export(TARGETS)? Because that's meant to register the build directory (according to docs anyway), but that's surely not the one being picked up (which is the opm-data-PR install). Oh well, let's try a rebuild with the new cmake.

jokva:
jenkins build this with downstreams opm-common=248 please

jokva:
jenkins build this with downstreams opm-common=248 please

jokva:
jenkins build this with downstreams opm-common=248 please

blattms:
I recommend against using the cache registry with export(PACKAGE). It
breaks things as it always uses the last configured build/install
directory of the module. For development packages like DUNE/OPM this
is rather annoying. I had a lot of trouble with this when doing
the CMake implementation for DUNE and finally ripped it out.

jokva:
jenkins build this with downstreams opm-common=248 please

jokva:
jenkins build this with downstreams opm-common=248 please

jokva:
jenkins build this with downstreams opm-common=248 please

jokva:
jenkins build this with downstreams opm-common=248 please

jokva:
jenkins build this with downstreams opm-common=248 please

jokva:
jenkins build this with downstreams opm-common=248 please

jokva:
jenkins build this with downstreams opm-common=248 please

jokva:
jenkins build this with downstreams opm-common=248 please

jokva:
And finally we're there.

jokva:
I had to do some hacking on opm-common because ewoms assumes `HAVE_ERT`, but nothing sets it anymore, and ewoms doesn't spell out its dependency. I leave sorting out the details there to someone who's more comfortable with the opm build system than me.

andlaus:
If you prefer, that #if can test for HAVE_OPM_PARSER. HAVE_ERT is slightly more concise IMO, but this is just the eWoms legacy ECL output code which is scheduled for replacement and it is not really useful in contexts where opm-parser is not present anyway.

jokva:
That's certainly less hacky, but for as long as there's an actual dependency I'd prefer that relationship is codified (i.e. make `ecl` an (optional) dependency and check for that).

joakim-hove:
OK - this should be merged with opm/opm-common#248.

PR #1066, opened 2017-05-25
Summary rpr

joakim-hove:
The SUMMARY section keywords like `RPR` is used to ask for per region results. If the `Rxxx`keywords are given without region numbers the results should be reported *for all regions*:

```txt
RPR
/
```
currently the opm code interprets *all regions* as all FIPNUM values which are present in the grid, that could for instance result in the following set of summary keys: `{RPR:4, RPR:3, RPR:7}`. The Eclipse interpretation of *all regions* is the set `{"RPR:1, RPR:2, RPR:3, .... RPR:NTFIP}` - where `NTFIP`is the maximum number of fluid in place regions from either the `TABDIMS` or `REGDIMS` keyword. Observe that set of summary keys generated by Eclipse does not consider which regions are actually present in the deck, so we can get many summary variables which only get value zero.

I encourage @alfbr to merge this - if he agrees with the slight change of policy induced in this PR.

### DIFF
--- a/opm/parser/eclipse/EclipseState/EclipseConfig.cpp
+++ b/opm/parser/eclipse/EclipseState/EclipseConfig.cpp
@@ -22,6 +22,7 @@
 #include <opm/parser/eclipse/Deck/Deck.hpp>
 #include <opm/parser/eclipse/Parser/ParseContext.hpp>
 #include <opm/parser/eclipse/Deck/Section.hpp>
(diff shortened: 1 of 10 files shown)


joakim-hove:
jenkins build this with downstreams opm-output=187 please

joakim-hove:

joakim-hove:
Does anyone care? I'll just merge this tomorrow.

PR #1073, opened 2017-06-06
Set INTERFACE_INCLUDE_DIRECTORIES on boost targets

jokva:


jokva:
This doesn't seem to warn more on my cmake 2.8.12 system (which may be a false negative), but I suppose it's good enough.

PR #1075, opened 2017-06-06
Set imported boost locations without build type

jokva:


dr-robertk:
This seems to fix the BOOST issue.

jokva:
I suppose the problem was that when using a non-DEBUG or RELEASE config it wouldn't file the libraries, or when some library didn't have a `_${CONFIG}` version. Falling back to a defaulted one is reasonable in these cases anyway, and I suppose cmake's internals won't look through the variations on `IMPORTED_LOCATION_`.

PR #1080, opened 2017-06-09
Fix errors / warnings from static analyzers 

akva2:
These are the errors report from static analyzers in this module.

As the string_view is designed around implicit casts, it does not make sense to make those constructors explicit so I chose the suppression route there. 

Just as a FYI: should you not like the suppression tags, we can use a regular expression in the wrapper script instead.

### DIFF
--- a/lib/eclipse/Deck/DeckKeyword.cpp
+++ b/lib/eclipse/Deck/DeckKeyword.cpp
@@ -23,20 +23,16 @@
 
 namespace Opm {
 
-    DeckKeyword::DeckKeyword(const std::string& keywordName) {
-        m_knownKeyword = true;
-        m_keywordName = keywordName;
-        m_isDataKeyword = false;
-        m_fileName = "";
-        m_lineNumber = -1;
+    DeckKeyword::DeckKeyword(const std::string& keywordName) :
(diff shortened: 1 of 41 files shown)


lib/eclipse/Deck/DeckKeyword.cpp
```diff
@@ -23,19 +23,19 @@
...
         m_isDataKeyword = false;
-        m_fileName = "";
         m_lineNumber = -1;
     }
 
-    DeckKeyword::DeckKeyword(const std::string& keywordName, bool knownKeyword) {
+    DeckKeyword::DeckKeyword(const std::string& keywordName, bool knownKeyword) :
+        m_keywordName(keywordName)
+    {
         m_knownKeyword = knownKeyword;
```
joakim-hove: I guess the `knownKeyword` could also be initialized in the initializer list?

akva2: Yes though it doesnt matter for primitives (no default constructors to call). If you prefer to use init list you shall have init list.

akva2:
there you go. all warnings quell + init list change requested included.

PR #1081, opened 2017-06-09
Added explicit list of keywordfiles.

joakim-hove:
**Keyword generation:** Initially the keywords were generated by recursively scanning a directory, this scanning process was done for every build - and *if* changes were detected updated versions of generated source and header files were written to disk. I.e. this was a sort of manual dependency tracking.

One of the changes in #1065 was that the keyword directory was scanned during configure time, and normal cmake dependencies were established on the various keyword files. By using the standard cmake approach to dependencies the build time was  reduced and also the build system complexity was reduced. 

After #1065 we had dependencies on all the individual keywords, but not on the set itself - i.e. if a keyword was added a rebuild of the keywords was not triggered. In particular this implied that after a `git pull` which added new keywords you would have to manually invoke `cmake`. 

With this patch the set of keywords is handled *very explicitly* -  they are listed in a file - and there is no scanning involved. This way it will be the sole responsability of the developer adding or removing keywords to make sure things are consistent, for everyone else it should Just Work&trade;


joakim-hove:

joakim-hove:
@jokva: Do you understand the Appveyor failure?

.travis.yml
```diff
@@ -22,6 +22,11 @@ addons:
       - cmake
       - cmake-data
 
+env:
+  global:
+  - TRAVIS_CI=1
```
jokva: Why is this necessary? Travis sets plenty environment anyway, why can't that be used?

lib/eclipse/CMakeLists.txt
```diff
@@ -24,21 +24,18 @@ add_executable(genkw Parser/createDefaultKeywordList.cpp
...
-    DEPENDS genkw ${keyword_templates}
-)
-
-add_custom_command(
-    OUTPUT  ${CMAKE_CURRENT_BINARY_DIR}/inlinekw.cpp
-    COMMAND genkw ${PROJECT_SOURCE_DIR}/share/keywords
                   ${CMAKE_CURRENT_BINARY_DIR}/inlinekw.cpp
-    DEPENDS genkw ${keyword_templates}
+                  ${keyword_files}
+    DEPENDS genkw ${keyword_files} share/keywords/keyword_list.cmake
```
jokva: Why does it depend on the cmake file? Cmake will notice that it is touched anyway, and force a reconfigure, if it's modified in any way, even without this.

jokva:
@joakim-hove My guess would be that it's handed the  list of keyword names as a long, semi-colon separated string, and that makes the code generator crash.

jokva:
Which might actually be wrong as well, because cmake might change the list from a string to a space-separated set of arguments when invoking the `add_custom_command`.

jokva:
I just checked - on linux, is definitely the latter. Which leads me to think that cmd.exe might have a hard limit on the length of the argument string

jokva:
https://stackoverflow.com/questions/3205027/maximum-length-of-command-line-string

jokva:
An immediate, simple solution is to mark the keyword list file as a dependency, pass it as an argument to genkw and spin through the lines in the file, rather than argv.

joakim-hove:
> I just checked - on linux, is definitely the latter. Which leads me to think that cmd.exe might have a hard limit on the length of the argument string

Yes - that was clearly it; thank you,

> An immediate, simple solution is to mark the keyword list file as a dependency, pass it as an argument to genkw and spin through the lines in the file, rather than argv.

That is certainly a viable solution.

.travis.yml
joakim-hove: The variable has already been used: https://github.com/OPM/opm-output/pull/194 - but yes you are right; a better solution would have been to just use one the variables already set by Travis.

lib/eclipse/CMakeLists.txt
joakim-hove: I tried this and concluded that the `share/keywords/keyword_list.cmake` file must be listed as en explicit dependency: My understanding is that changing the `share/keywords/keyword_list.cmake` fill force a reconfigure, but the `genkw` target will still be up to date, i.e. no new set of keywords will be generated - happy to be proven wrong though.

jokva: It's a dependency to the full cmake system when you `include()` it, or have you some experiment that suggest otherwise? If so I have some assumptions I must reconsider.

joakim-hove: I agree that when you use `include()` you have a dependency to the *full*  cmake system, i.e. changing the keyword file will give a full reconfigure. However that reconfigure will not rerun the custom command which creates keyword files.

I have tested this - but might have stumbled in the testing.

jokva: That might be, sure, or maybe it only will if the content of the *keyword list* changes.

andlaus:
> With this patch the set of keywords is handled very explicitly - they are listed in a file - and there is no scanning involved. This way it will be the sole responsability of the developer adding or removing keywords to make sure things are consistent, for everyone else it should Just Work™

the approach looks fine to me; I'm just wondering what the practical difference to offline source generation is, i.e., the developer who opens a PR that adds a keyword forgets to run the code generator and the new keyword is not available for users. That's exactly the same story when forgetting to add the new file to `keyword_list.cmake` after this PR?!

jokva:
Yes, if you don't include code, it won't be compiled.

joakim-hove:
>  I'm just wondering what the practical difference to offline source generation is

Maybe not much - I like the online source generation though.

joakim-hove:

joakim-hove:

PR #1084, opened 2017-06-12
Add SPOLY and make SSOL a grid property

andlaus:
this adds a JSON definition plus grid property for the SPOLY keyword and it also adds the already existing SSOL keyword to the list of grid properties. with this PR it should be possible to access all properties required for polymer and solvent enabled simulators without resorting to the deck object.


GitPaean:
A small question here. From this PR, I could not tell how far the keywords are supported. For example the keyword, `SPOLY`

Can we specify the initial polymer concentration with something like 
```
SPOLY
10*0.1
10*0.2 /
```
?
Or currently, we only support the default initial polymer concentration to be all `0.0`? 

Thanks.

GitPaean:
It is supported through somewhere related to `GridProperties`. Sorry for the noise.

PR #1086, opened 2017-06-12
export( PACKAGE ) can be configured.

joakim-hove:
Ability to disable export to package registry.

Comments: @blattms / @jokva ?


blattms:
I would do it the other way around: Always export but only use the package registry during find_package when requested. If not requested use `NO_CMAKE_PACKAGE_REGISTRY` and `NO_CMAKE_SYSTEM_PACKAGE_REGISTRY` with the find_package call. That way there is a complete registry available whenever users want to use it.

joakim-hove:
> I would do it the other way around: Always export but only use the package registry during find_package when requested. If not requested use NO_CMAKE_PACKAGE_REGISTRY and NO_CMAKE_SYSTEM_PACKAGE_REGISTRY with the find_package call. That way there is a complete registry available whenever users want to use it.

OK - that makes sense to me; but then the changes must be in the `find_package( )` calls in downstream modules right? I.e. the changes must be in opm-common in our case - or??

jokva:
Unnecessary and already supported via NO_CMAKE_PACKAGE_REGISTRY

blattms:
Sorry, but I do not understand what you mean with your comment,
@jokva.

jokva:
I don't think there should be a "please don't export" option at all. CMake has built-in support for ignoring the cache and for specifing entries, and the problem a non-export feature would solve is really not that desirable to support anyway (multiple incompatible build-directories of some module lying about).

atgeirr:
> Having 14 different, incompatible copies of the opm-parser build directory is not the norm, should not maybe be the norm and is not something that should be optimised for.

But 2 is the common case I believe (debug and release), I think that should not be made hard. In any case @blattms' suggestion to change the find macros to by default ignore the registry seems to me like it should satisfy everyone, right?

joakim-hove:
OK - but then my take on the problem is not really that opm-parser is exporting a package entry, but rather that the downstream modules by default use those entries?

If we agree that is the problem - then this should be fixed/improved downstream - e.g. by setting the `NO_CMAKE_PACKAGE_REGISTRY` in the default opm find functionality?

blattms:
I am getting really confused by these discussions. I run into this
problem because at some point I decided to not use dunecontrol on my
compute server as people from the company that @jokva belongs to had
problems with it and reproducing problems is better if I build in the
same way as they do. Sadly, there is no easy way to deactivate the
cache without patching OPM that I know of. Either I need to have such
a switch or I will be forced to use dunecontrol which will specify the
<opm-module>_DIR to tell OPM the correct version of modules to use.

I need multiple builds lying around (with, and without MPI and a
debugging version). There is no way around this without wasting even
more time building OPM.

joakim-hove:
> I am getting really confused by these discussions.

Well - that applies to me to. I will try supply a fix in opm-common.

joakim-hove:
OK: please continue the discussion here: https://github.com/OPM/opm-common/pull/256

jokva:
> Sadly, there is no easy way to deactivate the cache without patching OPM that I know of.

Sure there is: https://cmake.org/cmake/help/v3.1/variable/CMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY.html https://cmake.org/cmake/help/v3.1/manual/cmake-packages.7.html#disabling-the-package-registry

jokva:
> I need multiple builds lying around (with, and without MPI and a debugging version). There is no way around this without wasting even more time building OPM.

You'll have to look for modules somehow when configuring fresh. Once it's configured its path won't change (until you delete the build dir it had registered)

blattms:
Sorry, that option is not available in version 3.0.2

jokva:
https://cmake.org/cmake/help/v2.8.12/ctest.html#command:find_package

Quite right, it's an option to `find_package`

joakim-hove:
Please - head over to: OPM/opm-common#256

PR #1090, opened 2017-06-14
Create boost targets names dynamically

jokva:
On some systems, in particular Windows or custom boosts, plenty
information about configuration is embedded into the library name.
However, opm-parser's boost names were hard coded to the vanilla library
names, which leads to libraries not being found.

Generate the target names dynamically (and address them as variables) so
that the target names match the name of the library actually being
linked to - a name which is also then later exported to the
opm-parser-config.cmake file.

Supersedes https://github.com/OPM/opm-parser/pull/1088

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,32 +76,80 @@ else ()
 endif ()
 
 find_package(Boost 1.44.0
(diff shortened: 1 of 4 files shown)


joakim-hove:

PR #1093, opened 2017-06-20
Add plmixnum

andlaus:
the first patch changes the default value for the OPERNUM grid property from 0 to 1 and removes a `//*` token (which should be considered to be a *really* bad way to distinguish between C and C++).

the second patch adds the PLMIXNUM keyword and the corresponding grid property. this property would be good to have for the new implementation of the polymer blackoil extension: https://github.com/OPM/ewoms/pull/195#discussion_r122908138

### DIFF
--- a/lib/eclipse/EclipseState/Eclipse3DProperties.cpp
+++ b/lib/eclipse/EclipseState/Eclipse3DProperties.cpp
@@ -207,7 +207,8 @@ namespace Opm {
             GridProperties< int >::SupportedKeywordInfo( "OPERNUM", 1, "1" ),
             GridProperties< int >::SupportedKeywordInfo( "PVTNUM" , 1, "1" ),
             GridProperties< int >::SupportedKeywordInfo( "ROCKNUM", 1, "1" ),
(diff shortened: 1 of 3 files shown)


andlaus:
okay, on closer inspection the second patch was bogous as it seems like there is no PLMIXNUM keyword. on the other hand, the default value for OPERNUM should probably still be changed and the `//*` should be removed, IMO.

joakim-hove:
This can obvously be merged as is (please do so @andlaus) - I am holding back because it was not entirely clear to me whether there should be a `PLMIXNUM` or not?

andlaus:
> I am holding back because it was not entirely clear to me whether there should be a PLMIXNUM or not?

I was just confused: It looks like this keyword is a recent addition and @totto82 told me to add this, but it was not yet present in the documentation I looked at yesterday. I re-added the patch, so a slightly closer look is warranted before you merge...

OPMUSER:
There is Eclipse 100


/*E*/

	

/*I*/

	

/*P*/

	

/*C*/

	

Managing Director

T:  +(65)-9173-7031

F:  +(65)-6732-2382

E: david.baxendale@eipc.co

PR #1094, opened 2017-06-20
Support radial properties

bska:
This change-set adds support for internalising the `PERMR` and `PERMTHT` keywords.  The former was already known to the parser, but was not recognised as a 3D grid property.  The latter is new as of this change-set.

We also relocate the `RADIAL` keyword to the `RUNSPEC` section as per the official specification.


bska:

PR #1095, opened 2017-06-21
3D Properties Test: Check Radial Tensor Support

bska:
Mostly for completeness since the underlying implementation just piggy-backs on the existing support for Cartesian tensors.

This should really have been part of #1094.  My apologies for not thinking of this sooner.

### DIFF
--- a/lib/eclipse/tests/Eclipse3DPropertiesTests.cpp
+++ b/lib/eclipse/tests/Eclipse3DPropertiesTests.cpp
@@ -167,6 +167,90 @@ static Opm::Deck createValidPERMXDeck() {
     return parser.parseString(deckData, Opm::ParseContext() );
 }
 
+static Opm::Deck createQuarterCircleDeck() {
+    const auto* input = R"(
+RUNSPEC
+
+TITLE
+  'Quarter Circle 100x21x20' /
+
+DIMENS
+  100 21 20 /
+
+METRIC
+
+RADIAL
+
+OIL
+WATER
+
+TABDIMS
+/
+
+START
+  19 JUN 2017
+/
+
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

joakim-hove:
Yes - the reason I never added the `GridProperties` `PERMR` and `PERMTHT` was that I was a bit uncertain about the expected alias behaviour - and then I forgot the whole thing. By reading the ECLIPSE documentation you could get the impression `PERMX <-> PERMR` and `PERMY <-> PERMTHT` - whereas our implementation implies that they are separate objects which can be manipulated independently. 

To actully implement "property reference semantics" is probably not worth it?

bska:
> To actually implement "property reference semantics" is probably not worth it?

Agreed.  I thought it was more of the case that one (i.e., a simulation case) would use `PERM{R,THT}` when the `RUNSPEC` has `RADIAL` and `PERM{X,Y}` otherwise.

I guess it ultimately comes down to whether or not the simulator, or at least the transmissibility calculation code, wants to be agnostic to the specific nature of the grid.  While that could be a goal from a software engineering point of view, the formulae for computing transmissibilities really are different in the two coordinate systems and trying to hide that behind permeability aliases would probably  just obscure the implementation.

joakim-hove:
> Agreed. I thought it was more of the case that one (i.e., a simulation case) would use PERM{R,THT} when the RUNSPEC has RADIAL and PERM{X,Y} otherwise.

That would make sense - hopefully you are right; I thought it was more like the keywords `PERMR` and `PERMX`could be interchanged freely. Anyway - we'll burn that if we ever get there.

PR #1098, opened 2017-06-26
Tempi init

joakim-hove:
Initialize TEMPI property from the `RTEMP` keyword.

See: https://github.com/OPM/opm-simulators/issues/1231

### DIFF
--- a/lib/eclipse/share/keywords/000_Eclipse100/R/RTEMP
+++ b/lib/eclipse/share/keywords/000_Eclipse100/R/RTEMP
@@ -1,0 +1,4 @@
+{"name" : "RTEMP" , "sections" : ["PROPS","SOLUTION"],
+ "items" :
+ [{"name" : "TEMP" , "value_type" : "DOUBLE" , "default" : 15.555 , "dimension" : "Temperature"}]}
+ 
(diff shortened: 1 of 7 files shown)


andlaus:
I think we clashed on this one, so #1101 also fixes this. I'm equally fine with merging any of these PRs.

lib/eclipse/share/keywords/000_Eclipse100/R/RTEMP
```diff
@@ -0,0 +1,4 @@
+{"name" : "RTEMP" , "sections" : ["PROPS","SOLUTION"],
+ "items" :
+ [{"name" : "TEMP" , "value_type" : "DOUBLE" , "default" : 15.555 , "dimension" : "Temperature"}]}
```
andlaus: shouldn't the default be 33.333 °C (-> 60 °F)?

andlaus: actually, the 15.5 °C are right. my confusion was caused by a possible bug of the 'units' command line utility: it produces 33.33 degC if asked for 60 degF...

PR #1100, opened 2017-06-26
Moved integration tests.

joakim-hove:
When building in source there was a name clash on the "symbol" `IntegrationTests`.


joakim-hove:
@andlaus will you test this.

andlaus:
I just did so: With this everything works fine on my machine, thanks for the patch.

PR #1101, opened 2017-06-26
implement proper support for constant initial reservoir temperature

andlaus:
This involves adding the RTEMP and RTEMPA keywords and using them in the grid property initializer of TEMPI if no temperature-vs-depth table is specified.

on my machine, this fixes OPM/opm-simulators#1231


joakim-hove:
Will merge a mix of this #1098

joakim-hove:
I merged #1098 - mainly because that PR made less use of the `Deck`class.

andlaus:
good. I agree that #1098 was the better choice.

PR #1102, opened 2017-06-27
time_t in time map

jijordre:


atgeirr:
I will test that this resolves the "long timestep" issue for me, thanks!

lib/eclipse/EclipseState/Schedule/TimeMap.cpp
```diff
@@ -30,29 +30,34 @@
 
 
 namespace Opm {
+    TimeMap::TimeMap(std::time_t startTime) {
+        m_timeList.push_back(startTime);
+    }
+
     TimeMap::TimeMap(boost::posix_time::ptime startDate) {
```
joakim-hove: I guess the boost based constructor could just be removed.

lib/eclipse/EclipseState/Schedule/TimeMap.cpp
```diff
@@ -88,36 +93,56 @@ namespace Opm {
...
+
+            if (new_date.tm_year != prev_date.tm_year)
+                m_first_timestep_years.push_back(step);
+
+            m_timeList.push_back(newTime);
+        } else
+            throw std::invalid_argument("Times added must be in strictly increasing order.");
+    }
 
     void TimeMap::addTime(boost::posix_time::ptime newTime) {
```
joakim-hove: Remove the boost based overload.

lib/eclipse/EclipseState/Schedule/TimeMap.cpp
```diff
@@ -88,36 +93,56 @@ namespace Opm {
...
-        return static_cast<double>(deltaT.total_milliseconds())/1000.0;
+        const time_t deltaT = m_timeList.back() - m_timeList.front();
+        return static_cast<double>(deltaT);
     }
 
+    void TimeMap::addTime(std::time_t newTime) {
+        const std::time_t lastTime = m_timeList.back();
+        const size_t step = m_timeList.size();
+        if (newTime > lastTime) {
+            const std::tm new_date = *gmtime(&newTime);
```
joakim-hove: The `gmtime` function is purely posix - this:
```
util_set_date_values_utc( t1, &day , &month, &year);
```
is a portable alternative.

lib/eclipse/EclipseState/Schedule/TimeMap.cpp
joakim-hove: Remove boost overload.

lib/eclipse/EclipseState/Schedule/TimeMap.cpp
```diff
@@ -88,36 +93,56 @@ namespace Opm {
...
             if (new_date.year() != prev_date.year())
                 m_first_timestep_years.push_back( step );
 
-            m_timeList.push_back( newTime );
+            m_timeList.push_back(to_time_t(newTime));
         } else
             throw std::invalid_argument("Times added must be in strictly increasing order.");
     }
 
+    void TimeMap::addTStep(std::time_t step) {
```
joakim-hove: I would prefer if `long` was used for a duration in seconds - and not `std::time_t` which will be interpreted (at least in developers heads) with an absolute reference.

lib/eclipse/EclipseState/Schedule/TimeMap.cpp
```diff
@@ -322,20 +386,20 @@ namespace Opm {
         return 0;
     }
 
+    std::time_t TimeMap::to_time_t(boost::posix_time::ptime t) {
```
joakim-hove: Remove

lib/eclipse/EclipseState/Schedule/TimeMap.cpp
```diff
@@ -151,6 +176,31 @@ namespace Opm {
         return monthNames;
     }
 
+    const std::map<std::string , int>& TimeMap::eclipseMonthIndices() {
+        static std::map<std::string , int> monthIndices;
+
+        if (monthIndices.size() == 0) {
+            monthIndices.insert( std::make_pair( "JAN" , 0 ));
```
joakim-hove: I would just shift these number with 1 - then they immediately fit in `TimeMap::mkdate( )`

lib/eclipse/EclipseState/Schedule/TimeMap.cpp
jokva: If you're looking for 64 bit support, `long` is not sufficient. It's only guaranteed to be at least 32 bit.

lib/eclipse/EclipseState/Schedule/TimeMap.cpp
```diff
@@ -189,57 +239,71 @@ namespace Opm {
         }
     }
 
-    void TimeMap::addFromDATESKeyword( const DeckKeyword& DATESKeyword ) {
+    std::time_t TimeMap::timeTFromEclipse(const DeckRecord &dateRecord) {
+        static const std::string errorMsg("The datarecord must consist of the four values "
```
joakim-hove: This will be validated by the `getItem( )` callls anyway. As I see it this function can be significantly simplifed:

```C++
const auto &dayItem = dateRecord.getItem(0);
const auto &monthItem = dateRecord.getItem(1);
const auto &yearItem = dateRecord.getItem(2);
const auto &timeItem = dateRecord.getItem(3);

std::time_t date= mkdate(yearItem.get<int>(0), 
                                          eclipseMonthIndices().at(monthItem.get<std::string>(0)) + 1, 
                                          dayItem.get<int>(0));
return date;
```

lib/eclipse/EclipseState/Schedule/TimeMap.cpp
joakim-hove: OK - `int64_t` then.

jijordre: You mean that the step, as opposed to epoch value, rather be described as `long` or `long long` instead of as `time_t`?

joakim-hove:
OK; overall this looks good. Please change:

1. Remove all boost references.
2. Add tests with a input deck (see: https://github.com/OPM/opm-parser/blob/master/lib/eclipse/tests/TimeMapTest.cpp#L282) which has very long TSTEP, i.e. something like:

```
START
   1 'JAN' 1983 /
...
...

TSTEP
     25550  /   -- Corresponds to 70 years
```

And verify that the end of the simulation is sometime beyond 2040. Also create a test deck with `DATES` keywords extending beyond 2038:

 ```
START
   1 'JAN' 1983 /
...
...

DATES
     1 'JAN' 2040 /


DATES
  1 'JAN'  2050/
```

File: lib/eclipse/EclipseState/Schedule/TimeMap.cpp (hunk 4 above)
Comment (joakim-hove) [inline lib/eclipse/EclipseState/Schedule/TimeMap.cpp]: Yes - `long long` or - as you say `difftime( )` compatible.

**Comment by atgeirr** [comment]
I can confirm that this branch compiles and that Flow then accepts timesteps of a million days.

File: lib/eclipse/EclipseState/Schedule/TimeMap.cpp (hunk 7 above)
Comment (jijordre) [inline lib/eclipse/EclipseState/Schedule/TimeMap.cpp]: Do you here suggest that the time component from the data record, here `timeItem`, be discarded in the calculation of `date`?

Comment (joakim-hove) [inline lib/eclipse/EclipseState/Schedule/TimeMap.cpp]: Ahh sorry - will get back to you.

Comment (joakim-hove) [inline lib/eclipse/EclipseState/Schedule/TimeMap.cpp]: Implement a `TimeMap::mkdate_time()` based on `util_set_datetime_values_utc()` similar to the `TimeMap::mkdate()`

File: lib/eclipse/EclipseState/Schedule/TimeMap.cpp (hunk 1 above)
Comment (jijordre) [inline lib/eclipse/EclipseState/Schedule/TimeMap.cpp]: Done

File: lib/eclipse/EclipseState/Schedule/TimeMap.cpp (hunk 2 above)
Comment (jijordre) [inline lib/eclipse/EclipseState/Schedule/TimeMap.cpp]: Done

File: lib/eclipse/EclipseState/Schedule/TimeMap.cpp (hunk 3 above)
Comment (jijordre) [inline lib/eclipse/EclipseState/Schedule/TimeMap.cpp]: Done

File: lib/eclipse/EclipseState/Schedule/TimeMap.cpp (hunk 2 above)
Comment (jijordre) [inline lib/eclipse/EclipseState/Schedule/TimeMap.cpp]: Done

File: lib/eclipse/EclipseState/Schedule/TimeMap.cpp (hunk 4 above)
Comment (jijordre) [inline lib/eclipse/EclipseState/Schedule/TimeMap.cpp]: Done

File: lib/eclipse/EclipseState/Schedule/TimeMap.cpp (hunk 5 above)
Comment (jijordre) [inline lib/eclipse/EclipseState/Schedule/TimeMap.cpp]: Done

File: lib/eclipse/EclipseState/Schedule/TimeMap.cpp (hunk 6 above)
Comment (jijordre) [inline lib/eclipse/EclipseState/Schedule/TimeMap.cpp]: Done

File: lib/eclipse/EclipseState/Schedule/TimeMap.cpp (hunk 7 above)
Comment (jijordre) [inline lib/eclipse/EclipseState/Schedule/TimeMap.cpp]: I added static `TimeMap::mkdatetime() and simplified the logics down to the following:
```
        const auto &dayItem = dateRecord.getItem(0);
        const auto &monthItem = dateRecord.getItem(1);
        const auto &yearItem = dateRecord.getItem(2);
        const auto &timeItem = dateRecord.getItem(3);

        std::tm tm;
        strptime(timeItem.get<std::string>(0).c_str(), "%H:%M:%S", &tm);

        std::time_t date = mkdatetime(yearItem.get<int>(0),
                                      eclipseMonthIndices().at(monthItem.get<std::string>(0)),
                                      dayItem.get<int>(0),
                                      tm.tm_hour,
                                      tm.tm_min,
                                      tm.tm_sec);
        return date;
```

**Comment by jijordre** [comment]
Two new tests added, one reflecting long TSTEP of 25550 days and one with distant DATES at Jan 1st 2040 and 2050.

File: lib/eclipse/EclipseState/Schedule/TimeMap.cpp [hunk 8]
```diff
@@ -88,36 +85,34 @@ namespace Opm {
...
+            m_timeList.push_back(newTime);
         } else
             throw std::invalid_argument("Times added must be in strictly increasing order.");
     }
 
-
-    void TimeMap::addTStep(boost::posix_time::time_duration step) {
-        boost::posix_time::ptime newTime = m_timeList.back() + step;
+    void TimeMap::addTStep(int64_t step) {
+        const std::time_t newTime = m_timeList.back() + step;
```
joakim-hove: Instead of explicitly using `+` I would recommend the `TimeMap::forward( )` method:
```
addTime( forward( m_timeList.back() , step ));
```
that way we are - at least slightly - better prepared to introduce a non-scalar time type.

lib/eclipse/EclipseState/Schedule/TimeMap.cpp
```diff
@@ -126,125 +121,94 @@ namespace Opm {
...
+        const auto &monthItem = dateRecord.getItem(1);
+        const auto &yearItem = dateRecord.getItem(2);
+        const auto &timeItem = dateRecord.getItem(3);
 
-        try {
-            int day = dayItem.get< int >(0);
-            const std::string& month = monthItem.get< std::string >(0);
-            int year = yearItem.get< int >(0);
-            std::string eclipseTimeString = timeItem.get< std::string >(0);
+        std::tm tm;
```
joakim-hove: Please avoid the use of the Posix only `strptime`:
```
// Check with: timeItem.hasValue(0) first:
int hour,min,second;
if (sscanf(timeItem.get<std::string>(0).c_str(), "%d:%d:%d" , &hour,&min,&second) == 3) {
    // Good
}

joakim-hove:

joakim-hove:

PR #1106, opened 2017-07-05
make opm-parser fully independent of opm-common

andlaus:
i.e. now it does not use anything from opm-common anymore even when it is build using dunecontrol.

this should make it quite a bit easier to use opm-parser stand-alone or as a upstream dependency for dune modules.

### DIFF
--- a/configure
+++ b/configure
@@ -1,35 +1,553 @@
-#!/bin/sh
-# this file is supposed to be located in the source directory
-src_dir=$(dirname $0)
+#!/bin/bash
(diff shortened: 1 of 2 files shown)


andlaus:
travis stubbornly keeps timing out. since none of the affected functionality is used by it and it also compiled a few simulators just before the timeout in one run, I'm quite positive that it should be ignored.

joakim-hove:
Sure - will merge tomorrow.

PR #1107, opened 2017-07-13
make the travis CI scripts independent of opm-common

andlaus:
the reason is that everything in opm-parser is now independent of opm-common, so why should the travis support scripts still rely on it?

besides this, the downstreams are not build and tested by travis anymore because travis recently aquired the habit of timing out if this is enabled, so the downstream builder is more of a distraction than a help. also, the OPM downstream modules nowadays ought to use the OPM Jenkins rig ( http://ci.opm-project.org ) as their primary CI system anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,6 @@ addons:
       - libboost1.55-all-dev
       - gcc-4.8
       - g++-4.8
-      - gfortran-4.8
-      - liblapack-dev
-      - libgmp3-dev
-      - libsuitesparse-dev
-      - libeigen3-dev
       - bc
       - cmake
       - cmake-data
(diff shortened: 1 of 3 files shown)


joakim-hove:
>the reason is that everything in opm-parser is now independent of opm-common, so why should the travis support scripts still rely on it?

Good point - on the other hand we do achieve some DRY goodness the current way; so personally I see no harm in travis depending on opm-common.

>the downstreams are not build and tested by travis anymore because travis recently aquired the habit of timing out if this is enabled, 

Yes that is true; the reason we are still clinging on to the travis builder for opm-parser is that we intend to try out a paid travis account, then we should avoid the timeout. It has taken veeeery long time to get the paid travis account through the enterprise red-tape though.

So - thank you for the PR; I'm still closing it :-(

Issue #1105, opened 2017-07-03
Dunecontrol and OPM modules


MathiasLorenz:
I'm trying to create a dune module which uses OPM functionality from opm-parser. Relevant dune models, opm-common and opm-parser are installed globally (/usr/local/lib) from source. The duneproject script is able to detect the opm-common and opm-parser modules. I'm trying to build my code with the dunecontrol script, but I can see that -Dopm-common_DIR=/usr/local while the correct path would be /usr/local/lib/dunecontrol/opm-common and similarly for opm-parser. I have attached the error messages from the dunecontrol script in [errReadDeckFiles.txt](https://github.com/OPM/opm-parser/files/1119235/errReadDeckFiles.txt).

I have tried to create the project with an opts file, where I specified CMAKE_FLAGS="-Dopm-common_DIR=/usr/local/lib/dunecontrol/opm-common" manually, however, this did not solve the problem.

andlaus:
this looks like a bug in dunecomtrol for the latest dune version because it thinks your module is called `ReadDeckFiles`. (I assume you named it differently.)

even if this was fixed, you milage with dune >= 2.5 may vary considerably due to several issues of the OPM build system, so we recommend that you use Dune 2.4 for now.

alfbr:
Both Dune and OPM supports building directly with cmake. If you are building Dune out-of-tree, you will need to point OPM modules to its location with the CMAKE_PREFIX_PATH variable. There should be no need for dunecontrol. Basically, OPM modules are downstream of Dune modules. If you want to use opm-parser as an upstream module of Dune, I lack the overview to answer.

andlaus:
> Both Dune and OPM supports building directly with cmake.

yes, but the names of the parameters which the OPM build system expects are sometimes different from DUNE's and sometimes they are not. Also, some config.h macros -- e.g. those for SuperLU -- are renamed in newer versions of dune and the build will thus break if these dependencies are to be used. since I suppose that this situation does not make anyone involved with OPM happy, I guess it is time to seriously consider moving the whole thing to dune's build system to avoid the whack-a-mole game for Dune build system changes if a new version ought to be supported in the future.

> There should be no need for dunecontrol.

technically that's correct. there was never any strict need for using dunecontrol. It just makes building the modules in the correct dependency order *much* easier. (I vaguely remember someone complaining about the fact that this is too freaking complicated.)

alfbr:
> It just makes building the modules in the correct dependency order much easier.

[...]

PR #1108, opened 2017-07-13
make the build system compatible with Dune's cmake build system

andlaus:
before this, dune modules that are not using the OPM build system needed to treat opm-parser as an external library, which is IMHO quite sub-optimal cf. #1105 . With this patch, opm-parser can be build as part of duncontrol build chains, i.e., the pkg-config support is fixed, the `opm-parser-config.cmake` provides everything which is expected by dune's cmake based build system and all required files are installed to the right places (hopefully). This means that with this other Dune modules just need to specify a dependency on opm-parser in their dune.module file and opm-parser's libraries and headers should be picked up just fine.

Things like generating opm-parser-config.cmake should quite likely be done in a better way, but I'm not really a cmake expert and the solution proposed by this patch seems to get things done regardless of whether opm-parser is system-installed or not. Comprehensive solutions are highly appreciated. @jokva: it would be nice if you had a look.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -230,7 +230,32 @@ add_subdirectory(lib/json)
 add_subdirectory(lib/eclipse)
 add_subdirectory(applications)
 
-install(FILES dune.module DESTINATION lib/dunecontrol/opm-parser)
-install(EXPORT opm-parser-config DESTINATION share/cmake/opm-parser)
-export(TARGETS opmjson opmparser FILE opm-parserConfig.cmake)
+file(COPY "cmake/modules/OpmParserMacros.cmake" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/cmake/modules")
+
+set(OPM_PARSER_PREFIX "${CMAKE_INSTALL_PREFIX}")
+set(OPM_PARSER_MODULE_PATH "${OPM_PARSER_PREFIX}/share/dune/cmake/modules")
+set(OPM_PARSER_INCLUDEDIRS "${OPM_PARSER_PREFIX}/${CMAKE_INSTALL_INCLUDEDIR}")
+set(OPM_PARSER_LIBDIRS "${OPM_PARSER_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
+set(OPM_PARSER_INSTALLED ON)
+configure_file(opm-parser-config.cmake.in
+  "${PROJECT_BINARY_DIR}/install/opm-parser-config.cmake" @ONLY)
+
+set(OPM_PARSER_PREFIX "${CMAKE_SOURCE_DIR}")
+set(OPM_PARSER_INCLUDEDIRS "${OPM_PARSER_PREFIX}/lib/eclipse/include;${OPM_PARSER_PREFIX}/lib/json/include")
+set(OPM_PARSER_MODULE_PATH "${OPM_PARSER_PREFIX}/cmake/modules")
+set(OPM_PARSER_LIBDIRS "${CMAKE_BINARY_DIR}/lib/eclipse;${CMAKE_BINARY_DIR}/lib/json")
+set(OPM_PARSER_INSTALLED OFF)
+configure_file(opm-parser-config.cmake.in
+  "${PROJECT_BINARY_DIR}/opm-parser-config.cmake" @ONLY)
+
+configure_file(opm-parser.pc.in
+  "${PROJECT_BINARY_DIR}/opm-parser.pc" @ONLY)
+
+install(FILES "cmake/modules/OpmParserMacros.cmake" DESTINATION "share/dune/cmake/modules")
+install(FILES "dune.module" DESTINATION "lib/dunecontrol/opm-parser")
+install(FILES  "config.h.cmake" DESTINATION "share/opm-parser")
+install(FILES  "${CMAKE_CURRENT_BINARY_DIR}/opm-parser.pc" DESTINATION "lib/pkgconfig")
+install(FILES  "${CMAKE_CURRENT_BINARY_DIR}/install/opm-parser-config.cmake" DESTINATION "lib/cmake/opm-parser")
+export(TARGETS opmjson opmparser FILE opm-parser-targets.cmake)
+install(EXPORT opm-parser-targets DESTINATION "lib/cmake/opm-parser")
 export(PACKAGE opm-parser)
--- a/cmake/modules/OpmParserMacros.cmake
+++ b/cmake/modules/OpmParserMacros.cmake
@@ -1,0 +1,16 @@
+# .. cmake_module::
+#
+# This module's content is executed whenever a Dune module requires or
+# suggests opm-parser!
+#
+
+find_package(ecl REQUIRED)
+
+if (ecl_FOUND)
+  set(HAVE_ERT "1")
+  set(HAVE_LIBECL "1")
+endif()
+
+find_package(Boost
+  COMPONENTS regex filesystem system date_time
+  REQUIRED)
--- a/config.h.cmake
+++ b/config.h.cmake
@@ -1,0 +1,50 @@
+/* begin opm-parser
+   put the definitions for config.h specific to
+   your project here. Everything above will be
+   overwritten
+*/
+/* begin private */
+/* Name of package */
+#define PACKAGE "@DUNE_MOD_NAME@"
+
+/* Define to the address where bug reports for this package should be sent. */
+#define PACKAGE_BUGREPORT "@DUNE_MAINTAINER@"
+
+/* Define to the full name of this package. */
+#define PACKAGE_NAME "@DUNE_MOD_NAME@"
+
+/* Define to the full name and version of this package. */
+#define PACKAGE_STRING "@DUNE_MOD_NAME@ @DUNE_MOD_VERSION@"
+
+/* Define to the one symbol short name of this package. */
+#define PACKAGE_TARNAME "@DUNE_MOD_NAME@"
+
+/* Define to the home page for this package. */
+#define PACKAGE_URL "@DUNE_MOD_URL@"
+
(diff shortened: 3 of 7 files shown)


andlaus:
as usual, travis times out. #1107 should fix that.

joakim-hove:

joakim-hove:
I'm on holiday now, but my general comments are:
1.  I am generally positive to make the dc experience as flawless as possible.
2. I am skeptical to the changes in the config file generation, the current approach is 100% automagic plain cmake, as I see it the manual steps in the current PR is a step in the wrong direction. DC should be able to work with standard cmake, or at least piggyback/extend standard cmake instead of going all manual.

andlaus:
> I am skeptical to the changes in the config file generation, the current approach is 100% automagic plain cmake, as I see it the manual steps in the current PR is a step in the wrong direction. DC should be able to work with standard cmake, or at least piggyback/extend standard cmake instead of going all manual.

as said: better solutions are highly appreciated as long as they work fine with the Dune 2.4/Dune 2.5 cmake build system (the current opm-parserConfig.cmake in master doesn't, at least not on my machine). I could just not come up with anything better, though :(

alfbr:
> better solutions are highly appreciated as long as they work fine with the Dune 2.4/Dune 2.5

You can of course use cmake directly, and if you want it in one command you can put it in a ten-line bash script, just modify according to your needs:

#!/bin/bash                                                                                                                                                           
for folder in opm-common opm-parser opm-material opm-output opm-grid opm-core ewoms opm-simulators
do
cd $folder && rm -rf build
mkdir build && cd build
cmake ..
make -j 4
cd ../..
done

This has the added benefit that you actually now what is going on when things fail, and you can easily use existing cmake knowledge around the net to track it down. Even better benefit, you have the full powers of cmake options at your hand to tweak various build options without having to dig through 1332 lines of shell script in dunecontrol to figure out that you can bypass the whole thing by feeding it a cmake options file with some obscure command line option.

andlaus:
> You can of course use cmake directly, and if you want it in one command you can put it in a ten-line bash script, just modify according to your needs:

I don't really see how this is gonna help, put if you open a PR against my PR, I'd be happy to merge it if @joakim-hove would press green thereafter.

joakim-hove:
> better solutions are highly appreciated as long as they work fine with the Dune 2.4/Dune 2.5 cmake build system (the current opm-parserConfig.cmake in master doesn't, at least not on my machine)

Sorry - you have to do better. The build system in opm is cmake, the current opm-parser cmake setup will create a `opm-parser-config.cmake` file using 100% default / plain vanilla cmake functionality. You are now suggesting to replace that with manual functionality which is not based on plain cmake, but rather geared towards a specific extension/framework(?) on top of cmake - to me that is regressing the opm cmake based build system. So although I am positive to merge changes to the build system which make use of e.g. dune-control easier, I am reluctant to do that when I feel it implies a degradation of the native opm cmake build system. 

> [...] as long as they work fine with the Dune 2.4/Dune 2.5 cmake build 

As I have said the `opm-parser-config.cmake` file currently generated by the opm-parser build system is 100% standard cmake functionality, if the build system in Dune 2.4/2.5 can not utilize that - well then maybe the fix  should be in Dune? 


Mind you - I am not hellbent on these matters - I will listen to good arguments; but would be grateful if others in addition to @andlaus would point out the flaws in my arguments.

andlaus:
first, I'd like to reply to the above: you should probably agree that for most OPM modules, Dune is not "a specific extension/framework", but that it is *the* dependency nowadays. Further, trying to imply that  opm-parser uses the same build system as the remaining OPM modules -- or something related to it -- is quite a bit misleading because it rolls something completely separate. In these regards opm-parser is the exception, not the rule.

anyway, regardless of your position in this matter, there seem to be a substantial number of people who want to use opm-parser in the context of a Dune module. I'm of the considered opinion that we should not actively make life for these people harder than necessary, and as far as I can see, there are two possible solutions: The first is to make the dune build system happy, the second is to give up the pretense that opm-parser is a dune module (i.e. handle it like a fully external library like e.g. ERT).

I'd prefer the first approach, and just so that you know that I generally put my money where my mouth is, be aware that doing this PR cost me the better part of last week and that developing it was not exactly what I consider to be fun. Also, I honestly don't know how this can be done in a better than proposed way -- but as said, I just like to get stuff done, I'm not a dogmatic cmake ninja.

You pick.

joakim-hove:
> Dune is not "a specific extension/framework", but that it is the dependency nowadays.

That I certainly agree - but many (most?) people build opm towards an installed version of Dune. The fact that the dune code is the single most important dependency does not imply that we have to adapt to the dune build system when building opm?

> Further, trying to imply that opm-parser uses the same build system as the remaining OPM modules -- or something related to it -- is quite a bit misleading

Fair enough - but it is plain cmake though.

> anyway, regardless of your position in this matter, there seem to be a substantial number of people who want to use opm-parser in the context of a Dune module. 

Yes - and that is very good; so I fully agree that this must be settled.

> [...] and as far as I can see, there are two possible solutions: The first is to make the dune build system happy, 

That is obviously a solution; you have done a significant effort in that regard. The only thing I am skeptical about in this PR is the non-standard configuration file. If that approach is required I would at least strongly prefer that the settings required by dune were clearly documented as such.

> the second is to give up the pretense that opm-parser is a dune module

Well - I guess it is not a dune module, it is an opm module, which might be made up to look like a dune module. Unless you are actively developing on opm-parser - what would be the drawbrack from this?

But I would say there is a third possibility; fix the dune build system so that it can work with a standard compliant `opm-parser-config.cmake` file.

blattms:
> But I would say there is a third possibility; fix the dune build system so that it can work with a standard compliant opm-parser-config.cmake file.

No offence, but I doubt that opm-parser-config.cmake is standard compliant (why else would one need to manually link the boost libs when using it) and I even doubt that there is a standard (It is CMake...).

BTW if opm-parser has a clean usage of CMake, i.e. sets all the defines, libs, and includes in opm-parser-config.cmake then setting opm-parser_PREFIX there to the root of the installation should also fix this problem. Seems a bit less invasive.

andlaus:
> That I certainly agree - but many (most?) people build opm towards an installed version of Dune.

possibly. the problem is that installed versions of dune get confused by the presence of dune.module while at the same time dune modules that depend opm-parser fail. 

> The fact that the dune code is the single most important dependency does not imply that we have to adapt to the dune build system when building opm?

we should at least make it compatible. so far, that was one of the major features of the opm build system and a lot of effort has been expended on this when the original system was written in 2013 (?).

> Well - I guess it is not a dune module, it is an opm module, which might be made up to look like a dune module. Unless you are actively developing on opm-parser - what would be the drawbrack from this?

the drawback is lacking convenience: to update, clean and build all dune modules that lead to opm-simulators from scratch (and no more), you can use

```
dunecontrol --module=opm-simulators git clean -fdx
dunecontrol --module=opm-simulators git pull
dunecontrol --module=opm-simulators all
```

whereas you woulf have to manually deal with opm-parser every time it needs to be updated. on this matter, ERT is quite annoying in my workflow, BTW.

> But I would say there is a third possibility; fix the dune build system so that it can work with a standard compliant opm-parser-config.cmake file.

I just looked at the opm-parserConfig.cmake file from master: it seems to be garbage because it only defines the following variables:

```
-- opm-parser-DIR=/home/and/tmp/src-dumux/opm-parser/install
-- opm-parser_CONFIG=/home/and/src/opm-parser/opm-parserConfig.cmake
-- opm-parser_CONSIDERED_CONFIGS=/home/and/src/opm-parser/opm-parserConfig.cmake
-- opm-parser_CONSIDERED_VERSIONS=unknown
-- opm-parser_DIR=/home/and/src/opm-parser
-- opm-parser_FOUND=1
-- opm-parser_VERSION_COUNT=0
-- opm-parser_VERSION_MAJOR=0
-- opm-parser_VERSION_MINOR=0
-- opm-parser_VERSION_PATCH=0
-- opm-parser_VERSION_TWEAK=0
```

once this file fixed, dunecontrol _might_ work fine, if the `opm-parser_INSTALLED` and `opm-parser_MODULE_PATH` variables are also added, I'm pretty sure that it will. I also checked if the installed version of `opm-parser-config.cmake` is different: same story. 

As far as I can see, the only reason why the remaining OPM modules work, is that they use [Findopm-parser.cmake](https://github.com/OPM/opm-common/blob/master/cmake/Modules/Findopm-parser.cmake), which does things in your favourite mode ("all manual") and employs a few hacks which are IMO worse than the ones proposed by the present PR.

blattms:
Well, I have to agree that making it a non-dune module is a fix too. Not sure how involved that is for downstream modules, though.

I also agree that this should be fixed upstream. Seems as giving precedence to the version information in the <dune-package>-config-version.cmake files. That would work if opm-parser would provide a version there, which it does not seem to do as of now. I read:
```
```
# this is the version that is installed
set (PACKAGE_VERSION )
```

joakim-hove:
OK - this getting more constructive; I am still on holiday so I am just rambling on github and not actually testing/developing code.

> That would work if opm-parser would provide a version there, which it does not seem to do as of now. I read:

OK - this should be fixed.

> I just looked at the opm-parserConfig.cmake file from master: it seems to be garbage because it only defines the following variables:

OK - this is/should be generated by the standard cmake export functionality, and it has been my main point all along that it should be usable from dune. If this file seems broken that must be fixed - the calls to generate are here: https://github.com/OPM/opm-parser/blob/master/CMakeLists.txt#L233


> BTW if opm-parser has a clean usage of CMake, i.e. sets all the defines, libs, and includes in opm-parser-config.cmake 

Well - that is certainly the goal; I thought it was like this - but if not we must certainly fix it.

andlaus:
I guess that with #1117 this PR is superfluous. Close it at will.

andlaus:
ok, #1117 has been merged, so this one is obsolete. closing.

PR #1110, opened 2017-07-21
changed: avoid use of targets for exports

akva2:
the exported file refers to imported targets through their
name with the expectation that these are defined whenever
the target file is included.

the use of custom imported targets for boost thus becomes
problematic in downstreams, since they are forced to define
these targets prior to opm-parser inclusion.

as the targets are named the same as their library
counterparts it appears to work per murphy's law.
cmake will simply fall back to interpreting the target names as
library names, e.g. it will do -lboost_filesystem.

there are two ways to fix this: either through manually
writing a opm-parser-config.cmake file which defines the boost targets,
then includes the target file, or by forgoing the use of targets for
the linking. 

i have here done the latter.

### DIFF
--- a/lib/eclipse/CMakeLists.txt
+++ b/lib/eclipse/CMakeLists.txt
@@ -126,15 +126,13 @@ add_library(opmparser ${opmparser_SOURCES})
 
 target_link_libraries(opmparser PUBLIC opmjson
                                        ecl
-                                       ${boost_filesystem}
-                                       ${boost_system}
-                                       ${boost_regex}
-                                       ${boost_date_time})
+                                       ${Boost_LIBRARIES})
 target_compile_definitions(opmparser PRIVATE -DOPM_PARSER_DECK_API=1)
 target_include_directories(opmparser
     PUBLIC  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
             $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
             $<INSTALL_INTERFACE:include>
+            ${Boost_INCLUDE_DIRS}
     PRIVATE ${CMAKE_CURRENT_BINARY_DIR}/include
 )
 
--- a/lib/json/CMakeLists.txt
+++ b/lib/json/CMakeLists.txt
@@ -1,7 +1,7 @@
 project(opm-json CXX)
 
 add_library(opmjson JsonObject.cpp $<TARGET_OBJECTS:cjson>)
-target_link_libraries(opmjson PUBLIC ${boost_filesystem})
+target_link_libraries(opmjson PUBLIC ${Boost_FILESYSTEM_LIBRARY})
 
 if (CJSON_LIBRARY)
     target_link_libraries(opmjson PUBLIC cjson)
@@ -14,6 +14,7 @@ add_static_analysis_tests(JSON_SOURCES JSON_INCLUDES)
 target_include_directories(opmjson
     PUBLIC  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
             $<INSTALL_INTERFACE:include>
+            ${Boost_INCLUDE_DIRS}
     PRIVATE $<TARGET_PROPERTY:cjson,INTERFACE_INCLUDE_DIRECTORIES>
 )
 set_target_properties(opmjson PROPERTIES


akva2:
ref old file: https://pastebin.com/gB3BrtCa vs new file: https://pastebin.com/yTVKBZRv

notice the complete absence of the custom boost directory in the old file.

alfbr:
Thank you! Is this ready for testing with dune 2.4 on Debian stable?

akva2:
yes.

joakim-hove:
@bska: Can you run this through your configuration.

alfbr:
Cmake succeeds on opm-parser, and seems to find the out-of-tree Boost fine. Build of opm-parser fails immediately, fails to find boost headers:
Scanning dependencies of target cjson
[  0%] Building C object external/cjson/CMakeFiles/cjson.dir/cJSON.c.o
[  0%] Built target cjson
Scanning dependencies of target opmjson
[  0%] Building CXX object lib/json/CMakeFiles/opmjson.dir/JsonObject.cpp.o
/home/rustad/OPM/opm-parser/lib/json/JsonObject.cpp:26:37: fatal error: boost/filesystem/path.hpp: No such file or directory
 #include <boost/filesystem/path.hpp>

I tested setting Boost location with CMAKE_PREFIX_PATH, and with BOOST_ROOT, both fail with above output.

akva2:
make VERBOSE=1

clean build tree i assume ?

akva2:
there, now it should work for your case as well i believe.

alfbr:
Will test in a couple of hours.

joakim-hove:
The appveyor test is actually very valuable for this particular problem because the Boost layout on the Appveyor build nodes is highly non-standard. Initially it failed with exactly the same error message as @alfbr reported - it works now.

alfbr:
Then we are back again to the original error, linking of boost fails when building opm-core. Did you test at your end?

bska:
> @bska: Can you run this through your configuration.

I'm away from my main workstation at the moment, but I'll try to set up a reasonable environment on this freshly installed laptop Soon[tm].

joakim-hove:
My understanding of this situation is:

1. Cmake standard functionality for imported dependencies was used to transitively pass dependencies on externals like Boost on to downstream modules.
2. The imported dependencies functionality used features from cmake 3.x - which broke on cmake 2.8, and things were shoehorned and hacked to work on cmake 2.8

As I understand it things now fail on Debian stable, which seems to have cmake version 3.7.x - i.e. quite a lot newer than 2.8 we are currently targeting. Maybe we should have a `IF (CMAKE_VERSION)` switch?

akva2:
it's a good understanding, but it rests on some lacking information.

i did test it on stretch and it did/does work for me there (did not try past opm-core but issue was said to be there so..). so it's not as simple as cmake version dependent. while i do not invoke the pebkac, there is some aspect to the way @alfbr does things which have not been communicated.

the only thing that is different in my setup is my use of boost 1.64. i haven't gotten around to rebuilding with 1.58 but i highly doubt that's the issue. so there is something more to it. if i cannot reproduce it's hard to fix..

alfbr:
Finally success :-)

We have multiple issues at play which happened to surface simultaneously. To give a brief summary, a new Debian stable was released at the same time as we discovered the need to go from Dune 2.3.1 to 2.4.1. Flow does work with 2.3.1, but tests breaks due to numerical instabilities and some MPI runs has degraded performance. It was fairly easy to move our Jenkins server over to Dune 2.4.1 since the packages was available in the backports repository. However, we want to migrate the Jenkins server over to the new Debian stable, and Dune 2.4.1 is simply not available there due to build breakage with Boost. Work-around is to build Boost out-of-tree on the new Debian stable, but that broke building of OPM due to newer cmake version.

This patch finally provides a fix, where out-of-tree Boost can be built with OPM also on newer cmake versions, while keeping build working on enterprise linux. However, there are gotchas, so I believe we should continue work to get rid of the rough edges. The gotchas I see here now is the following:
-we need to disable the cmake registry for this patch to work, setting the cmake variable CMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=1
-setting the variable above makes use of suibling build fail with libecl (since suibling build was basically never implemented, it coincidently worked by fetching the suibling directory from the registry)

To me it seems like we finally should consider retiring the use of the cmake registry completely, thus avoiding the non-intuitive cmake switch above. We can consider enabling sibling functionality in libecl (similarly with Dune), but I am afraid this tight integration external dependencies is the wrong approach. Lastly, we may need to consider skipping Dune 2.4.1 completely, and move straight to 2.5, but this will cost quite a few work hours on the quality assurance side.

Conclusively, this patch is good to go from my end. @joakim-hove , maybe you can review, and give it a test run on Red Hat?

joakim-hove:
> @joakim-hove , maybe you can review, and give it a test run on Red Hat?

I will test it on RedHat tomorrow.

joakim-hove:
OK - this works on my RedHat6 setup. Merging.

PR #1111, opened 2017-07-25
Added Schedule::getWells( group, time) overload

joakim-hove:
The new Schedule::getWells( group, time) overload will recursively scan
th group tree rooted at 'group' and return all the wells in that tree.


joakim-hove:

PR #1116, opened 2017-08-09
Change order of section scan.

joakim-hove:


joakim-hove:

PR #1117, opened 2017-08-10
Fix build system

akva2:
This is a continuation of https://github.com/OPM/opm-parser/pull/1108

#1108 corrects the usage of the targets file as a config file. This builds further on this, restoring the ability to use ecl from sibling directories. I have removed  the usage of the OpmParserMacros.cmake as this only serves dunecontrol in combination with installed libraries and not the generic situation.

The last commit is optional. The cmake package registry is icky at best, and it is the cause of much confusion for developers since cmake registers random build directories at (perceived) indeterministic times. In my opinion it cause more headaches than the slight convenience it may offer for end users.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -230,7 +230,29 @@ add_subdirectory(lib/json)
 add_subdirectory(lib/eclipse)
 add_subdirectory(applications)
 
-install(FILES dune.module DESTINATION lib/dunecontrol/opm-parser)
-install(EXPORT opm-parser-config DESTINATION share/cmake/opm-parser)
-export(TARGETS opmjson opmparser FILE opm-parserConfig.cmake)
+set(OPM_PARSER_PREFIX "${CMAKE_INSTALL_PREFIX}")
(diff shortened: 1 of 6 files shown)


alfbr:
Thank you, I will try to get this tested today.

joakim-hove:

joakim-hove:
Anyone who use dune-control who would give this a spin? @andlaus, @dr-robertk, @blattms

akva2:
i have tested with dc.

joakim-hove:

joakim-hove:
@alfbr : I have tested this on my local stock RHEL 6 work environment; where it works nicely.  So we know have 3 x CI + "Works on my machine" - I'm ready to merge this.

alfbr:
Good, just allow me one more hour to check out relevant use cases on Debian.

joakim-hove:
Ok - I will let you merge it then.

alfbr:
Thanks. All tests pass, merging.

PR #1120, opened 2017-08-28
Deck operator<<

joakim-hove:


atgeirr:
Sounds neat, but I guess this is not just for fun... Are you building general "output this as a complete deck" facilities? I am thinking something like loading a deck, changing some items, then writing it as a new modified deck?

joakim-hove:
> Are you building general "output this as a complete deck" facilities? I am thinking something like loading a deck, changing some items, then writing it as a new modified deck?

Yes - that is the general plan; very much on a *minimum effort required* basis - but hopefully it will be valuable.

PR #1121, opened 2017-08-28
Added status message about the location of ecl.

blattms:
There have been various occasions where having this information would have been useful for debugging.


joakim-hove:
Sounds like a smart thing - thank you.

PR #1124, opened 2017-08-30
Policy change: Missing xxxDIMS keyword - go directly to default.

joakim-hove:
For some keywords - like e.g. `EQUIL` the number of records are given as an item of another keyword, `EQLDIMS` in the case of `EQUIL`. The `EQLDIMS` keyword has defaults specified, so strictly speaking it is not necessary to require that keyword.

Currently the code will see it as an error situation if `EQLDIMS` is not there when `EQUIL`is parsed, however it can be handled with the ParseContext system. With this PR the situation of a missing `xxxDIMS` keyword is just plain silently ignored.

This is a change of policy - we can no longer enforce that all required `xxxDIMS` keywords are present, but there should be no change in end functionality. I suggest we merge this - but I am open to other opinions.

### DIFF
--- a/lib/eclipse/Parser/ParseContext.cpp
+++ b/lib/eclipse/Parser/ParseContext.cpp
@@ -60,7 +60,6 @@ namespace Opm {
         addKey(PARSE_UNKNOWN_KEYWORD);
         addKey(PARSE_RANDOM_TEXT);
         addKey(PARSE_RANDOM_SLASH);
-        addKey(PARSE_MISSING_DIMS_KEYWORD);
         addKey(PARSE_EXTRA_DATA);
         addKey(PARSE_MISSING_INCLUDE);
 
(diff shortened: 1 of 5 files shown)


joakim-hove:

OPMUSER:
Don't like this as it will make it difficult to find errors; however, if we write out a WARNING message to the print file saying EQILDIMS not found using defaults instead then at least the user would know what is being used.

joakim-hove:
> Don't like this as it will make it difficult to find errors;

I have no strong opinions on this - it will slightly simplify a quite common testing case, and it might make error reporting in the case of an invalid deck input slightly less accurate. As I said - no strong opinions, would prefer if those responsible for the end-user experience, i.e. flow to voice an opinion @alfbr / @atgeirr.

In case this should be merged I will create followup PRs where that is required,

atgeirr:
To see if I understood correctly: this removes the PARSE_MISSING_DIMS ParseContext control, instead behaving as if we set that to ignore it. I am quite happy with the existing behaviour used by flow (print a warning) which I think is reasonable, since the deck is not "cross-simulator-compatible" in that case. I therefore agree with @OPMUSER on this one.

In general I think the ParseContext controls are a good feature that allows us to resolve tricky questions about error handling in a good way, I hope that feature is not causing trouble in general?

joakim-hove:
> To see if I understood correctly: this removes the PARSE_MISSING_DIMS ParseContext control, instead behaving as if we set that to ignore it

Yes - that is correct.

> since the deck is not "cross-simulator-compatible" in that case. 

Mind you that ECLIPSE will hum along just nicely without the `xxxDIMS` keyword.

> [....]  I hope that feature is not causing trouble in general?

No - absolutely not. This was merely a minor suggestion of a simplification (and ECLIPSE alignment); seems like this PR will be closed unmerged.

PR #1127, opened 2017-09-08
Rft config bug

joakim-hove:


alfbr:
Huge thanks for the rapid effort! Successfully tested, merging as suggested.

PR #1135, opened 2017-09-26
Fix shut well logic when WELOPEN and COMPDAT is combined

totto82:
Do not shut well if all completions are shut using WELOPEN, but new
completions are opened using COMPDAT

### DIFF
--- a/lib/eclipse/tests/ScheduleTests.cpp
+++ b/lib/eclipse/tests/ScheduleTests.cpp
@@ -658,6 +658,72 @@ BOOST_AUTO_TEST_CASE(CreateScheduleDeckWithWELOPEN_TryToOpenWellWithShutCompleti
   BOOST_CHECK_EQUAL(WellCommon::StatusEnum::SHUT, well->getStatus(currentStep));
 }
 
+BOOST_AUTO_TEST_CASE(CreateScheduleDeckWithWELOPEN_CombineShutCompletionsAndAddNewCompletionsDoNotShutWell) {
+  Opm::Parser parser;
+  std::string input =
+          "START             -- 0 \n"
+                  "1 NOV 1979 / \n"
+                  "SCHEDULE\n"
+                  "DATES             -- 1\n"
+                  " 1 DES 1979/ \n"
+                  "/\n"
+                  "WELSPECS\n"
+                  "    'OP_1'       'OP'   9   9 1*     'OIL' 1*      1*  1*   1*  1*   1*  1*  / \n"
+                  "/\n"
+                  "COMPDAT\n"
+                  " 'OP_1'  9  9   1   1 'OPEN' 1*   32.948   0.311  3047.839 1*  1*  'X'  22.100 / \n"
+                  " 'OP_1'  9  9   2   2 'OPEN' 1*   46.825   0.311  4332.346 1*  1*  'X'  22.123 / \n"
+                  " 'OP_1'  9  9   3  9 'OPEN' 1*   32.948   0.311  3047.839 1*  1*  'X'  22.100 / \n"
+                  "/\n"
+                  "DATES             -- 2\n"
+                  " 10  JUL 2008 / \n"
+                  "/\n"
+                  "WELOPEN\n"
+                  " 'OP_1' OPEN / \n"
+                  "/\n"
+                  "DATES             -- 3\n"
+                  " 10  OKT 2008 / \n"
+                  "/\n"
+                  "WELOPEN\n"
+                  " 'OP_1' SHUT 0 0 0 0 0 / \n "
+                  "/\n"
+                  "COMPDAT\n"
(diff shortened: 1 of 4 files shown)


lib/eclipse/tests/ScheduleTests.cpp
```diff
@@ -658,6 +658,53 @@ BOOST_AUTO_TEST_CASE(CreateScheduleDeckWithWELOPEN_TryToOpenWellWithShutCompleti
...
+                  "/\n"
+                  "DATES             -- 3\n"
+                  " 10  OKT 2008 / \n"
+                  "/\n"
+                  "WELOPEN\n"
+                  " 'OP_1' SHUT 0 0 0 0 0 / \n "
+                  "/\n"
+                  "COMPDAT\n"
+                  " 'OP_1'  9  9   1   1 'OPEN' 1*   32.948   0.311  3047.839 1*  1*  'X'  22.100 / \n"
+                  "/\n"
```
joakim-hove: OK - I want a test of the situation:
```txt
WELOPEN
         " 'OP_1' SHUT 0 0 0 0 0 /
/
``` 
*without* the additional `COMPDAT` - in that case we should still get `WellChangeEvent` - and I think that is lost now?

totto82: Isn't "CreateScheduleDeckWithWELOPEN_TryToOpenWellWithShutCompletionsDoNotOpenWell" (The test above) testing exactly that?

joakim-hove: > Isn't [...] (The test above) testing exactly that?

I might misunderstand - but I *think* that if remove the `COMPDAT` on lines 690 - 692 this will happen:

1.  The well will change status to `SHUT`- this I think is correct in the current code, although not tested for?
2. Since the well is actively shut we should "emit" the `WELL_STATE_CHANGE` event - and I am quite certain that will not be set in this case?!

I might be wrong though?

totto82: I have added two more tests addressing (1). 

When should the `WELL_STATE_CHANGE` event be triggered? Is it if all completions are shut after reading all keywords? (After time step 4) 
Should opening new completions trigger a new event? (Time step 5) 
Or should it be triggered within WELOPEN when all completions are shut, and then again in COMPDAT if new completions are open? This will give two events for time step 3, one for time step 4 and one for time step 5 in the test.  
Or is the `COMPLETION_CHANGE` event sufficient? 
Or do I misunderstand your comment. 

I don't have strong opinions on this matter, so please specify what you think is best.

joakim-hove: > I have added two more tests addressing (1).

Very good

joakim-hove: > [...] This will give two events for time step 3, one for time step 4 and one for time step 5 in the test.
Or is the COMPLETION_CHANGE event sufficient?
Or do I misunderstand your comment.

First of all - the *event* noun is slightly misleading, the functionality provided is to set a flag if a certain condition has happened during a timestep. So for instance several events during one timestep is impossible by construction, the event should reflect the net change during the time step, so that if one well is first opened and then closed again during one timestep there should be no `WELL_STATE_CHANGE` event for this timestep[1]. 

> When should the WELL_STATE_CHANGE event be triggered? 

When the return value from `Well::getStatus(time_step)` changes from one timestep to the next - the `WELL_STATE_CHANGE` flag should be set. So the problem with the current pull request is that when all completions are shut the `Well::getStatus( )`method will return closed status, but no `WELL:_STATE_CHANGE` signal will be triggered.

It gets kind of ugly, but I guess it is necessary to spin through the wells at the end of a timestep here: https://github.com/totto82/opm-parser/blob/d0b0dae99c55ac4b09a8e06eeefe6f192ae50aea/lib/eclipse/EclipseState/Schedule/Schedule.cpp#L143 and here: https://github.com/totto82/opm-parser/blob/d0b0dae99c55ac4b09a8e06eeefe6f192ae50aea/lib/eclipse/EclipseState/Schedule/Schedule.cpp#L140 and check if all completions are closed, and then call the updateStatus->closed method.


[1]: The current implementation will probably not generally handle this situation.

totto82: I did as you suggested. Notice that I have changed to logic of the last test. i.e if a well is shut due to all completions being shut, and than new connections is added later, the well is kept shut.

joakim-hove: > Notice that I have changed to logic of the last test. i.e if a well is shut due to all completions being shut, and than new connections is added later, the well is kept shut.

OK - but I guess the logic here is dictated by Eclipse and not us?

totto82: Eclipse is the truth!

joakim-hove: OK - then I will merge this as soon as Jenkins comes clean!

joakim-hove:

PR #1137, opened 2017-09-28
Removed unused function tracking first RFT.

joakim-hove:


joakim-hove:

Issue #1085, opened 2017-06-12
Package registry breaks builds of other OPM modules (was: Undefined symbols for diagnose_relperm)


blattms:
Just updated OPM and cannot compile opm-core anymore:
```
Linking CXX executable bin/diagnose_relperm
CMakeFiles/diagnose_relperm.dir/examples/diagnose_relperm.cpp.o: In function `main':
diagnose_relperm.cpp:(.text.startup+0x17f): undefined reference to `Opm::ParseContext::ParseContext(std::vector<std::pair<std::string, Opm::InputError::Action>, std::allocator<std::pair<std::string, Opm::InputError::Action> > > const&)'
collect2: error: ld returned 1 exit status
CMakeFiles/diagnose_relperm.dir/build.make:113: recipe for target 'bin/diagnose_relperm' failed
make[2]: *** [bin/diagnose_relperm] Error 1
CMakeFiles/Makefile2:329: recipe for target 'CMakeFiles/diagnose_relperm.dir/all' failed
make[1]: *** [CMakeFiles/diagnose_relperm.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2
```

blattms:
I get a similar error when trying to compile flow_ebos.

akva2:
i think you need to rebuild your opm-parser. i changed that constructor from pass by value to pass by reference, and it seems your headers and binaries are not in sync. at least i cannot reproduce on my box, and several builds on jenkins done after the change have passed (though none in opm-core, just flow_ebos).

blattms:
*grrrr* and here we go again: Seems like the package registry might have kicked in again. An outdated unrelated opm version gets chosen:
```
mblatt@loft11004:~/DUNE-test/opm-simulators/opm-parallel$ cat CMakeCache.txt | grep opm-parser_ROOT
opm-parser_ROOT:STRING=/home/mblatt/DUNE-test/opm-parser/opm-analyze
mblatt@loft11004:~/DUNE-test/opm-simulators/opm-parallel$ ls -l ~/.cmake/packages/opm-parser/
total 36
-rw-r--r-- 1 mblatt mblatt 41 Jun  6 14:11 0e598f77d119826fdec4fbb245f33e1d
-rw-r--r-- 1 mblatt mblatt 48 Jun  6 19:20 20937009162acf5223ed4e167f718916
[...]

PR #1141, opened 2017-09-29
OPMify sibling search for ecl.

blattms:
Previously only ecl/build was considered during the search. Now the subdirectory name can be arbitrary. One can also use a common build root (/path/to/build_root/{ecl|opm-parser}) located somewhere else.

This change is in line with the other OPM modules.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,9 +50,30 @@ else()
 endif()
 
 #-----------------------------------------------------------------
-
-if(SIBLING_SEARCH AND EXISTS ${PROJECT_SOURCE_DIR}/../libecl/build)
-  set(ecl_DIR ${PROJECT_SOURCE_DIR}/../libecl/build)
+if(SIBLING_SEARCH AND NOT ecl_DIR)
+  # guess the sibling dir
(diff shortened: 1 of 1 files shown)


pgdr:
> This change is in line with the other OPM modules.

Well the build systems in the other OPM modules are not exactly pinnacle of build systems.

May I ask: Why do you need this?  Which problem is it that you are trying to solve?

blattms:
You support only this setup:

I we build in opm-parser/build then we try to find ecl in ecl/build.
This commit changes the restriction on the name of the build directory
and allows out of source builds wher we have a build root directory
with one subdirectory per module that has the same name as the module.

pgdr:
But _why_ do you need to search for ecl?  You can just build it and cmake will help you find it (in `~/.cmake`).

```
.cmake
└── packages
    ├── ecl
    │   └── 91ffab8e046906477c8d8f93416af172
    └── res
        └── 63adb6c276fea84ac9d3946d119c8ef5
```

akva2:
Cmake package cache is utterly broken.

blattms:
@pgdr Please read OPM/opm-parser#1085 and test it yourself.

PR #1143, opened 2017-10-05
UnitSystem: Add Support for PVT-M Unit Conventions

bska:
The PVT-M unit convention is essentially the same as the Metric collection, except that pressure values are measured in atmospheres rather than bars.

### DIFF
--- a/lib/eclipse/Deck/Deck.cpp
+++ b/lib/eclipse/Deck/Deck.cpp
@@ -146,6 +146,8 @@ namespace Opm {
          * lab, and field over metric, for as long as we have no easy way of
          * figuring out which was requested last.
          */
+        if( this->hasKeyword( "PVT-M" ) )
+            this->activeUnits = UnitSystem::newPVT_M();
         if( this->hasKeyword( "LAB" ) )
             this->activeUnits = UnitSystem::newLAB();
         if( this->hasKeyword( "FIELD" ) )
--- a/lib/eclipse/Units/UnitSystem.cpp
+++ b/lib/eclipse/Units/UnitSystem.cpp
@@ -299,6 +299,93 @@ namespace {
         "SCC/RCC", /* oil inverse formation volume factor */
         "SCC/RCC", /* water inverse formation volume factor */
     };
+
+    static const double to_pvt_m[] = {
+        1,
+        1 / PVT_M::Length,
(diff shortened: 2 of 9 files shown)


joakim-hove:
This is beautifull - will certainly merge; just one minor detail/question - I guess a switch should be added here: https://github.com/OPM/opm-parser/blob/master/lib/eclipse/Deck/Deck.cpp#L137?

bska:
> I guess a switch should be added here:

Ooooops.  You are of course correct.  I will update my patch set.  Thank you for pointing out the deficiency!

bska:
> I will update my patch set

Updated in commit 754ac5a.  On a side note, I don't think the comment above that particular code section really reflects the structure of the implementation. The actual code seems to prefer METRIC above all other conventions.

bska:
One more thing if I may: The "PVT-M" keyword is a perfectly good string, but it is not a valid C++ identifier.  While I **believe** that the JSON specification below is correct, I'd like to verify that.  Is there a good place to put a unit test to verify that the parser recognises "PVT-M" as a keyword?

--- 

```json
{"name" : "PVT_M" ,
 "sections" : ["RUNSPEC"],

 "deck_names" : [
  "PVT-M"
 ]
}
```

joakim-hove:
> Is there a good place to put a unit test to verify that the parser recognises "PVT-M" as a keyword?

E.g. here: https://github.com/OPM/opm-parser/blob/master/lib/eclipse/tests/ParserTests.cpp

```C++

Parser parser;
const auto deck = parser.parseString( "PVT-M", ParseContext() );
BOOST_CHECK(  deck.hasKeyword( "PVT-M" ));
```

bska:
>> Is there a good place to put a unit test to verify that the parser recognises "PVT-M" as a keyword?
>
> E.g. here ([`ParserTests.cpp`](https://github.com/OPM/opm-parser/blob/master/lib/eclipse/tests/ParserTests.cpp)).

Thanks.  I've updated my patch set to add the verification (commit 346c2d6).  I added tests for the other two unit conventions too, which prompted adding a JSON description for `LAB`.

bska:
jenkins build this please

joakim-hove:
jenkins chill please

bska:
> jenkins chill please

:smile: 

Thanks a lot.

Issue #1140, opened 2017-09-29
Using shared libs installed libopmparser links to libopmjson in build folder


blattms:
During my tests with installing OPM I discovered this:
```
ldd ~/test-install2/lib/libopmparser.so
	linux-vdso.so.1 (0x00007ffdb5ccd000)
	libopmjson.so.2017 => /home/mblatt/test-install2/lib/libopmjson.so.2017 (0x00007f65f9bef000)
        ....
```
I know the default are static libs but we might want to fix this anyway.

blattms:
I guess to fix this there is some serious refactoring needed due the use of subprojects.

PR #1144, opened 2017-10-06
First try to find ecl with deactivated package registry.

blattms:
This allows to configure with-DCMAKE_INSTALL_PREFIX and find the installed version and not the build tree of ecl (that was used the last time).

I just did an installation test with `-DSIBLING_SEARCH=Off -DCMAKE_INSTALL_PREFIX=/path/to/install` and had a hard time figuring out why opm-parser still found the sibling build of ecl and not the installed version. Well there is still the package registry lying around as ecl is using it...

This is for discussion. One can also either explicitly set ecl_DIR (not tested) or set CMAKE_PREFIX_PATH (tested) to the installed version. But this seems more complicated than needed for users/admins.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,7 +78,7 @@ endif()
 
 # Give us a chance to find ecl installed to CMAKE_INSTALL_PREFIX.
 # We need to deactivate the package registry for this.
-find_package(ecl NO_CMAKE_SYSTEM_PACKAGE_REGISTRY)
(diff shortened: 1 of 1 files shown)


joakim-hove:

joakim-hove:
Markus: I am on holiday, but please merge this - possibly in cooperation with @akva2 when you see fit.

blattms:
CMake works in mysterious ways. With this PR libopmparser is now dynamically linked to the installed version of libopmjson. Closes #1140

blattms:
Yes, please.

PR #1145, opened 2017-10-09
obtaining the depth for the completions based on segment information

GitPaean:
it corrects the depth information when more than one completions is related to one segment, basically it provides one more feature of multi-segment well.


joakim-hove:
Test

GitPaean:
> Test 
I have that in mind, while I need an Eclipse result to verify the algorithm first. That is why it is marked as `WIP` for now.

GitPaean:
I pushed a new commit 553ade74fcf03b698ff22f85920fc4892bd886e2 to the branch, while I can not see the PR updated with the new commit?

GitPaean:
This PR works for the only example we have that has more than one perforation related to one of the segments. I does not break any our hand-made example either. I do realize the algorithm can fail for some situation. For example, if the perforation is related one segment, while it is farther than the segment from the well head, should we use the segment it is related to to do extrapolation or based on the inlet segment of the segment owns the perforation to do interpolation to calculate the depth of the perforation. 

It sounds confusing probably, while basically, it is something we need to create example to test with.

GitPaean:
The PR is ready for review now. The test is updated.

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

lib/eclipse/EclipseState/Schedule/MSW/Compsegs.cpp
```diff
@@ -168,18 +168,61 @@ namespace Opm {
...
-                compseg.m_center_depth = segment_set[segment_location].depth();
+                compseg.calculateCenterDepthWithSegments(segment_set);
             }
         }
     }
 
+    void Compsegs::calculateCenterDepthWithSegments(const SegmentSet& segment_set) {
+
+        // the depth and distance of the segment to the well head
+        const int segment_location = segment_set.numberToLocation(m_segment_number);
```
joakim-hove: Instead of using variable `segment_location`[1] I think this would be easier to read if it were something like this:
```C++
const auto& segment = segment_set.getFromNumber( m_segment_number );
// And then continue with the segment reference.
```



[1]: `segment_location` is an integer and should probably be named something like `segment_index` or `segment_nr`? The name `segment_location`hints at something with dimension length?

lib/eclipse/EclipseState/Schedule/MSW/Compsegs.cpp
```diff
@@ -168,18 +168,61 @@ namespace Opm {
...
+        const double segment_depth = segment_set[segment_location].depth();
+        const double segment_distance = segment_set[segment_location].totalLength();
+
+        // for top segment or the distance of segment and perforation are the same
+        // no interpolation is needed
+        const double center_distance = (m_distance_start + m_distance_end) / 2.0;
+        if (center_distance == segment_distance || m_segment_number == 1) {
```
joakim-hove: I would assume that `m_segment_number == 1` would be enough to check if we are actually considering the top segment? Iff the distance test is indeed required it needs some "epsilon protection" - checking two `double` values for equality with `==` is probably not correct?

lib/eclipse/EclipseState/Schedule/MSW/Compsegs.cpp
```diff
@@ -168,18 +168,61 @@ namespace Opm {
...
+            for (const int inlet : segment_set[segment_location].inletSegments()) {
+                const int inlet_location = segment_set.numberToLocation(inlet);
+                if (segment_set[inlet_location].branchNumber() == m_branch_number) {
+                    interpolation_segment = inlet;
+                    break;
+                }
+            }
+        }
+
+        assert(interpolation_segment != 0);
```
joakim-hove: This seems so fragile it should be protected in release as well - i.e.:
```C++
if (interpolation_segment == 0) throw std::runtime_error(" ... ")
```

lib/eclipse/EclipseState/Schedule/MSW/Compsegs.cpp
```diff
@@ -168,18 +168,61 @@ namespace Opm {
...
+
+        // performing the interpolation
+        const int interpolation_location = segment_set.numberToLocation(interpolation_segment);
+        const double interpolation_detph = segment_set[interpolation_location].depth();
+        const double interpolation_distance = segment_set[interpolation_location].totalLength();
+
+        const double depth_change_segment = segment_depth - interpolation_detph;
+        const double segment_length = segment_distance - interpolation_distance;
+
+        assert(segment_length != 0.);
```
joakim-hove: Throw.

lib/eclipse/EclipseState/Schedule/MSW/Segment.cpp
```diff
@@ -109,6 +109,19 @@ namespace Opm {
         m_volume = volume_in;
     }
 
+    const std::vector<int>& Segment::inletSegments() const {
+        return m_inlet_segments;
+    }
+
+    void Segment::clearInletSegments() {
+        // in case the segment structure changes during the simulation
```
joakim-hove: I do not like this; preferably the variable stays constant for the entire simulation - and then no `clearXxx( )` method is required. In the case an immutable segment is not feasible we should perform the clearing exactly when and where it is needed - this "mutate the state just to certain" makes it more difficult to reason about the code.

joakim-hove:
I must admit I find this code quite difficult to follow; some comments on the roles of the three classes `Compsegs`, `Segment` and `SegmentSet` would be valuable.

joakim-hove:
It is good that you have added a test with an input deck, but I will not merge this before you have taken the full blue pill and written some basic instantion tests:
```C++

std::vector<CompSegs> compsegs;

// Add three segments ...
compsegs.emplace_back( i,j,k,..... ); 
compsegs.emplace_back( ... );
compsegs.emplace_back( ... );

// Create a segment set and a completion set
SegmentSet segments;
CompletionSet completions;
...

// Call the static process routines in the Compsegs class
Compsegs::processCOMPSEGS( compesgs, segments );
Compsegs::updateCompletionsWithSegment( compsegs, completions );

// Check the results - what is actually the expected outcome here?

```

GitPaean:
> I must admit I find this code quite difficult to follow

Yes. it was difficult to program too. Basically, the algorithm follows the result, and some part is not very logical.

lib/eclipse/EclipseState/Schedule/MSW/Compsegs.cpp
GitPaean: Yes. The "m_segment_number == 1" is for top segment, the other one is consider the case that the distance is input as exactly the same with the segment distance.  If they are slightly different, then we go to interpolation. We can remove `center_distance == segment_distance`, let it always do the interpolation, which might kill slightly more computation, which we might not care.  How do you think?

lib/eclipse/EclipseState/Schedule/MSW/Segment.cpp
GitPaean: I do not know.  To me, one thing can come later while extremely difficult about multisegment well is that, we introduce a multisegment well, we run for some period, then we change some of the segments changing the segment structure, then we run  for some other period.  To change the segment structure, you only input the changed or the new ones.  `clearInletSegments()` is to consider that.  I can remove this function for now for sure without breaking anything.  I have to admit that I am not familiar with `DynamicState`. If you think with my case, each instance will be immutable with `DynamicState`, I will remove this function.

GitPaean:
>  It is my impression that the Compsegs class is only used temporarily during the construction phase? If that is indeed the case I would suggest that Compsegs.hpp header file is located with source files - and not installed.

Yes. `Compseg` is used to build the relation between completions and segments.  Do you want me to move the `Compsegs.h` out from `lib/eclipse/include` folder  to the `lib/eclpse/EclipseState/...`?

lib/eclipse/EclipseState/Schedule/MSW/Compsegs.cpp
GitPaean: the naming is hard here. `segment number` is a ID, can be any positive number and does not mean anything about the position of the segment in the whole well.  `segment_nr` I am worrying can be a little confusing with `segment number`. 

Basically, `segment_location` means the location of the segment in the whole vector of segments. To me, `segment_location` and `segment_index` are equally good / bad. `well index` in well can be something related to transmissibility / injectvity (all different though).  Anyway, I will change all the location to index, and add some comments there.

joakim-hove:
> Do you want me to move the Compsegs.hpp out from lib/eclipse/include folder to the lib/eclpse/EclipseState/...?

Yes.

lib/eclipse/EclipseState/Schedule/MSW/Compsegs.cpp
GitPaean: I decided to remove the comparison. It will increase the computation a little bit, while we do not worry about that.

lib/eclipse/EclipseState/Schedule/MSW/Compsegs.cpp
GitPaean: fixed.

lib/eclipse/EclipseState/Schedule/MSW/Compsegs.cpp
GitPaean: fixed.

lib/eclipse/EclipseState/Schedule/MSW/Segment.cpp
joakim-hove: > I can remove this function for now for sure without breaking anything.

Do that.

GitPaean:
I think I have addressed all the comments except the instance test.  I will try to add it as soon as possible. 

Please let me know if you have more comments.  Thanks for the review on the weekend, which is very helpful.

joakim-hove:
> Yes. Compseg is used to build the relation between completions and segments.

OK - I have looked a bit more on this code. As I understand it the net effect of `Compsegs` implementation is to update a `CompletionSet` instance with segment information on the completion objects. My suggestion is then that the *only* public function should be something like:

```C++
CompletionSet segment_completions( const DeckKeyword& compsegs, const CompletionSet& input_completions, const SegmentSet& segments) {
     CompletionSet new_completions;
     ...
     return new_completions;
}
```
Here I have just created a brand new `CompletionSet` instead of mutating the existing set. Whether this is implemented with a structure like `Compsegs` will then just be an implementation detail. For the completions I would reccomend a "copy" constructor like:

```C++
Completion::Completion( const Completion& initial, int segment_nr, double center_depth) {
     
}
```
then the segment information can be attached at construction time instead of mutating the objects afterwards.

lib/eclipse/EclipseState/Schedule/MSW/Compsegs.cpp
joakim-hove: Well - to me `segment_index` - or just `index` is much better.

GitPaean:
> OK - I have looked a bit more on this code. As I understand it the net effect of Compsegs implementation is to update a CompletionSet instance with segment information on the completion objects. My suggestion is then that the only public function should be something like:
``` CompletionSet segment_completions( const DeckKeyword& compsegs, const CompletionSet& input_completions, const SegmentSet& segments) {
     CompletionSet new_completions;
     ...
     return new_completions;
} ```

I addressed the second one. For the first one, it involves re-factoring the Compsegs struct.  This PR is only supposed to add one more feature (although not very trivial in some sense). Can we address this in a separate PR specifically for this comment with more discussion?  If it is acceptable to you, can we also add the instance test in that PR later? There were no instance test, although probably it was also due to me.

File: lib/eclipse/EclipseState/Schedule/MSW/Compsegs.cpp (hunk 2 above)
Comment (joakim-hove) [reply lib/eclipse/EclipseState/Schedule/MSW/Compsegs.cpp]: > [...] which might kill slightly more computation, which we might not care. How do you think?

Always interpolate - clarity is paramount for this code.

**Comment by joakim-hove** [comment]
> This PR is only supposed to add one more feature (although not very trivial in some sense). Can we address this in a separate PR [...]

yes

> If it is acceptable to you, can we also add the instance test in that PR later? 

No

**Comment by GitPaean** [comment]
>>  Do you want me to move the Compsegs.hpp out from lib/eclipse/include folder to the lib/eclpse/EclipseState/...?

> Yes.

Any suggestion how to include this header file in a tests now?

**Comment by joakim-hove** [comment]
> Any suggestion how to include this header file in a tests now?

Well; welcome to the world of testing. The `Compsegs` structure is an internal detail - what you really want to test is only the ability to add segment information to completions - the difficulties you are now facing clearly tell you this. 

Maybe you should create a very small public segment structure which has only `segment_nr` and `segment_depth` - or maybe just a `std::map<int,double>` - after all *that is all the interesting information extracted from the `COMPSEGS` keyword* - and thereby also what the testing should focus on.

**Comment by GitPaean** [comment]
> Well; welcome to the world of testing. The Compsegs structure is an internal detail - what you really want to test is only the ability to add segment information to completions - the difficulties you are now facing clearly tell you this.

But basically this test will not test the functionality introduces in this PR, which makes adding this test is to fix the old missing stuff. 

Should I move this header file back to the include folder, which probably looks ugly to you? Any other suggestion?

**Comment by joakim-hove** [comment]
> Should I move this header file back to the include folder, which probably looks ugly to you? Any other suggestion?

My suggestion:

 - Make a public header in the include hierarchy `Compsegs.hpp` which has only one exported symbol:
```C++
std::map<int,double> parseCOMPSEGS( const DeckKeyword& kw);
```
which will parse a `COMPSEGS` keyword and return a sgement_nr / segment_depth mapping. How that function is implemented - is a detail; but I gase basing it on the now private (as entirely in the .cpp file) `Compsegs` class makes sense.

GitPaean:
> Make a public header in the include hierarchy `Compsegs.hpp` which has only one exported symbol:
``` 
std::map<int,double> parseCOMPSEGS( const DeckKeyword& kw);
```
> which will parse a COMPSEGS keyword and return a sgement_nr / segment_depth mapping. How that function is implemented - is a detail; but I gase basing it on the now private (as entirely in the .cpp file) Compsegs class makes sense.

Sorry, but I need a little more details here. Do we need to use the implementation in the current `Compsegs`? then we might still facing how to include the `Compsegs.hpp` there. 
I can see some parts related to MSW can be improved here and there.  Since we agreed to re-factor the implementation `Compsegs`, can we put this instance test along with that part, since the instance test will be more for the interface design.  The functionality introduced in this PR is well tested with an updated deck test. 

There will be a new PR this or next week to address these both which really needs your input. Does it sound sensible to you?

atgeirr:
I think we got it now, we'll make a public header with a single function that will be tested and also used from handleCOMPSEGS().

joakim-hove:
> I think we got it now, we'll make a public header with a single function that will be tested and also used from handleCOMPSEGS().

Yes - that was my suggestion.

GitPaean:
Another question, what is the best way to create a Keyword like `COMPSEGS`? Do we include an example anywhere? Thanks.

GitPaean:
Is `parseString` the correct thing to do in this test to create the `Keyword` object?

joakim-hove:
> Is parseString the correct thing to do in this test to create the Keyword object?

yes

GitPaean:
Thanks. 

What the naming style requirement for a pure function header?

Should I capitalize `updatingCompletionsWithSegments.hpp` ?

atgeirr:
> Should I capitalize updatingCompletionsWithSegments.hpp ?

In opm-core and opm-simulators headers for functions have not been capitalized, just as you have done. For opm-parser I do not know, but I assume that is also ok there?

joakim-hove:
> For opm-parser I do not know, but I assume that is also ok there?

Sure.

GitPaean:
I added a public header file and added a test. Please review.

lib/eclipse/EclipseState/Schedule/MSW/Compsegs.cpp
```diff
@@ -189,9 +237,8 @@ namespace Opm {
             const int j = compseg.m_j;
             const int k = compseg.m_k;
 
-            auto new_completion = completion_set.getFromIJK( i, j, k );
-            new_completion.attachSegment(compseg.m_segment_number, compseg.m_center_depth);
-            completion_set.add(new_completion);
+            const Completion completion = completion_set.getFromIJK( i, j, k );
```
joakim-hove: This should be a reference.

GitPaean: Thanks. fixed.

GitPaean:
It break the downstream compilation due to the interface in `SegmentSet`, which will be addressed in OPM/opm-simulators#1279.

atgeirr:
> It break the downstream compilation due to the interface in SegmentSet, which will be addressed in OPM/opm-simulators#1279.

When updates to that PR have been pushed we'll make a final Jenkins run for that PR which should include this.

@joakim-hove Do you have further requirements for this PR, or can we merge this when Jenkins is happy?

GitPaean:
OPM/opm-simulators#1279 has been updated to incorporate the interface change from this PR and waiting for the Jenkis result.

Local building and running are normal.

joakim-hove:
> @joakim-hove Do you have further requirements for this PR, or can we merge this when Jenkins is happy?

Thank you for the effort - yes this can be merged now. I will let you do it when appropriate.

joakim-hove:
> yes this can be merged now.

Cool if you squash it on the way in.

atgeirr (approved):
I have read the main routine and it looks fine, but to be honest I am mostly writing this review to get a green button for merging, now that all is ready...

GitPaean:
Now, the only public interface related to `COMPSEGS` keyword is the `updatingCompletionsWithSegments()` . Please let me know if there is anything more is required to improve the `COMPSEGS` related and the test, I would like to address them. @joakim-hove .

PR #1146, opened 2017-10-10
Use correct default clone directory in sibling search.

blattms:
Previously we assumed it to be ecl (like the project name).
That is not correct. With this commit we now use the correct
default clone directory, libecl, in the sibling search.


akva2:
i'm a bit anxious to merge due to the appveyor failure. however it seems to have failed for a while and it does not at all look related to this change, so let's yolo this thing.

PR #1147, opened 2017-10-10
Fix search path of libecl for installed opm-parser.

blattms:
Previously, ecl_DIR was always set to the build directory used even
for the installed cmake configuration file. This patch at least allows
to overwrite that location and for the installed version tries to guess
the correct location. The guess is that libecl and opm-parser are installed
under the same prefix. As a fallback we use the package registry.


joakim-hove:
I do not manage to follow all these build system updates. Please merge when fit.

PR #1149, opened 2017-10-25
update redhat packaging

akva2:
Should be backported to release branch after merge.


akva2:
The version number in master is irrelevant as such, we never package using it. It is the other changes (ecl, toolchain on rhel7) that was the point.

PR #1150, opened 2017-10-26
New keywords

nairr:
Added parser support for the following keywords:
AQUCT
AQUTAB
RSCONST
OPERATER


andlaus:
as far as I can see the changes here are good. I'd prefer 'edit' patches to be squashed into the main one, though...

PR #1152, opened 2017-11-06
Added message to update list of missing features.

joakim-hove:


lib/eclipse/share/keywords/keyword_list.cmake
```diff
@@ -2,6 +2,13 @@
 # explicitly. In the root CMakeLists.txt which includes this file
 # individual dependencies on all the Json files is established, in
 # addition to an explicit dependency on this file.
+#
+# When a keyword is added here the parser will recognize the keyword and
+# correctly internalize it as a DeckKeyword, but the simuator 'flow' in
```
atgeirr: Appreciated!

BTW, typo "simuator" here, "therefor" two lines down.

joakim-hove: OK thanks - typos fixed.

PR #1153, opened 2017-11-06
Added missing include in Equil.hpp

andreabrambilla:
When running test in opm-parser, this never showed up as a problem since
Equil.hpp seems to be always included after <vector> is included (either
directly or by including other files). However, this can become a
problem when using opm-parser in other projects (eg, sunbeam)

If possible, this should backported to 2017.10/final. Alternatively, a workaround is to include <vector> before Equil.hpp whenever necessary

### DIFF
--- a/lib/eclipse/include/opm/parser/eclipse/EclipseState/InitConfig/Equil.hpp
+++ b/lib/eclipse/include/opm/parser/eclipse/EclipseState/InitConfig/Equil.hpp
@@ -1,6 +1,8 @@
 #ifndef OPM_EQUIL_HPP
 #define OPM_EQUIL_HPP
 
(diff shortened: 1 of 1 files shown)


joakim-hove:
This of course a correct fix; and I'm merging it now - but *has it* created problems so far, or is that just something you (correctly) envision it might?

pgdr:
I  think it crashed sunbeam.

andreabrambilla:
@joakim-hove it was a problem when compiling sunbeam on travis. In sunbeam I added a workaround (https://github.com/Statoil/sunbeam/pull/52/commits/4f7a476d320c899a93d3414347ef10eae8451e6e), but this should be thge correct fix

PR #1154, opened 2017-11-16
Scatter vector

stefoss23:
This pr suggets a scatterVector function in EclipseGrid:
```C++
template<typename T>
std::vector<T> scatterVector(const std::vector<T>& input_vector, T def = -1) const
```
Argument input_vector can have either size equal to the numbers of grid cells or the active number of grid cells, else an exception is thrown. For for former case,, the return value equals the input_vector. For the latter case: `output_entry[i] = input_vector[j]`, where `i` is the global index and `j` is the active index. For inactive cells, `output_entry[i] = def`.

### DIFF
--- a/lib/eclipse/EclipseState/Grid/EclipseGrid.cpp
+++ b/lib/eclipse/EclipseState/Grid/EclipseGrid.cpp
@@ -898,6 +898,7 @@ namespace Opm {
         return this->activeMap;
     }
 
+
     void EclipseGrid::resetACTNUM( const int * actnum) {
         ecl_grid_reset_actnum( m_grid.get() , actnum );
         /* re-build the active map cache */
--- a/lib/eclipse/include/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.hpp
+++ b/lib/eclipse/include/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.hpp
@@ -144,6 +144,27 @@ namespace Opm {
             }
         }
 
+        template<typename T>
+        std::vector<T> scatterVector(const std::vector<T>& input_vector, T def = -1) const {
+            if (input_vector.size() == this->getCartesianSize())
+                return input_vector;
+
+            if (input_vector.size() != this->getNumActive())
+                throw std::invalid_argument("Input vector must have full or active size");
+
+            std::vector<T> scatter_vector( this->getCartesianSize() );
+
+            for (size_t i = 0; i < this->getCartesianSize(); i++) {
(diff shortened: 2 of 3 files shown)


atgeirr:
Three comments:
 - Commits with no description are not acceptable (I guess squishing them is fine here).
 - I believe code like this already exists in a multitude of places, perhaps we should try to find some common ground (although it may not be worth it in this particular context), what do you think @bska?
 - The proposed interface is guaranteed worst-case from a performance point of view: you must take a copy even if you could just use what you have. There are many ways to address this, depending on how you plan on using it, one way could be to take an rvalue reference, and in the first case you just move it to the destination.

joakim-hove:
> Commits with no description are not acceptable (I guess squishing them is fine here).

No worries - we'll squash!

> I believe code like this already exists in a multitude of places

On would think so - but at least in terms of the `EclipseGrid` the answer is no.

> There are many ways to address this, depending on how you plan on using it, one way could be to take an rvalue reference, and in the first case you just move it to the destination.

Makes sense.

bska:
> I believe code like this already exists in a multitude of places, perhaps we should try to find some common ground (although it may not be worth it in this particular context), what do you think @bska?

On principle we really should try to centralise this type of operation.  It is common/important enough that having one implementation that's usable for all cases is worthwhile.

My main reservation with this particular proposal is that it encourages using vectors of global size.  I'd much prefer that we (the OPM project) go in a direction of most of our vectors sized according to the number active cells rather than the number of global/declared/Cartesian cells.  This is of course not always possible, but it is possible in more cases than we currently do so.

joakim-hove:
> My main reservation with this particular proposal is that it encourages using vectors of global size.

This is created to support this: https://github.com/OPM/opm-output/pull/216 i.e. output of arbitrary integer fields like e.g. `PVTNUM`[1] to the INIT file. I *think* downstream tools like Resinsight expect these fields to have `nx*ny*nz` elements.

bska:
> I **think** downstream tools like Resinsight expect these fields [i.e., integer vectors] to have `nx*ny*nz` elements.

If they do, then they're imposing too much structure on the vectors.  In particular, the INIT file `*NUM` vectors are supposed to be sized according to the number of active cells.  Real-type vectors can be sized either according to the number of active cells or the number of global cells.  The only real restriction is that the PORV vector must be sized according to the number of global cells in order to provide an implicit definition of the set of active cells.

joakim-hove:
> In particular, the INIT file *NUM vectors are supposed to be sized according to the number of active cells.

Ahhh - this was embarrassing; you are right. Then we close this PR; and update opm/opm-output#216 to just through whatever size is present in the input argument.

PR #1155, opened 2017-11-20
AQUTAB table parsing

nairr:
Added parser support for the AQUDIMS and the AQUTAB table. An example of the AQUTAB table is:

> AQUTAB 
> ------------------------------------ 
> --   ANALYTIC AQUIFIER Rd = 2.0 
> ------------------------------------ 
> --   TD     PD 
>    0.01     0.112 
>    0.05     0.229 
>    0.10     0.315 
>    0.15     0.376 
>    0.20     0.424 
>    0.22     0.443 
>    0.24     0.459 
>    0.26     0.476 
>    0.28     0.492 
>    0.30     0.507 
>    0.32     0.522 
>    0.34     0.536 
>    0.36     0.551 
>    0.38     0.565 
>    0.40     0.579 
>    0.42     0.593 
>    0.44     0.607 
>    0.46     0.621 
>    0.48     0.634 
>    0.50     0.648 
>    0.60     0.715 
>    0.70     0.782 
>    0.80     0.849 
>    0.90     0.915 
>    1.00     0.982 
>    2.00     1.649 
>    3.00     2.316 
>    5.00     3.649 
>   10.00     6.982 
>   20.00    13.649 
>   30.00    20.316 
>   50.00    33.649 
>  100.00    66.982 
>  200.00   133.649 
>  300.00   200.316 
>  500.00   333.649 
> 1000.00   666.982 / 
> -- --------------------------- 
> --   ANALYTIC AQUIFIER Rd = 3.0 
> -- --------------------------- 
> --   TD     PD 
>    0.01     0.112      
>    0.05     0.229      
>    0.10     0.315      
>    0.15     0.376      
>    0.20     0.424      
>    0.25     0.469      
>    0.30     0.503      
>    0.40     0.564      
>    0.50     0.616      
>    0.52     0.627      
>    0.54     0.636      
>    0.56     0.645      
>    0.60     0.662      
>    0.65     0.683      
>    0.70     0.703      
>    0.75     0.721      
>    0.80     0.740      
>    0.85     0.758      
>    0.90     0.776      
>    0.95     0.791      
>    1.00     0.806      
>    1.20     0.865      
>    1.40     0.920      
>    1.60     0.973      
>    2.00     1.076      
>    3.00     1.328      
>    4.00     1.578      
>    5.00     1.828      
>   10.00     3.078     
>   20.00     5.578     
>   30.00     8.078     
>   50.00    13.078    
>  100.00    25.578   
>  200.00    50.578   
>  300.00    75.578   
>  500.00   125.578  
> 1000.00   250.578 /


lib/eclipse/EclipseState/Tables/TableManager.cpp
```diff
@@ -142,6 +144,21 @@ namespace Opm {
...
+            const auto& record = keyword.getRecord(0);
+            int mxnaqn  = record.getItem<AQUDIMS::MXNAQN>().get< int >(0);
+            int mxnaqc = record.getItem<AQUDIMS::MXNAQC>().get< int >(0);
+            int niftbl = record.getItem<AQUDIMS::NIFTBL>().get< int >(0);
+            int nriftb = record.getItem<AQUDIMS::NRIFTB>().get< int >(0);
+            int nanaqu = record.getItem<AQUDIMS::NANAQU>().get< int >(0);
+            int ncamax = record.getItem<AQUDIMS::NCAMAX>().get< int >(0);
+            int mxnali = record.getItem<AQUDIMS::MXNALI>().get< int >(0);
+            int mxaaql = record.getItem<AQUDIMS::MXAAQL>().get< int >(0);    
+            m_aqudims = std::make_shared<Aqudims>( mxnaqn , mxnaqc , niftbl , nriftb , nanaqu, ncamax, mxnali, mxaaql );
```
joakim-hove: This should not use shared pointers.

joakim-hove:
Will review properly later today

joakim-hove:

joakim-hove:
ÒK - what is here is fine; apart from the use of `std::shared_ptr<>` for the `Aqudims` member variable. But as discussed previously I would prefer if the aquifer configuration was assembled in a separate object, as a starting point this could just be:

```C++
class Aquifer

    private:
         std::vector<Aquatable> aquifer_tables;
```
The existing `Aquatable` implementation is fine, it is just a question of creating this tiny container class and collecting the `Aquatable`instances in a `std::vector<>` instead of a `TableManager`. The reason I would like this is that the aquifer tables are semantically quite different - in particular the  collection of them - from the other PVT/relperm tables. 

I am not familiar with aquifer usage/requirements from the downstream code, it will be natural to extend the `Aquifer` class *when* that is clear.

Don't forget to add a (very) small test for the new functionality.

joakim-hove:
Please rebase on current master - #1161 should have fixed the weird parsing requirements of `AQUTAB`

joakim-hove:
Is this in progress?

joakim-hove:
Is this in progress?

nairr:
@joakim-hove - Thanks for the shifted tables fix. yes this is in progress.

lib/eclipse/EclipseState/Aquifer.cpp
```diff
@@ -0,0 +1,338 @@
...
+            && this->m_state == rhs.m_state
+            && this->m_direction == rhs.m_direction
+            && this->m_segment_number == rhs.m_segment_number
+            && this->m_center_depth == rhs.m_center_depth;
+    }
+
+    bool Completion::operator!=( const Completion& rhs ) const {
+        return !( *this == rhs );
+    }
+}
```
joakim-hove: This looked very weird - the existing `Completion` in a new file called `Aquiffer.c`?

lib/eclipse/include/opm/parser/eclipse/EclipseState/Aquifer.hpp
```diff
@@ -0,0 +1,170 @@
...
+              return default_pressure_;
+          }
+        else {
+              return Aqutab_tables_.at(m_influenceFnTableNum-1).getPressureColumn();
+          }
+        }
+
+    private:
+        size_t m_aquId , m_datDepth , m_aqPerm , m_aqPoro , m_aqCt , m_aqRad , m_aqThickness , m_aqInfluenceAngle, m_waterPressTableNum, m_influenceFnTableNum;
+	Deck& deck_;
```
joakim-hove: No - you must find another solution - internalizing a private `Deck&` will not be merged.

lib/eclipse/include/opm/parser/eclipse/EclipseState/Tables/Aqudims.hpp
```diff
@@ -0,0 +1,112 @@
...
+            m_niftbl( ParserKeywords::AQUDIMS::NIFTBL::defaultValue ),
+            m_nriftb( ParserKeywords::AQUDIMS::NRIFTB::defaultValue ),
+            m_nanaqu( ParserKeywords::AQUDIMS::NANAQU::defaultValue ),
+            m_ncamax( ParserKeywords::AQUDIMS::NCAMAX::defaultValue ),
+            m_mxnali( ParserKeywords::AQUDIMS::MXNALI::defaultValue ),
+            m_mxaaql( ParserKeywords::AQUDIMS::MXAAQL::defaultValue )
+            
+        { }
+
+        explicit Aqudims(const Deck& deck) :
```
joakim-hove: The content of the `Aqudims.hpp` file is OK; but it the implementation is *untested header only* - that will not be merged. The pattern in the rest of opm-parser is that implementation is in a `.cpp` file; and I do not see this file as sufficient reason to break that pattern.

I am not religious about the use of `.cpp` files for the implementation, but in the case of a header-only implementation the requirement of a test is 100.00% - otherwise we get into the very unfortunate situation where problems - even compile errors - are not caught before the code is explicitly invoked, which might be in not-yet-written code three modules down the line.

lib/eclipse/include/opm/parser/eclipse/EclipseState/Tables/AquanconTable.hpp
```diff
@@ -0,0 +1,48 @@
...
+#ifndef OPM_PARSER_AQUANCON_TABLE_HPP
+#define	OPM_PARSER_AQUANCON_TABLE_HPP
+
+#include "SimpleTable.hpp"
+
+namespace Opm {
+
+    class DeckItem;
+
+    class AquanconTable : public SimpleTable {
```
joakim-hove: The `AQUANCON` keyword does not implement a table; I might grudgingly accept the (private) use of `SimpleTable` inheritance and the `TableManager` for internalizing the `AQUTAB` keyword, but not for the `AQUANCON` keyword - this is *not* a table.

I don't know the full details of this, but you should implement something like an `AquiferConfig` class and that should among other things include the internalized content of the `AQUANCON` kewyord.

lib/eclipse/include/opm/parser/eclipse/EclipseState/Aquifer.hpp
```diff
@@ -0,0 +1,170 @@
...
+            return m_waterPressTableNum;
+        }        
+
+        size_t getAqInfluenceFunctionTableNumber() const
+        {
+            return m_influenceFnTableNum;
+        }        
+      
+        std::vector<double> getTimeColumn() const{
+	if (deck_.hasKeyword("AQUTAB")) {
```
joakim-hove: This just shows that basing the implementation on the `TableManager` is wrong. My suggestion is that the `Aquifer` class is based on something like:
```C++
typedef struct {
     std::vector<double> time;
     std::vector<double> p;
} influence_fn;

std::vector<influence_fn> functions;
```

This can of course be implemented in many ways, and personally I think it would have been nice to return a callable instance instead of raw `std::vector<double>` - but anyway the current implementation which is based on storing a private `Deck&` and checking that to get logic right - will not be merged. Sorry.

lib/eclipse/EclipseState/AquiferCT.cpp
```diff
@@ -0,0 +1,20 @@
+#include <opm/parser/eclipse/EclipseState/AquiferCT.hpp>
+
+namespace Opm {
+
+    AquiferCT::AquiferCT(const EclipseState& eclState, const Deck& deck)
+    {
+        m_aquifers = init_object(eclState, deck);
```
joakim-hove: I would prefer if the `init_objects` method was removed, this should just be the constructor.

lib/eclipse/EclipseState/AquiferCT.cpp
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    std::vector<AquiferCT::AQUCT_data> AquiferCT::getAquifers()
```
joakim-hove: 1. Should this maybe just return *one* `Aquifer::AQUCT_data` instance? 
2. The return value should (probably) be a `const &` and not a copy.
3. The method as a whole should be marked as `const`

lib/eclipse/include/opm/parser/eclipse/EclipseState/AquiferCT.hpp
```diff
@@ -0,0 +1,155 @@
...
+                    int inftableID, pvttableID;
+                    // Perforation cell id
+                    std::vector<int> cell_id;
+                    // Variables constants
+                    double  phi_aq , //aquifer porosity
+                            d0,   //aquifer datum depth
+                            C_t , //total compressibility
+                            r_o , //aquifer inner radius
+                            k_a , //aquifer permeability
+                            c1, // 0.008527 (METRIC, PVT-M); 0.006328 (FIELD); 3.6 (LAB)
```
joakim-hove: These can be initialized here:
```C++
...
double c1 = 0.008527; // What the f... is this?
double c2 = 6.283;       //
```

lib/eclipse/include/opm/parser/eclipse/EclipseState/AquiferCT.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+            const double getAqPorosity(size_t aquiferIndex);
+      
+    private:
+        
+        std::vector<AquiferCT::AQUCT_data> m_aquifers;
+
+        // Initialize function
+        inline std::vector<AquiferCT::AQUCT_data>
+        init_object(const EclipseState& eclState, const Deck& deck)
```
joakim-hove: This should be *in* the constructor.

lib/eclipse/include/opm/parser/eclipse/EclipseState/AquiferCT.hpp
```diff
@@ -0,0 +1,155 @@
...
+                {
+                    const auto& aqutabTable = eclState.getTableManager().getAqutabTables().getTable(aquctParams.at(aquctRecordIdx).inftableID - 1);
+                    const auto& aqutab_tdColumn = aqutabTable.getColumn(0);
+                    const auto& aqutab_piColumn = aqutabTable.getColumn(1);
+                    aquctParams.at(aquctRecordIdx).td = aqutab_tdColumn.vectorCopy();
+                    aquctParams.at(aquctRecordIdx).pi = aqutab_piColumn.vectorCopy();
+                }
+                else
+                {
+                    set_default_tables(aquctParams.at(aquctRecordIdx).td,aquctParams.at(aquctRecordIdx).pi);
```
joakim-hove: The defaults should be added anyway - irrespective of whether the deck has aquifer keywords.

lib/eclipse/include/opm/parser/eclipse/EclipseState/AquiferCT.hpp
```diff
@@ -0,0 +1,155 @@
...
+            if (!deck.hasKeyword("AQUCT")){
+                std::cout << "Nothing is shown! Where is AQUCT!????" << std::endl;
+            }
+
+            const auto& aquctKeyword = deck.getKeyword("AQUCT");
+
+            std::vector<AquiferCT::AQUCT_data> aquctParams;
+            // Resize the parameter vector container based on row entries in aquct
+            // We do the same for aquifers too because number of aquifers is assumed to be for each entry in aquct
+            aquctParams.resize(aquctKeyword.size());
```
joakim-hove: No - avoid resizing. Instead the code should be rewritten to create `AquiferCT::AQUCT_data` instances in the `for` loop below, and those should be added to the `aquctParams` vector using `push_back`or `emplace_back`.

lib/eclipse/include/opm/parser/eclipse/EclipseState/AquiferCT.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+                aquctParams.at(aquctRecordIdx).cell_id.push_back(cellID);
+            }
+
+            return aquctParams;
+        }
+
+        inline void set_default_tables(std::vector<double>& td, std::vector<double>& pi)
+        {
+            std::vector<double> default_pressure_ = { 0.112, 0.229, 0.315, 0.376, 0.424, 0.469, 0.503, 0.564, 0.616, 0.659, 0.702, 0.735,
```
joakim-hove: Here initialize a reference with a local variable; might be some C++ copying saves you - but it does not look good.

joakim-hove:
Please follow the suggestion [here](https://github.com/OPM/opm-parser/pull/1155#pullrequestreview-84010445) of splitting the PR in three parts. Then we can complete and merge one-at-a-time; right now it is a bit unwieldy.

lib/eclipse/EclipseState/AquiferCT.cpp
kel85uk: The idea of it is that either opm-grid or opm-simulator can iterate through all the carter tracy aquifers easily, with the data being self contained in each vector element. Agreed that other classes must not be able to modify the class private members! I'll change to the points 2 and 3 accordingly at a later PR as my main focus is in opm-simulator this week.

lib/eclipse/EclipseState/AquiferCT.cpp
kel85uk: Can be done according to your review. It's up to @nairr how he wants to implement this. I made it as a skeleton/starting point.

lib/eclipse/include/opm/parser/eclipse/EclipseState/AquiferCT.hpp
kel85uk: Why avoid the resize in this situation if I may ask? Because if I already know the number of elements the vector container should have prior to the for loop, it is still safe as I am also using the at function, no? I am genuinely curious as I am not a super C++ programmer myself, and would like to learn more.

lib/eclipse/EclipseState/AquiferCT.cpp
joakim-hove: > The idea of it is that either opm-grid or opm-simulator can iterate through all the carter tracy aquifers easily, with the data being self contained in each vector element.

OK - I do not know how this will be used; maybe your suggestion is the best way to do it.

> I'll change to the points 2 and 3 accordingly 

Good

lib/eclipse/include/opm/parser/eclipse/EclipseState/AquiferCT.hpp
joakim-hove: What you have done is certainly safe; and the best from a performance point view. My argument is not really about C++, but just a general coding principle - where I think your approach is less readable and more difficult to maintain (not very difficult to read/maintain - mind you; but still). An objective argument for side of the story is that you create *couplings* between different parts of the code:
```C++
vector.resize( );         <----------\
...                                  |
...                                  |
for ( ) {                            |
     vector.at( ) = ...;   <---------/
}
```

non-local couplings generally reduce readability and maintainability. But nothing to get religious about.

nairr:
Closing since this PR has been split into 3 separate PR's

PR #1158, opened 2017-11-23
Thermal stuff

andlaus:
This is required to properly support simulations which consider thermal effects and conserve energy. Although the corresponding simulator code already works, it is still a work in progress to some extend and I want to do more testing before reviewing it makes much sense. That said, I feel confident enough that this PR covers the vast majority of changes required on the input-parsing side so that some feedback is solicited.


andlaus:
travis complains about hitting its timeout, appveyor seems to be speechless. Note that AppVeyor's debug build succeeded but its release build failed, I strongly suspect a timeout as well. Fun!

lib/eclipse/EclipseState/Runspec.cpp
```diff
@@ -52,12 +54,13 @@ std::ostream& operator<<( std::ostream& stream, const Phase& p ) {
...
 using un = std::underlying_type< Phase >::type;
 
-Phases::Phases( bool oil, bool gas, bool wat, bool sol, bool pol ) noexcept :
+Phases::Phases( bool oil, bool gas, bool wat, bool sol, bool pol, bool energy ) noexcept :
     bits( (oil ? (1 << static_cast< un >( Phase::OIL ) )     : 0) |
           (gas ? (1 << static_cast< un >( Phase::GAS ) )     : 0) |
           (wat ? (1 << static_cast< un >( Phase::WATER ) )   : 0) |
           (sol ? (1 << static_cast< un >( Phase::SOLVENT ) ) : 0) |
-          (pol ? (1 << static_cast< un >( Phase::POLYMER ) ) : 0) )
+          (pol ? (1 << static_cast< un >( Phase::POLYMER ) ) : 0) |
```
joakim-hove: That we treat energy as a phase must be the extended phase concept :-)

joakim-hove:
OK - this seems to fit nicely into the general scheme; so in that respect there is no reason not merge apart - a test or two maybe?

However as a general question; which I guess should go as much to e.g. @atgeirr and @alfbr. The polymer, solvent, and now thermal stuff certainly increases the complexity of the code. Where is the ambition:

1. Worlds f...ing best black-oil simulator.
2. A very versatile simulator.

??

lib/eclipse/EclipseState/Runspec.cpp
andlaus: the same applies to polymer and solvent, though. see the commit comment for this:

https://github.com/OPM/opm-parser/pull/1158/commits/1703ad5cb00875314b11da381d103a897d229620

andlaus:
> a test or two maybe?

sure. any ideas for a meaningful one? (I can crank out a deck which contains most of the new keywords, but IMO the utility of that would be rather limited.)

alfbr:
> Worlds f...ing best black-oil simulator.

Absolutely, the sky is the limit :-)

atgeirr:
> A very versatile simulator.

Definitely!
```
make cake | tee eatit > keepit
```

OPMUSER:
Alf,

Do you have access to the commercial simulator's test decks?

/*E*/

	

/*I*/

	

/*P*/

	

/*C*/

	

Managing Director

T:  +(65)-9173-7031

F:  +(65)-6732-2382

E: david.baxendale@eipc.co

alfbr:
> Do you have access to the commercial simulator's test decks? 

Probably, but they are copyright Schlumberger I believe, so not available to OPM. In this case I assume @joakim-hove simply asks for unit tests to accompany new keywords, which is how it is done for the parser, and consists of providing a simple use case example for each keyword.

For more general test decks in integration tests (i.e., testing simulator correctness using a combination of keywords) we hope to step up the efforts significantly the coming year. We are currently very far from 100% test coverage of functionality supported in flow, and this needs to change. Using full field cases is not a feasible approach due to computational intensiveness (in addition to no being allowed to open them up), so smaller tailored decks will need to be created. You will find some examples of such decks in opm-data.

OPMUSER:
Well I could we could take the commercial simulator test decks and 
reformat to the OPM documentation style and use them,

/*E*/

	

/*I*/

	

/*P*/

	

/*C*/

	

Managing Director

T:  +(65)-9173-7031

F:  +(65)-6732-2382

E: david.baxendale@eipc.co

alfbr:
We will need to write the decks from scratch to avoid any legal issues, so I think we should aim to make even better tests :-)

lib/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -714,6 +717,26 @@ namespace Opm {
...
     }
 
+    void Schedule::handleWTEMP( const DeckKeyword& keyword, size_t currentStep) {
+        for( const auto& record : keyword ) {
+            const std::string& wellNamePattern = record.getItem("WELL").getTrimmedString(0);
+
+            for (auto* well : getWells(wellNamePattern)) {
+                // TODO: Can this be done like this? Setting the temperature only has an
+                // effect on injectors, but specifying this for producers won't hurt and
+                // wells can also switch their injector/producer status. Note that
```
joakim-hove: Regarding this comment block: As I understand it there is an undetonated bomb due to: "Note modifying the injector properties for producer wells currently leads to a very weird segmentation fault downstream." - I guess that should be understood/fixed before this is merged?

lib/eclipse/share/keywords/001_Eclipse300/W/WATDENT
```diff
@@ -10,8 +10,8 @@
   "comment": "item, the RM is _very_ inconsistent. (it says that the default for the second item is 1.67/degR in",
   "comment": "FIELD units which is the same the 3e-4/K for METRIC but four orders of magnitude off!)",
   "items" : [
-    {"name" : "REFERENCE_TEMPERATURE" , "value_type" : "DOUBLE", "dimension" : "AbsoluteTemperature", "default": 527.67 },
-    {"name" : "EXPANSION_COEFF_LINEAR" , "value_type" : "DOUBLE", "dimension" : "1/AbsoluteTemperature", "default": 1.67e-4 },
-    {"name" : "EXPANSION_COEFF_QUADRATIC" , "value_type" : "DOUBLE", "dimension" : "1/AbsoluteTemperature*AbsoluteTemperature", "default": 9.26e-7 }
+    {"name" : "REFERENCE_TEMPERATURE" , "value_type" : "DOUBLE", "dimension" : "AbsoluteTemperature", "default": 293.15 },
```
joakim-hove: Holy diver - do we have two *different* default temperatures?

joakim-hove:
> I can crank out a deck which contains most of the new keywords, but IMO the utility of that would be rather limited.)

I agree that testing this is not trivial, it is mostly just more of the same, of functionality which is already there. But still I would like (limited) test utilizing these crazy temperature conversions.

lib/eclipse/EclipseState/Runspec.cpp
atgeirr: I agree with the commit comment, that ConservedQuantities is more precise. Not sure that it is practical/wanted/needed to switch though. I'm sure that would lead to different annoyances!

lib/eclipse/EclipseState/Schedule/Schedule.cpp
andlaus: @GitPaean, @totto82, @atgeirr: someone of you probably needs to have a go at this. as far as I understood is that the well controls code gets confused if production wells have injection properties. That said, WTEMP is likely not the only keyword which only makes sense for injection wells, so this kind of problem is probably already present in the current opm-parser code in other contexts.

lib/eclipse/EclipseState/Schedule/WellInjectionProperties.cpp
```diff
@@ -27,6 +27,7 @@ namespace Opm {
     WellInjectionProperties::WellInjectionProperties() {
         surfaceInjectionRate=0.0;
         reservoirInjectionRate=0.0;
+        temperature=273.15 + 15.56;
```
joakim-hove: Please avoid the use of magic numbers - the `273.15`can be found [here](https://github.com/OPM/opm-parser/blob/f763c862a6e1efc8b2d0d1e5d047f78aae31c9e8/lib/eclipse/include/opm/parser/eclipse/Units/Units.hpp#L161) - can add another named constant for `15.56`.

lib/eclipse/share/keywords/001_Eclipse300/W/WATDENT
andlaus: at least if the documentation and what I've been told about it is correct.

andlaus:
> But still I would like (limited) test utilizing these crazy temperature conversions.

do you mean the rtemp()? also remarkable is the fact that SPECROCK uses *volumetric* heat capacity while SPECHEAT uses *mass specific* ones. (combine that with the fact that the heat capacities need to be integrated to get the enthalpies and you have a recipe for a lot of fun.)

lib/eclipse/EclipseState/Schedule/WellInjectionProperties.cpp
andlaus: Ok, changed&pushed. for the 15.56 degrees Celsius, I misused the default value of the STCOND keyword...

joakim-hove:
Hello - it is not entirely clear to me whether this is still work in progress, or if you are waiting to get it merged. I will fire off a Jenkins round now - but assuming that comes home clean I am ready to merge this.

joakim-hove:

andlaus:
> it is not entirely clear to me whether this is still work in progress, or if you are waiting to get it merged.

As far as I'm aware, it can be merged but if you want to have a unit test, I don't mind if it stays open for a few more days. (I'm pretty busy ATM, sorry.) The unit test likely will not be very original, though.

pgdr:
> The unit test likely will not be very original, though.

I have never heard of originality as a requirement for a good test.

andlaus:
> I have never heard of originality as a requirement for a good test.

I did: A pointless test is -- in my not so humble opinion -- worse than no test at all!

pgdr:
>> I have never heard of originality as a requirement for a good test.
>
> I did: A pointless test is -- in my not so humble opinion -- worse than no test at all!

Thanks, this explains quite a lot.

joakim-hove:
For this particular code I agree with Andreas that testing will just be more of the existing; there is no new functionality in the current PR. Merging

PR #1160, opened 2017-11-24
remove the configure wrapper script

andlaus:
this used to provide autotools compatibility, but it has not been working for a while. Thus it is IMO better to remove it in order not to mislead people.


atgeirr:
@joakim-hove merge this?

PR #1162, opened 2017-11-29
Add support for WTHPH+WBHPH keywords

sveinung-r:


joakim-hove:
The code in itself looks perfectly OK, but there have been *insane* amounts of confusion regarding BHP in this code, and I fear the current PR will add to that. I will look carefully through - hopefully we can guard against this confusion by just using better names, but I am not certain. The problem is that the term BHP is used in (at least) three different ways:

1. The flowing bottom hole pressure in the well - something calculated by the simulator.
2. A BHP control limit imposed by the user, and used as a well boundary condition.
3. The historical BHP values when the well is run in historical mode - which is what this PR is about.

What I fear is that there is some non-trivial interaction between 2 and 3 above, need to verify that before this can be merged.

joakim-hove:
I quite frankly do not fully understand if this is enough; but I think this should be enough:

1. The new member variables should be renamed `BHPH` and `THPH` for *historical* mode.
2. In the function updating production properties the `historical` flag must be consulted, and only if this is `true` the `BHPH` and `THPH` member variables should be set.

sveinung-r:
I opened an issue ( #1163  ) regarding the ambiguities in WellProductionProperties and WellInjectionProperties as these are used for representing both history matching and control data, with a boolean property separating the two. Having separate types for history matching and control would enable terms such as BHP (as well as Oil production rate etc. ) to be interpreted as historic data or targets/limits from context. 

As far as I understand, case 1 will not be represented in WellProductionProperties and WellInjectionProperties.

GitPaean:
Do the keywords `WTHPH` and `WBHPH` only affects the summary output?  The output might need to wait the `WCONHIST` and `WCONINJH` parsing to be updated / corrected first.  This PR looks to me is trying to partly to correct the treatment related to `WCONINJH`, and the real problem here is actually the one you pointed out in #1163.

joakim-hove:
> Do the keywords WTHPH and WBHPH only affect the summary output?

These are not keywords, but suggested class member names. Should only affect output.

sveinung-r:
To clarify, WTHPH and WBHPH are keywords for the summary output (THPH and BHPH are suggested class member names). This PR is adding support for historic BHP and THP data to be accessed in opm-output in order to support these keywords, as these are not read from WCONHIST and WCONINJH as of now.

GitPaean:
I have not looked at `WCONINJH`, but for `WCONHIST`, the observed values were read as limits. Correcting this will impact the simulator running. I will try to provide a solution.

sveinung-r:
Which values were read as limits?

GitPaean:
> Which values were read as limits?

If I understand the function correctly, 

```
WellProductionProperties WellProductionProperties::history(
```

sveinung-r:
The rates are read into the same variables for limits and historic data. However, this is not the case for BHP and THP. Since the BHPLimit variable is used for other purposes in the code this can not be done for this variable. Using BHPH and THPH could help to make this distinction somewhat clearer. 

 It is true that solving issue #1163 would solve this. Should I fix this PR or await an alternative solution?

GitPaean:
A question about SUMMARY keyword, do we support customization the output of SUMMARY, or what we output through summary is fixed by the simulator. 

The definition of the keyword `SUMMARY`  is rather simple without much content. 

The support of the keyword WTHPH+WBHPH should be defined in the `SUMMARY` first? or we will always output them, then it is a little messy for non-history matching wells.

GitPaean:
> It is true that solving issue #1163 would solve this. Should I fix this PR or await an alternative solution?

Fixing #1163 will involve much more than this PR's purpose (for output), maybe you can wait a little bit. Hopefully, we will be able to provide a fix soon.

joakim-hove:
> A question about SUMMARY keyword, do we support customization the output of SUMMARY, or what we output through summary is fixed by the simulator.

Yes - the parser checks the `SUMMARY` section and creates a `SummaryConfig` object with all the requested summary output. What we do not check/validate very well is if the summary output requested is consistent with the abilities of the current simulation; i.e. if you ask for oil rate in a gas+water system you will get no error/warnings.

> The support of the keyword WTHPH+WBHPH should be defined in the SUMMARY first? 

yes - and that is already implemented and works.

> or we will always output them, then it is a little messy for non-history matching wells.

We will only output them if asked for - but we will output them also for non-history matching wells if they are requested. I guess you could call that a logical f...up.

GitPaean:
> yes - and that is already implemented and works.

good, then we will focus on make the correct values in place

joakim-hove:
> Fixing #1163  will involve much more than this PR's purpose (for output), maybe you  can wait a little bit. Hopefully, we will be able to provide a fix soon.

The plan is to merge this PR (#1162) quite soon, that should be a pure additive change not conflicting with anything. 

I think I agree that a Good Solution&trade; involves splitting the well properties objects into a historical class and a prediction class - along the lines of #1032, but experience has shown that these topics (historical/prediction/rates/controls/ ...) *is* a massive conceptual clusterf..k in Eclipse - so the chance of getting this correct and well implemented in our code is maybe not so large.

GitPaean:
> The plan is to merge this PR (#1162) quite soon, that should be a pure additive change not conflicting with anything.

Sure. Please feel free to do that. My focus is making the treatment of history matching wells correct. Especially, when we remove the usage of the WellControls from opm-core, using the injection and production properties from opm-parser directly, it will look bad if we redo the same trick with WellControls again with the injection and production properties from opm-parser.

lib/eclipse/tests/ScheduleTests.cpp
```diff
@@ -2178,3 +2178,40 @@ BOOST_AUTO_TEST_CASE(handleWEFAC) {
...
+        "19 JUN 2007 / \n"
+        "SCHEDULE\n"
+        "DATES             -- 1\n"
+        " 10  OKT 2008 / \n"
+        "/\n"
+        "WELSPECS\n"
+        " 'P' 'OP' 9 9 1* 'OIL' 1* / \n"
+        " 'I' 'OP' 9 9 1* 'WATER' 1* / \n"
+        "/\n"
+        "WCONHIST\n"
```
joakim-hove: Add a `WCONPROD` keyword with well `P2`- set the `BHP` and `THP` items of that well, and verify that the `BHPH` and `THPH` variables have *not* been set.

lib/eclipse/EclipseState/Schedule/WellProductionProperties.cpp
```diff
@@ -86,6 +86,9 @@ namespace Opm {
                 throw std::invalid_argument("Setting CMODE to unspecified control");
         }
 
+        p.BHPH = record.getItem("BHP").getSIDouble(0);
```
joakim-hove: The `BHP` and `THP` items do not have default values, so this must be "protected" with `hasValue(0)` checks.

lib/eclipse/EclipseState/Schedule/Schedule.cpp
```diff
@@ -740,6 +740,13 @@ namespace Opm {
             }
             properties.predictionMode = false;
 
+            properties.BHPH = record.getItem( "BHP" ).hasValue(0)
```
joakim-hove: The `BHPH` and `THPH` elements should be initialized to zero; no need to explicitly assign to 0.0.

joakim-hove:
jenkins build this with downstreams opm-output=222 please

PR #1164, opened 2017-12-03
Added one-argument ParseContext constructor.

joakim-hove:


joakim-hove:

atgeirr:
Just one thing: one-argument constructors should be explicit unless you specifically want to have implicit conversion.

joakim-hove:
> Just one thing: one-argument constructors should be explicit unless you specifically want to have implicit conversion.

Ehhh - well; it is marked as `explicit` in the header file.

atgeirr:
> Ehhh - well; it is marked as explicit in the header file.

Yes of course! Sorry for the noise, I must have looked at the cpp file and not noticed.

andlaus:
>> Ehhh - well; it is marked as explicit in the header file.
>
> Yes of course! Sorry for the noise, I must have looked at the cpp file and not noticed.

did you just deliver an argument that all function bodies should go beneath their declarations? ;)

PR #1165, opened 2017-12-03
New sched ctx

joakim-hove:
Added Schedule constructor which takes `EclipseState`argument.


joakim-hove:

PR #1169, opened 2017-12-07
WIP: adding VFP table from GRUPNET

ReedOnly:
VFP table from GRUPNET keyword is added to Group in EclipseState for further use in Sunbeam


joakim-hove:

PR #1170, opened 2017-12-11
add more thermal keywords and the associated infrastructure

andlaus:
this is a small follow-up to #1158. It just adds the THC* keyword family to specify thermal conductivity. BTW: the approach of this keyword family is just a much too complicated way to set constant conductivity. That's because the final heat conductivity does not depend on the current solution at all, i.e. instead of just computing the arithmetic average of all fluid phases, it would be better to weight each phase conductivity with the phase's saturation.


joakim-hove:
jenkins run this with downstreams please

akva2:
seems my edit was too late.

akva2:

PR #1177, opened 2018-01-05
Carter-Tracy aquifer object 

nairr:
A C-T aquifer object to collect information from keywords AQUCT, AQUANCON and AQUTAB

### DIFF
--- a/lib/eclipse/EclipseState/AquiferCT.cpp
+++ b/lib/eclipse/EclipseState/AquiferCT.cpp
@@ -1,0 +1,79 @@
+/*
+  Copyright (C) 2017 TNO
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <opm/parser/eclipse/EclipseState/AquiferCT.hpp>
+
+namespace Opm {
+
+    AquiferCT::AquiferCT(const EclipseState& eclState, const Deck& deck)
+    { 
+        if (!deck.hasKeyword("AQUCT"))
+            return;
+
+        const auto& aquctKeyword = deck.getKeyword("AQUCT");
+
+        for (auto& aquctRecord : aquctKeyword){
+  
+            AquiferCT::AQUCT_data data;
+    
+            data.aquiferID = aquctRecord.getItem("AQUIFER_ID").template get<int>(0);
+            data.h = aquctRecord.getItem("THICKNESS_AQ").getSIDouble(0);
+            data.phi_aq = aquctRecord.getItem("PORO_AQ").getSIDouble(0);
+            data.d0 = aquctRecord.getItem("DAT_DEPTH").getSIDouble(0);
+            data.C_t = aquctRecord.getItem("C_T").getSIDouble(0);
+            data.r_o = aquctRecord.getItem("RAD").getSIDouble(0);
+            data.k_a = aquctRecord.getItem("PERM_AQ").getSIDouble(0);
+            data.theta = aquctRecord.getItem("INFLUENCE_ANGLE").getSIDouble(0);
+            data.c1 = 0.008527; // We are using SI
+            data.c2 = 6.283;
+            data.inftableID = aquctRecord.getItem("TABLE_NUM_INFLUENCE_FN").template get<int>(0);
+            data.pvttableID = aquctRecord.getItem("TABLE_NUM_WATER_PRESS").template get<int>(0);
+
+            // Get the correct influence table values
+            if (data.inftableID > 1){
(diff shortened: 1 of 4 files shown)


joakim-hove:
**AQUANCON** 

The `AQUANCON` keyword is used to connect an aquifer to the cells of the grid, according to the documentation the aquifer itself can be defined using a range of different keywords. In the current PR the AQUANCON behavior is implemented inside the `AQUCT` class; since the AQUANCON keyword can be used in conjunction with other types of aquifer descriptions I want the `AQUANCON`implementation moved into a separate class *outside* the AQUCT implementation.

The `AQUANCON` class should be constructible independent of whether the deck has the`AQUANCON`keyword or not - if the keyword is not in the deck the `AQUANCON` keyword should just return zero in some `size()` like function. The AQUANCON constructor should not take a full `EclipseState` as input argument, but only an `EclipseGrid`:

```C++
Aquancon::Aquancon(const Deck& deck, const EclipseGrid grid) {
    if (deck.hasKeyword("AQUANCON)) {
         ///
    }
}
```

I strongly suggest that you make one PR with only the `AQUANCON` implementation first, and then finalize with the `AquCT` implementation.

lib/eclipse/EclipseState/AquiferCT.cpp
```diff
@@ -0,0 +1,120 @@
...
+            }
+        const auto& aquctKeyword = deck.getKeyword("AQUCT");
+        // Resize the parameter vector container based on row entries in aquct
+        // We do the same for aquifers too because number of aquifers is assumed to be for each entry in aquct
+        m_aquct.resize(aquctKeyword.size());
+        for (size_t aquctRecordIdx = 0; aquctRecordIdx < aquctKeyword.size(); ++ aquctRecordIdx) 
+            {
+            const auto& aquctRecord = aquctKeyword.getRecord(aquctRecordIdx);
+
+            m_aquct.at(aquctRecordIdx).aquiferID = aquctRecord.getItem("AQUIFER_ID").template get<int>(0);
```
joakim-hove: I want the  Aquancon object to consist of a  std::vector<AquanCell>

nairr:
Yes, the AQUANCON can be used in conjunction with any analytical aquifer. I agree with you completely here. Thanks for breaking down the steps for me. Will make the suggested changes and get back to you.

>  if the keyword is not in the deck the AQUANCON keyword should just return zero in some size() like function

E100 doesn't give any error message as well if AQUANCON is not part of the DATA file, even when there is an aquifer description.

nairr:
PR #1190 introduces a separate Aquancon object. I've additionally removed the aquancon collection from this AquiferCT object

akva2:
just so it is not forgotten: this has to be rebased before it is pulled (contains a merge).

nairr:
@akva2  - thanks for the heads up - rebasing...

nairr:
@joakim-hove - I've added a unit test here as well. Is this now merge-able?

lib/eclipse/EclipseState/AquiferCT.cpp
```diff
@@ -0,0 +1,80 @@
...
+    AquiferCT::AquiferCT(const EclipseState& eclState, const Deck& deck)
+    { 
+        if (!deck.hasKeyword("AQUCT"))
+            return;
+
+        const auto& aquctKeyword = deck.getKeyword("AQUCT");
+        // Resize the parameter vector container based on row entries in aquct
+        // We do the same for aquifers too because number of aquifers is assumed to be for each entry in aquct
+        m_aquct.resize(aquctKeyword.size());
+        for (size_t aquctRecordIdx = 0; aquctRecordIdx < aquctKeyword.size(); ++ aquctRecordIdx)
```
joakim-hove: This can be rewritten using a range based for loop over the keyword:
```C++
for (const auto& record : keyword) {
    ...
}
```

nairr:
Alright, I've made the changes

joakim-hove:
I will merge this when it has been rebased and the green merge button has reappeared.

joakim-hove:
Please rebase this and I will merge it!

atgeirr:
Maybe too late to change now, but would not AquiferCarterTracy be a better name for the class than AquiferCT?

lib/eclipse/include/opm/parser/eclipse/EclipseState/AquiferCT.hpp
```diff
@@ -0,0 +1,95 @@
...
+            const int getAqInflTabID(size_t aquiferIndex);
+            const int getAqPvtTabID(size_t aquiferIndex);
+    
+        private:
+  
+            std::vector<AquiferCT::AQUCT_data> m_aquct;
+    
+            inline void set_default_tables(std::vector<double>& td, std::vector<double>& pi)
+            {
+                std::vector<double> default_pressure_ = { 0.112, 0.229, 0.315, 0.376, 0.424, 0.469, 0.503, 0.564, 0.616, 0.659, 0.702, 0.735,
```
atgeirr: Is this the table from van Everdingen and Hurst that is also used by default in Eclipse?

nairr: Yes, it is the constant terminal rate case for an infinite aquifer as given by Van Everdingen and Hurst
[https://www.onepetro.org/download/journal-paper/SPE-949305-G?id=journal-paper%2FSPE-949305-G]

nairr:
I thought I rebased successfully but apparently there was still a conflict. Should be fixed now!

joakim-hove:
Ok - you probably did and then a new conflict came along.

Now final request - blame @atgeirr - please add a reference to the Everdingen and Hurst table.

joakim-hove:
jenkins build this please

nairr:
resolved conflicts and added reference. It should be all good to go now (fingers crossed) :)

joakim-hove:
jenkins build this please

joakim-hove:
@nairr : thank you for your patience with this work!

PR #1179, opened 2018-01-06
Fixing history matching wells part 1

GitPaean:
This is the first part split from  https://github.com/OPM/opm-parser/pull/1168
We agreed to split that PR to make it easier to let it in, this is the first part. 

The major part is that for WCONHIST control production wells, they always have a BHP limit, and they can have another RATE limit instead of always having all the phase rate limits.

### DIFF
--- a/lib/eclipse/EclipseState/Schedule/Schedule.cpp
+++ b/lib/eclipse/EclipseState/Schedule/Schedule.cpp
@@ -455,33 +455,16 @@ namespace Opm {
                     properties = WellProductionProperties::prediction( record, addGrupProductionControl );
                 } else {
                     const WellProductionProperties& prev_properties = well->getProductionProperties(currentStep);
-                    double BHPLimit = prev_properties.BHPLimit;
-                    properties = WellProductionProperties::history( BHPLimit , record, m_phases);
+                    const double BHPLimit = prev_properties.BHPLimit;
+                    properties = WellProductionProperties::history( BHPLimit , record);
                 }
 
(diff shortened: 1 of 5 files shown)


joakim-hove:
> We agreed to split that PR to make it easier to let it in, this is the first part.

Thank you for splitting - hopefully I remember enough of our discussion to review this meaningfully.

GitPaean:
>  Thank you for splitting - hopefully I remember enough of our discussion to review this meaningfully.

It is fine. It is a rather messy area and we probably will make new tests to verify when it is not straightforward to figure out by just reading.

GitPaean:
The travis failure is in opm-core, which should be addressed by OPM/opm-core#1196

joakim-hove:
jenkins build this opm-core=1196 with downstreams please

joakim-hove:
OK - I am ready to merge this when Jenkins goes through. @atgeirr - can I merge: OPM/opm-core#1196

atgeirr:
> can I merge: OPM/opm-core#1196

Sure!

GitPaean:
Please wait a little bit from OPM/opm-simulators#1362 . I will report back. I did not manage to draw an certain conclusion from last round of testing.

GitPaean:
This PR and OPM/opm-core#1196 can be merged without OPM/opm-simulators#1362 .  It shows very tiny performance fluctuation for Norne and it does not impact model 2. 

There is still confusion regarding to OPM/opm-simulators#1362. It needs to be improved before merging.

PR #1184, opened 2018-01-08
Update README with build instructions

pgdr:
`opm-parser` needs `libecl` to build, updated build instructions with cloning and building of both `libecl` and `opm-parser`.


README.md
```diff
@@ -1,10 +1,56 @@
...
+```
+git clone https://github.com/Statoil/libecl
+git clone https://github.com/OPM/opm-parser
+```
+
+
+Configure the paths:
+
+```
+export INSTALL_PATH=/usr/local
```
joakim-hove: Nice - since the instructions are general - and apply to an arbitrary path, it would maybe be better to use e.g `/opt/software` or something else totally non-standard for these environment variables; the all-default `/usr/local` based install should Just Work&trade;.

pgdr: Updated.

joakim-hove: Cool

joakim-hove:
Merge away - or I will do it later.

pgdr:
I cannot without approval :)

PR #1185, opened 2018-01-11
adding keyword FRICTION

GitPaean:
Just to make some model can be parsed.

### DIFF
--- a/lib/eclipse/share/keywords/000_Eclipse100/F/FRICTION
+++ b/lib/eclipse/share/keywords/000_Eclipse100/F/FRICTION
@@ -1,0 +1,4 @@
+     {"name" : "FRICTION", "sections" : ["RUNSPEC"], "size" : 1, "items" : [
+         {"name" : "NWFRIC"       , "value_type" : "INT" , "default" : 0},
+         {"name" : "NWFRIB"       , "value_type" : "INT" , "default" : 1}
+     ]}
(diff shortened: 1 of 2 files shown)


lib/eclipse/share/keywords/000_Eclipse100/F/FRICTION
```diff
@@ -0,0 +1,4 @@
+     {"name" : "FRICTION", "sections" : ["RUNSPEC"], "size" : 1, "items" : [
+         {"name" : "NWFRIC"       , "value_type" : "INT" , "default" : 0},
```
GitPaean: The names are given in the definition, while not totally sure how they are used or interact with other runspec options. It will be okay to me if it is suggested to change to more meaningful names here.

joakim-hove: > The names are given in the definition,

That is what we usually use - this looks fine.

GitPaean:
I do not know `travis` and `appveyor` well, while it looks like these two building system are not very helpful now. 

```
ERROR: cmake-data is broken - called emacs-package-install as a new-style add-on, but has no compat file.
```

GitPaean:
jenkins build this please

joakim-hove:
> while it looks like these two building system are not very helpful now.

We can agree on that - on the way out.

joakim-hove:
I'll merge as soon as Jenkins has started rolling.

PR #1187, opened 2018-01-15
WIP: Added ParseContext::IGNORE_EXTRA_RECORDS

stefoss23:
Adding a new keyword for handling reading of eclipse .DATA files. 
When reading a keyword whose length was specified in the previous keyword, an exception is thrown if there are more records than expected. 

With the new PARSE_EXTRA_RECORDS field, these extra records can be ignored.

### DIFF
--- a/lib/eclipse/Parser/Parser.cpp
+++ b/lib/eclipse/Parser/Parser.cpp
@@ -495,6 +495,7 @@ bool tryParseKeyword( ParserState& parserState, const Parser& parser ) {
         if( parserState.rawKeyword == NULL ) {
             if( RawKeyword::isKeywordPrefix( line, keywordString ) ) {
                 parserState.rawKeyword = createRawKeyword( keywordString, parserState, parser );
+                parserState.lastSizeType = SLASH_TERMINATED;
                 if ( parser.isRecognizedKeyword(line) ) {
                    const auto* parserKeyword = parser.getParserKeywordFromDeckName( line );
                    parserState.lastSizeType = parserKeyword->getSizeType();
--- a/lib/eclipse/Parser/Parser.cpp
+++ b/lib/eclipse/Parser/Parser.cpp
@@ -340,18 +340,20 @@ void ParserState::handleRandomText(const string_view& keywordString ) const {
     std::stringstream msg;
     std::string trimmedCopy = keywordString.string();
 
-    if (lastSizeType == OTHER_KEYWORD_IN_DECK) {
+
+    if (trimmedCopy == "/") {
+        errorKey = ParseContext::PARSE_RANDOM_SLASH;
+        msg << "Extra '/' detected at: "
+            << this->current_path()
+            << ":" << this->line();
+    }
+    else if (lastSizeType == OTHER_KEYWORD_IN_DECK) {
       errorKey = ParseContext::PARSE_EXTRA_RECORDS;
       msg << "Too many records in keyword: " 
           << lastKeyWord
           << ".\n";
     }
-    else if (trimmedCopy == "/") {
-        errorKey = ParseContext::PARSE_RANDOM_SLASH;
-        msg << "Extra '/' detected at: "
-            << this->current_path()
-            << ":" << this->line();
-    } else {
+    else {
         errorKey = ParseContext::PARSE_RANDOM_TEXT;
         msg << "String \'" << keywordString
             << "\' not formatted/recognized as valid keyword at: "
--- a/lib/eclipse/Parser/Parser.cpp
+++ b/lib/eclipse/Parser/Parser.cpp
@@ -229,7 +229,7 @@ class ParserState {
 
     public:
         std::shared_ptr< RawKeyword > rawKeyword;
-        ParserKeywordSizeEnum lastSizeType;
+        ParserKeywordSizeEnum lastSizeType = SLASH_TERMINATED;
         std::string lastKeyWord;
         
         string_view nextKeyword = emptystr;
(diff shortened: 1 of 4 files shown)


GitPaean:
One question about how to use it. I am not familiar with `ParseContext`. The default behavior will still throw an exception, right? How to make it not throw? Thanks.

atgeirr:
The ParseContext.hpp provides some documentation. Basically you can override the settings chosen (by Flow for example) by setting environment variables.

https://github.com/OPM/opm-parser/blob/master/lib/eclipse/include/opm/parser/eclipse/Parser/ParseContext.hpp

joakim-hove:
> The default behavior will still throw an exception, right? How to make it not throw? Thanks.

The default beavior will still be to throw yes; but downstream really determines - so in addition to setting the environment variable one could envisage `flow` having commandline options for this?

lib/eclipse/Parser/Parser.cpp
```diff
@@ -337,7 +340,13 @@ void ParserState::handleRandomText(const string_view& keywordString ) const {
...
     std::string trimmedCopy = keywordString.string();
 
-    if (trimmedCopy == "/") {
+    if (lastSizeType == OTHER_KEYWORD_IN_DECK) {
+      errorKey = ParseContext::PARSE_EXTRA_RECORDS;
+      msg << "Too many records in keyword: " 
+          << lastKeyWord
+          << ".\n";
+    }
+    else if (trimmedCopy == "/") {
```
joakim-hove: I think the ordering should change here, i.e. first check the `trimmedCopy == "/"` condition, and then check the `OTHER_KEYWORD_IN_DECK` - that should also solve the failing tests - I think.

joakim-hove:
This looks fine - I assume the mentioned reordering of test conditions will solve the test problems, and then this can be merged!

joakim-hove:

lib/eclipse/Parser/Parser.cpp
```diff
@@ -229,6 +229,9 @@ class ParserState {
 
     public:
         std::shared_ptr< RawKeyword > rawKeyword;
+        ParserKeywordSizeEnum lastSizeType;
```
joakim-hove: The `lastSizeType` variable should be initialized to `SLASH_TERMINATED` - to ensure well-defined behaviour for parsing errors in the start of the input stream.

joakim-hove:

joakim-hove:
OK - the summaryconfigTest is fixed in master; the ParseContext test I do not understand.

lib/eclipse/Parser/Parser.cpp
joakim-hove: The initialization should be here - i.e.:
```C++
ParserKeywordSizeEnum lastSizeType = SLASH_TERMINATED;
```

lib/eclipse/Parser/Parser.cpp
```diff
@@ -337,12 +340,20 @@ void ParserState::handleRandomText(const string_view& keywordString ) const {
...
             << this->current_path()
             << ":" << this->line();
-    } else {
+    }
+    else if (lastSizeType == OTHER_KEYWORD_IN_DECK) {
+      errorKey = ParseContext::PARSE_EXTRA_RECORDS;
+      msg << "Too many records in keyword: " 
+          << lastKeyWord
+          << ".\n";
+    }
```
joakim-hove: This is missing a call to `parseContext.handleError(ParseContext::PARSE_EXTRA_RECORDS, msg);`

lib/eclipse/Parser/Parser.cpp
```diff
@@ -484,10 +495,17 @@ bool tryParseKeyword( ParserState& parserState, const Parser& parser ) {
         if( parserState.rawKeyword == NULL ) {
             if( RawKeyword::isKeywordPrefix( line, keywordString ) ) {
                 parserState.rawKeyword = createRawKeyword( keywordString, parserState, parser );
+                parserState.lastSizeType = SLASH_TERMINATED;
```
joakim-hove: The init to `SLASH_TERMINATED` here should be removed.

lib/eclipse/Parser/Parser.cpp
stefoss23: The call is in line 363, below the if/elses.

lib/eclipse/Parser/Parser.cpp
stefoss23: There may be a situation in which lastSizeType is set to OTHER_KEYWORD_IN_DECK. Then follows the correct number of records. THEN there are some correct keywords, followed by random gibberish. If the lastSizeType is not reset, this will be handled as a PARSE_EXTRA_RECORDS situation, which it is not. Example:
         EQLDIMS
           2  100  20  1  1  /
         EQUIL
           2469   382.4   1705.0  0.0    500    0.0     1     1      20 /
           2469   382.4   1705.0  0.0    500    0.0     1     1      20 /
         GRID
         randomgibberish

I will also include this as test.

lib/eclipse/Parser/Parser.cpp
```diff
@@ -337,12 +340,20 @@ void ParserState::handleRandomText(const string_view& keywordString ) const {
...
         errorKey = ParseContext::PARSE_RANDOM_SLASH;
         msg << "Extra '/' detected at: "
             << this->current_path()
             << ":" << this->line();
-    } else {
+    }
+    else if (lastSizeType == OTHER_KEYWORD_IN_DECK) {
+      errorKey = ParseContext::PARSE_EXTRA_RECORDS;
+      msg << "Too many records in keyword: " 
+          << lastKeyWord
```
joakim-hove: OK - hopefully last comment; please attach the file and line information to the error message in the same as has been done for the two error blocks.

joakim-hove:

PR #1188, opened 2018-01-16
Add keywords to required fields map for FIP

totto82:


totto82:
That was fast! It confused jenkins downstream.

PR #1189, opened 2018-01-16
Updated test.

joakim-hove:


joakim-hove:
jenkins build this please

GitPaean:
so `;` actually ends that record?

joakim-hove:
> so ; actually ends that record?

Well - it ends the input string.

PR #1190, opened 2018-01-16
Aquancon object

nairr:
An Aquancon object which:
- collects information from the AQUANCON keyword, 
- converts the grid box information into global indices
- converts the per record information of AQUANCON into per Aquifer information
- Applies certain logic for connecting grid cells to aquifers
- returns an output vector with information of aquifer id, global index, cell face, influx coefficient and influx multiplier


lib/eclipse/include/opm/parser/eclipse/EclipseState/Aquancon.hpp
```diff
@@ -0,0 +1,88 @@
...
+  implement the analytical aquifer models in OPM Flow.
+*/
+
+#include <opm/parser/eclipse/EclipseState/Grid/EclipseGrid.hpp>
+#include <opm/parser/eclipse/Parser/ParserKeywords/A.hpp>
+#include <opm/parser/eclipse/Deck/Deck.hpp>
+#include <opm/parser/eclipse/Deck/DeckItem.hpp>
+#include <opm/parser/eclipse/Deck/DeckRecord.hpp>
+#include <opm/parser/eclipse/Deck/DeckKeyword.hpp>
+#include <boost/concept_check.hpp>
```
joakim-hove: I don't think this header is used?

lib/eclipse/include/opm/parser/eclipse/EclipseState/Aquancon.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+namespace Opm {
+
+    class Aquancon {
+        public:
+
+            struct AquanconOutput{
+                int aquiferID;
+                std::vector<size_t> global_index;
+                std::vector<double> influx_coeff; // Size = size(global_index)
```
joakim-hove: At least how I see it - the fundamental unit here should be *one cell* - connected to one aquifer; i.e. *not* vectors of indices and flow properties.

lib/eclipse/include/opm/parser/eclipse/EclipseState/Aquancon.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+            void logic_application(std::vector<Aquancon::AquanconOutput>& output_vector);
+
+            void collate_function(std::vector<Aquancon::AquanconOutput>& output_vector);
+
+            void convert_record_id_to_aquifer_id(std::vector<int>& record_indices_matching_id, int i);
+
+            std::vector<Aquancon::AquanconOutput> m_aquoutput;
+
+            std::vector<Aquancon::AquanconRecord> m_aqurecord;
```
joakim-hove: The `m_aqurecord`, `m_aquiferID_pre_record` and `m_maxAquID` variables are only using during the construction phase, they should be local variables in the constructor and not part of the class as such. Also `struct AquanconRecord`can just be in an anonymous namespace in the cpp file - no need to make the class definition so heavy.

lib/eclipse/EclipseState/Aquancon.cpp
```diff
@@ -0,0 +1,159 @@
...
+    { 
+        if (!deck.hasKeyword("AQUANCON"))
+        {
+            getAquOutput();
+        }
+        
+        else {
+            const auto& aquanconKeyword = deck.getKeyword("AQUANCON");
+            // Resize the parameter vector container based on row entries in aquancon
+            m_aqurecord.resize(aquanconKeyword.size());
```
joakim-hove: I do not manage to follow everything in your implementation, but to me it does seem overly complex. Personally I would go through the records, expand into global indices and add directly to the `m_aquoutput` vector - completely bypassing the `m_aqurecord` variable, and also the quite complex book keeping in the private functions - just need to make sure the `(global_index, aquifer_id)` pair is not already in the output vector.

```C++
// Using small utility Box to assist with the calculation of indices:
Box grid_dims( grid.getNX(), grid.getNY(), grid.getNZ());

for (const auto& record :  aquanconKeyword) {
     int I1 = record.getitem("I1").get<int>(0);
     int I2 = record.getItem("I2").get<int>(0);
     ...
     int k2 = record.getItem("K2").get<int>(0);
     int aquifer_id = ..
     double influx_coeff = ...
     ....
     Box box(grid_dims, I1 - 1, I2 - 1, J1 - 1, J2 - 1, K1 - 1, K2 -1 );
    
     for (size_t g : box) {
            const auto * it = std::find_if( m_aquoutput.begin(), m_aquoutput.end(), [&g,&aquifer_id]( const struct AquanconOutput aq) { return (aq.aquiferID == aquifer_id && aq.global_index == g); } 
            if (it == m_aquauput.end())
               m_aquaoutput.emplace_back( { } )
     }
}
```

lib/eclipse/EclipseState/Aquancon.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+#include <opm/parser/eclipse/EclipseState/Aquancon.hpp>
+#include <algorithm>
+#include <iterator>
+
+namespace Opm {
+
+    Aquancon::Aquancon(const EclipseGrid& grid, const Deck& deck)
+    { 
+        if (!deck.hasKeyword("AQUANCON"))
```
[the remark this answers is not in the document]
nairr: yes, agreed

lib/eclipse/EclipseState/Aquancon.cpp
nairr: The idea here is to output a vector which can identify all connections and essential quantities per aquifer instead of per AQUANCON record. m_aqurecord contains the per record information and m_aquoutput contains the per aquifer information. This is also what is asked for from the opm-simulators side. 

eg. The AQUANCON keyword could be as follows:


```
 AQUANCON
     1      1  33  1    33   1  2  J-  /
     1      1  33  1    33   1  2  J+  /
     1      1  33  1    33   1  2  I-  /
     2      1  33  1    33   1  2  J-  /
     1      1  33  1    33   1  2  I+  /
     2      1  33  1    33   1  2  J+  /
 /
```
 


Therefore, according to us eventhough the implementation is more complex, having this sort of a conversion to a per aquifer record has to be done somewhere, makes it easier to calculate face area fraction for each connection per aquifer and to implement logic such as: "Total number of grid blocks connected to aquifer must not exceed item 6 of AQUDIMS".

lib/eclipse/include/opm/parser/eclipse/EclipseState/Aquancon.hpp
nairr: no, its not, will remove it

lib/eclipse/include/opm/parser/eclipse/EclipseState/Aquancon.hpp
nairr: ok, will make the changes

lib/eclipse/EclipseState/Aquancon.cpp
joakim-hove: > The idea here is to output a vector which can identify all connections and essential quantities per aquifer instead of per AQUANCON

OK - if per aquifer is most convenient downstream, then that is fine; although I do find that surprising. The records are purely input constructions and certainly relevant; what I (still ...) think would be the best was a datastructure for *one single aquifer connection* - i.e. *one flat vector* of something like:
```
struct AquiferCell {
     int aquifer_id;
     size_t global_index;
     double influx_coefficient;
     double flux_multiplier;
     face_enum face;
}
```
then it would be quite simple to perform ad-hoc calculations, filtering and so on based on this vector.

lib/eclipse/include/opm/parser/eclipse/EclipseState/Aquancon.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+    class Aquancon {
+        public:
+
+            struct AquanconOutput{
+                int aquiferID;
+                std::vector<size_t> global_index;
+                std::vector<double> influx_coeff; // Size = size(global_index)
+                std::vector<double> influx_multiplier; // Size = size(global_index)
+                std::vector<std::string> reservoir_face_dir; // Size = size(global_index)
```
joakim-hove: Use [this enum](https://github.com/OPM/opm-parser/blob/master/lib/eclipse/include/opm/parser/eclipse/EclipseState/Grid/FaceDir.hpp) instead of strings for the face direction.

lib/eclipse/EclipseState/Aquancon.cpp
nairr: With @kel85uk we discussed exactly what you are suggesting, but based on the requirement from simulators, the existing implementation there and for the reasons mentioned above, we decided the per aquifer output implementation was easier

nairr:
I've made the requested changes with the latest commit

lib/eclipse/EclipseState/Aquancon.cpp
joakim-hove: OK - but then I will stop nagging about that.

joakim-hove:
OK - then you are just missing some basic tests of this class, and it is ready to be merged.

nairr:
The tests should be in now. Thanks for helping me out through this!

joakim-hove:

lib/eclipse/tests/AquanconTests.cpp
```diff
@@ -0,0 +1,79 @@
...
+        "   1      1  1  1    1   1  1  J-  1.0 1.0 NO /\n"
+        "/ \n";
+
+    Parser parser;
+    return parser.parseString(deckData, ParseContext());
+}
+
+inline std::vector<Aquancon::AquanconOutput> init_aquancon(){
+    auto deck = createAQUANCONDeck();
+    EclipseState eclState = createAQUANCONDeck();
```
joakim-hove: This should be:
```
EclipseState eclState( deck ); 
```
I guess the current works - but it looks really funny (to me at least).

lib/eclipse/tests/AquanconTests.cpp
```diff
@@ -0,0 +1,79 @@
...
+    Parser parser;
```
joakim-hove: This `Parser` instance is unused.

lib/eclipse/tests/AquanconTests.cpp
```diff
@@ -0,0 +1,79 @@
...
+    
+    Aquancon aqucon = Aquancon(eclState.getInputGrid(), deck);
```
joakim-hove:
```
Aquancon aqucon( eclState.getInputGrid(), deck);
```

lib/eclipse/tests/AquanconTests.cpp
```diff
@@ -0,0 +1,79 @@
...
+    std::vector<Aquancon::AquanconOutput> aquifers = aqucon.getAquOutput();
+
+    return aquifers;
+}
+
+BOOST_AUTO_TEST_CASE(AquanconTest){
+    std::vector< Aquancon::AquanconOutput > aquifers = init_aquancon();
+    for (auto it =aquifers.begin(); it != aquifers.end(); ++it){
```
joakim-hove: Use range based `for` instead of naked iterators - so much nicer to read and work with:
```C++
for (const auto& aq :  aquifers) {
     ...
     BOOST_CHECK_EQUAL( aq.aquiferID, 1);
     ...
}
```

joakim-hove:
OK - some minor C++ comments on the test; then this is ready to be merged. But I must admit I really don't like the approach with multiple aquifer connections in one struct. Before I merge - could you point me to the downstream code using this - that the current approach is more convenient there will be the main argument to merge the current approach.

kel85uk:
Hi @joakim-hove , sure. It's located here https://github.com/kel85uk/opm-simulators/blob/task/%235/Make_correct_initialization_procedure/opm/autodiff/BlackoilAquiferModel_impl.hpp at the init function

nairr:
cleaned up the test

joakim-hove:

PR #1193, opened 2018-01-18
Default box

joakim-hove:
Allow the use use of defaults in the operational keywords like `ADD` and `MULTIPLY`:

This now works:

```
MULTIPLY
   'PERMX' 0.25 * * * * 4 4 /
/
```
to multiply `PERMX`in layer 4 - previously '*' could not be used.

### DIFF
--- a/lib/eclipse/CMakeLists.txt
+++ b/lib/eclipse/CMakeLists.txt
@@ -67,6 +67,7 @@ set(opmparser_SOURCES Deck/Deck.cpp
                       EclipseState/Grid/NNC.cpp
                       EclipseState/Grid/PinchMode.cpp
                       EclipseState/Grid/SatfuncPropertyInitializers.cpp
+                      EclipseState/Grid/setKeywordBox.cpp
                       EclipseState/Grid/TransMult.cpp
                       EclipseState/InitConfig/Equil.cpp
                       EclipseState/InitConfig/InitConfig.cpp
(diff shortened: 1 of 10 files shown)


joakim-hove:

joakim-hove:

OPMUSER:
No, incorrect format should be

|MULTIPLY 'PERMX' 0.25 1* 1* 1* 1* 4 4 / /|


/*E*/

	

/*I*/

	

/*P*/

	

/*C*/

	

Managing Director

T:  +(65)-9173-7031

F:  +(65)-6732-2382

E: david.baxendale@eipc.co

joakim-hove:
> No, incorrect format should be

Well - incorrect is incorrect - I can live with that.

atgeirr:
> Well - incorrect is incorrect - I can live with that.

I can't help but think there was some miscommunication here? Does the current code support `*`, `1*` or both notations for defaulting?

joakim-hove:
> I  can't help but think there was some miscommunication here? 

Misunderstandings in written communication - that is our trade!

> Does the current code support *, 1* or both notations for defaulting?

The current code invokes the 100% normal opm-parser default handling scheme, which supports both `*` and `1*` for defaulting. Actually the previous code - which is now removed, was a special casing to circumvent the default default handling - due to a misunderstanding of the documentation. So now we actually have simpler and more capable code!
Does the current code support *, 1* or both notations for defaulting?

andlaus:
IIRC omitting the 1 in front of the asterisk is incorrect according to the ECL documentation but testing revealed that E100 eats such constructs just fine and there are apparently some decks out there which do this so opm-parser follows their great example as usual. maybe I don't remember this correctly, though. anyway, *if* above statement is correct this is a rare case where their implementation is more sensible than their documentation IMO.

PR #1194, opened 2018-01-18
[WIP] Support WSEGSICD

GitPaean:
The support of keyword WSEGSICD. 

The test towards the major model is not done yet, and there might be something needs to be adjusted / improved, so it is marked WIP.


joakim-hove:
Thank you for writing test code.

lib/eclipse/EclipseState/Schedule/MSW/Segment.cpp
```diff
@@ -146,4 +150,16 @@ namespace Opm {
             && this->m_volume            == rhs.m_volume
             && this->m_data_ready        == rhs.m_data_ready;
     }
+
+    void Segment::updateSpiralICD(const SpiralICD& spiral_icd) {
+        m_segment_type = WellSegment::SPIRALICD;
+        m_spiral_icd = std::make_shared<SpiralICD>(spiral_icd);
+    }
+
+    const std::shared_ptr<const SpiralICD>& Segment::spiralICD() const {
```
joakim-hove: opm-parser has been on a quite strict `shared_ptr<>` diet; do you really need a `std::shared_ptr` here - or are you just following the old opm-parser pattern of excessive `shared_ptr` abuse?

lib/eclipse/include/opm/parser/eclipse/EclipseState/Schedule/MSW/SpiralICD.hpp
```diff
@@ -0,0 +1,95 @@
...
+#include <opm/parser/eclipse/Deck/DeckKeyword.hpp>
+
+#include <limits>
+#include <map>
+#include <utility>
+
+
+namespace Opm {
+
+struct SpiralICD {
```
joakim-hove: Now - this was a real pattern breaker:
1. Implementation in `.hpp` file.
2. `struct` with public members instead of a class with private members.
3. No test?! New (non trivial) classes shall have tests!

I am not saying the patterns are sacred, but I would prefer you follow them unless there is a specific reason not to.

lib/eclipse/EclipseState/Schedule/MSW/Segment.cpp
joakim-hove: Well - first of all; from the outset opm-parser used `std::shared_ptr<T>` extensively; mostly in situations where *shared ownership* - was not relevant at all. Mainly due to @jokva the code base was put on a strict `std::shared_ptr<T>` diet - mostly replacing with references and automatically initialized stack variables - that was a great improvement.

My personal take on this is:

1. Use automatically initialized stack variables.
2. Try harder to use automatically initialized stack variables.
3. Use `std::unique_ptr<T>` for situations where you might - or might not - have an object.
4. Use `std::shared_ptr<T>` for situations where you actually have shared ownership.

I have not looked in detail on your code, and maybe it is actually appropriate to use the `std::shared_ptr<>` - I will look; feel free to explain briefly.

The problem with using these smart pointers in the API is that they spread to consumer code; i.e. the library implementer affects the memory model used by the application developer - that is kind of ugly.

lib/eclipse/include/opm/parser/eclipse/EclipseState/Schedule/MSW/SpiralICD.hpp
joakim-hove: > Please confirm a .cpp file for implementation is what you want here.

yes - that is what I want.

joakim-hove: > I can change to a class

Yes please; preferably with private members - but I will not insist.

lib/eclipse/EclipseState/Schedule/MSW/Segment.cpp
GitPaean: Yes. The natural choice is 3. `std::unique_ptr`, since it is not something  related to all segments, only exists for certain segments.  But `std::fill` used in `DynamicState` `update` make it better to use `std::shared_ptr`. 

And also, in the schedule related, there are quite some of places, we get a Segment, or SegmentSet, modify some, then replace the old one. Although we can use `std::move` to avoid the problem related to `std::unique_ptr`, it turned out to be not very easy. 

Anyway, I might have some wrong understanding here, please provides some suggestions.

joakim-hove: Then you stick to `std::shared_ptr<>` - no problem!

lib/eclipse/include/opm/parser/eclipse/EclipseState/Schedule/MSW/SpiralICD.hpp
GitPaean: I just provided some fixes to your comments. Please let me know if they are what you want.

lib/eclipse/EclipseState/Schedule/MSW/SpiralICD.cpp
```diff
@@ -0,0 +1,72 @@
...
+    {
+        strength = record.getItem("STRENGTH").getSIDouble(0);
+        length = record.getItem("LENGTH").getSIDouble(0);
+        density_calibration = record.getItem("DENSITY_CALI").getSIDouble(0);
+        viscosity_calibration = record.getItem("VISCOSITY_CALI").getSIDouble(0);;
+        critical_value = record.getItem("CRITICAL_VALUE").getSIDouble(0);
+        width_transition_region = record.getItem("WIDTH_TRANS").get< double >(0);
+        max_viscosity_ratio = record.getItem("MAX_VISC_RATIO").get< double >(0);
+        method_flow_scaling = record.getItem("METHOD_SCALING_FACTOR").get < int >(0);
+        if (record.getItem("MAX_ABS_RATE").hasValue(0) )
```
joakim-hove: OK; I guess you could add a default value of `1e-20` for the `MAX_ABS_RATE` item in the `WEGSICD` keyword definition - then everything here is *pure assignment* and the full constructor could be implemented with an initializer list(?):

```C++
SpiralICD::SpiralICD(const DeckRecord& record) :
   stringth( record.getItem("STRENGTH").getSIDouble(0)),
   length( record.getItem("LENGTH").getSiDouble(0)), 
    ...
    status( record.getItem("STATUS").getTrimmedString(0))
{  }
```

joakim-hove:

lib/eclipse/EclipseState/Schedule/MSW/SpiralICD.cpp
GitPaean: `1.e-20` will not work, since by default it should be no limit. `1.e-20` will give a too strong limit here. 

I can do the following, 
```
max_absolute_rate(record.getItem("MAX_ABS_RATE").hasValue(0) ? record.getItem("MAX_ABS_RATE").getSIDouble(0)  : std::numeric_limits<double>::lowest(); ```
```

Maybe `lowest()` should have been `max()`. I was thinking a negative value can be used to avoid comparison. It really depends on how to enforce `No limit` here.

joakim-hove: Ahhh I misunderstood; I interpreted `lowest` as *closest to zero*; but then I would say you should use `max( )` instead of `lowest( )`.

lib/eclipse/EclipseState/Schedule/MSW/SpiralICD.cpp
```diff
@@ -0,0 +1,71 @@
...
+            if (start_segment < 2 || end_segment < 2 || end_segment < start_segment) {
+                const std::string message = "Segment numbers " + std::to_string(start_segment) + " and "
+                                            + std::to_string(end_segment) + " specified in WSEGSICD for well " + well_name
+                                            + " are illegal ";
+                throw std::invalid_argument(message);
+            }
+
+            const SpiralICD spiral_icd(record);
+            for (int seg = start_segment; seg <= end_segment; seg++) {
+                res[well_name].push_back(std::make_pair(seg, spiral_icd));
```
GitPaean: I just began trying to use some IDE tool for my OPM development.  It is suggested to change the `push_back` to `emplace_back`  here. It is something I have not thought about. Any comments about this?

lib/eclipse/include/opm/parser/eclipse/EclipseState/Schedule/MSW/SpiralICD.hpp
```diff
@@ -0,0 +1,59 @@
...
+    public:
+        double strength;
+        double length;
+        double density_calibration;
+        double viscosity_calibration;
+        double critical_value;
+        double width_transition_region;
+        double max_viscosity_ratio;
+        int method_flow_scaling;
+        // negative value means no limit for max_absolute_rate
```
GitPaean: If we know the length of the completions at  this point, the `length` and the `method_flow_scaling` can be changed to another one single member named `flow_scaling_factor`, same with the one `flow_scaling_factors_` in the downstream PR OPM/opm-simulators#1389 , which will provide a better solution, since `flow_scaling_factor` is the one actually used in the simulation.

GitPaean:
Just added a member variable to Completion for its length. As far as I know, the places can make use of this information is the shear calculation in polymer simulation, the current ICD implementation and some other incoming development.  Not sure how it will be bitten by the pinching during the grid processing though.   Please let me know there are more things in the tests needed to be addressed for this completion length.  There are so many tests related to Completion, it is difficult to get all of them correctly.

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
Jenkins build this please

GitPaean:
Jenkins build this please

joakim-hove:
I am ready to merge this - but it is still marked as WIP?

akva2:
just note that any merges done to parser or output means i will have to redo https://github.com/OPM/opm-common/pull/341 (mostly automated but...)

GitPaean:
> I am ready to merge this - but it is still marked as WIP?

It has not been well tested against the major model we are targeting, so it is still WIP. 

> just note that any merges done to parser or output means i will have to redo OPM/opm-common#341 (mostly automated but...)

I understand that. It is okay to me to create a new pull request in the new opm-common repo.

atgeirr:
> just note that any merges done to parser or output means i will have to redo OPM/opm-common#341 (mostly automated but...)

While we cannot "move" PRs I think it is ok to require reopening the PR against the new repo. Agree?

atgeirr:
This PR must be moved to the opm-common module, following the recently merged reorganization.

PR #1198, opened 2018-01-24
changed: remove duplicated test

akva2:
Ref https://github.com/OPM/opm-parser/blob/master/lib/eclipse/CMakeLists.txt#L180


joakim-hove:
Hmmm - that was weird; I just added this in #1183 because some broken keyword configuration had not triggered the test failure it should. Anyway - I agree it is a duplicate.

PR #1203, opened 2018-01-31
respect the fact that non-SI temperature may be offset from zero

### LINKED ISSUE #1202: Unit conversion involving temperature
While testing the energy module flow I notice an issue with unit conversion and temperature. For me it seems like the "constant" scaling factor is not taken into account. I added the following simple test to UnitTests.cpp and it fails on all 4. 
  
```

BOOST_AUTO_TEST_CASE(Temperature) {
    auto field = UnitSystem::newFIELD();
    auto metric = UnitSystem::newMETRIC();

    BOOST_CHECK_CLOSE( 80.33 , field.from_si(UnitSystem::measure::temperature, 300.0) , 1.0e-6 );
    BOOST_CHECK_CLOSE( 26.85 , metric.from_si(UnitSystem::measure::temperature, 300.0) , 1.0e-6 );

    BOOST_CHECK_CLOSE( 310.928 , field.to_si(UnitSystem::measure::temperature, 100.0) , 1.0e-6 );
    BOOST_CHECK_CLOSE( 373.15 , metric.to_si(UnitSystem::measure::temperature, 100.0) , 1.0e-6 );

}
```

andlaus:
also, the UnitTests test was buggy because it tested absolute temperature when it was supposed to test the temperature.

this should fix #1202.

### DIFF
--- a/lib/eclipse/Units/UnitSystem.cpp
+++ b/lib/eclipse/Units/UnitSystem.cpp
@@ -38,6 +38,35 @@ namespace {
      * metric and field arrays. C++ does not support designated initializers, so
      * this cannot be done in a declaration-order independent matter.
      */
+    static const double from_metric_offset[] = {
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        Metric::TemperatureOffset,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0
+    };
 
     static const double to_metric[] = {
         1,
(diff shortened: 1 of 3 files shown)


andlaus:
jenkins build this please

totto82:
Maybe adding the test case I provided in #1202?

andlaus:
the non-absolute temperature is already tested, albeit only for PVT-M. I can still add it if there is agreement, but the benefits would not be very large?

akva2:
this broke the restart test in opm-output.

joakim-hove:
Damn - my fault, sorry.

I will force push `HEAD~~` and then we can repeat more carefully?

atgeirr:
These commits are probably correct, maybe the restart test is quick to fix?

joakim-hove:
OK - should be fixed here: https://github.com/OPM/opm-output/pull/241

I misunderstood the message from @akva2 slightly - I thought the whole test-rig was in shatters; things were not *that bad*.

atgeirr:
Failing opm-simulators tests (on both my machine and Jenkins) are cause by this:
```
38: Comparing TEMP...
38: Occurrence in first file    = 0
38: Occurrence in second file   = 0
38: Grid coordinate             = (1, 1, 1)
38: (first value, second value) = (-459.666, 519.669)
```

atgeirr:
Looking at the actual output temperatures, I think there is a significant chance that the code now has a significant bug. Consider the data above, the old output was 519.669 F, which is approx 15.5 C. I think that has been used as a default temperature, so it makes sense. The new output is -459.666 F, which is absolute zero. Clearly we still want to output the default temperature here.

The only use of the offset (outside the converter function here) I found in opm-parser is in TableManager.cpp line 98. I am not sure if it is correct or not?

joakim-hove:
Ok - what a mess. I suggest to revert this and the PR in opm-material.

atgeirr:
> Ok - what a mess. I suggest to revert this and the PR in opm-material.

I concur.

andlaus:
go ahead. `to_si()` and `from_si()` and the UnitsTest should be fixed, though?

andlaus:
BTW: the old output probably was about 520°R, not °F...

joakim-hove:
OK - I will now revert this and: https://github.com/OPM/opm-material/pull/282 to recover the Jenkins 

We (I at least ...) will have to look more into this, and recover the parts we should retain.

atgeirr:
> BTW: the old output probably was about 520°R, not °F...

Yes, I meant Rankine, sorry. So there is a bug already in the current master, since we are expected to output in Fahrenheit! (So we should at the end of this give the right output in F not R for this problem).

PR #1206, opened 2018-02-06
Added ability to filter out completions from inactive cells.

joakim-hove:


joakim-hove:
jenkins build this please

joakim-hove:
@GitPaean : does this look OK?

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

lib/eclipse/include/opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp
```diff
@@ -111,6 +111,11 @@ namespace Opm
         const MessageContainer& getMessageContainer() const;
         MessageContainer& getMessageContainer();
 
+        /*
+          Will remove all completions which are connected to cell which is not
+          active. Will scan through all wells and all timesteps.
+        */
+        void filter_completions(const EclipseGrid& grid);
```
atgeirr: Is there a reason whyyou call it filter_completions() rather than filterCompletions()?

GitPaean:
> @GitPaean : does this look OK?

Thanks. I will take a look and try to update the PR in opm-simulators.

lib/eclipse/EclipseState/Schedule/Well.cpp
```diff
@@ -574,4 +574,10 @@ namespace Opm {
     bool Well::hasEvent(uint64_t eventMask, size_t reportStep) const {
         return this->events.hasEvent( eventMask , reportStep );
     }
+
+
+    void Well::filter_completions(const EclipseGrid& grid) {
+        for (auto& completions : m_completions)
```
atgeirr: This was the only part I misunderstood at first: I did not realize that this was the loop over time steps! While I think the addition of begin() and end() to DynamicState is fine, it is not obvious to the reader (well, me at least...) if we are looping over times first and completions second (correct here, since we have DynamicState<CompletionSet>), or the other way around (if a CompletionSet had been a vector of DynamicState<Completion>).

atgeirr: Just an idea to make it more obvious: instead of providing begin() and end(), so you can use the DynamicState as a range, it could provide a timeRange() or allTimes() or similarly named functions that returned the iterator range. Then you would write the above as
```
for (auto& completions : m_completions.allTimes())
```
What do you think?

lib/eclipse/include/opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp
joakim-hove: Change of opinion; I have come to like `filter_completions`better - but maybe that is too late. I guess I should should change it to `filterCompletions()`?

lib/eclipse/EclipseState/Schedule/Well.cpp
joakim-hove: I think that was a good idea - what would the implementation of the `DynamicState::allTimes()` method look like? Return a pair of iterators?

lib/eclipse/include/opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp
atgeirr: Since that is the convention followed for the rest of opm-parser (and indeed most of the rest of OPM) I think we should keep it.

I used to use camelCaseMethods and underscore_data_members, but have nearly been convinced to camelCaseEverything since that is used by most of the other developers. So I was a bit surprised to see your opinion swinging the other way again... (although for methods).

joakim-hove: I will change.

lib/eclipse/EclipseState/Schedule/Well.cpp
atgeirr: I am not sure if pair<iterator,iterator> will be treated as a range, if so then that is the best. Worst case you'll have to use or import boost::iterator_range, which is what I have used up to now. There seems to be no std:: replacement for it yet, I did a brief search.

GitPaean:
> Thanks. I will take a look and try to update the PR in opm-simulators.

The functionality works well in the simulator side. Thanks!

joakim-hove:
> The functionality works well in the simulator side. Thanks!

Good - observe that I have pushed an update using `Schedule::filterCompletions( )` instead.

lib/eclipse/EclipseState/Schedule/Well.cpp
joakim-hove: OK - I liked your idea; but I did not venture into boost for this. Have added a comment though.

atgeirr: That's fine. The comment makes it very clear what is going on.

atgeirr:
jenkins build this please

PR #1214, opened 2018-02-14
fixed: set the dynamic boost test define when building shared libs

akva2:


joakim-hove:
jenkins build this opm-common=334 please

PR #1219, opened 2018-02-20
Silence warning.

atgeirr:
Another minor warning squashing: unlike returning const references, returning const values is no different from returning a non-const value.


atgeirr:
jenkins build this please

joakim-hove:
jenkins build this please

PR #1220, opened 2018-02-22
[testing purpose] adding keywords SKPRWAT and SKPRPOLY

GitPaean:
I did not manage to use the PVTX related facility to handle this keywords.  These keywords are more regular than the `PVTX` keywords. 

They are 2D tables, we have the sample points for one direction, and we have samples points for the other directions. Then we have all the sample values.  Since the PVTX tables are essentially 2D tables, while has some other characteristics, I was thinking to handle these two keywords to use the `PVTX` facilities. 

Now, I am thinking to use  2D tables more specific for these keywords. Basically, I will introduce a new class `Simple2DTable` to handle this  type of tables, with the x, y sampling points and all the sample values. 

An example of one of the keywords looks like follows, (it is more similar to the VFP tables)
```
SKPRPOLY
   1      2.0     /    -- table number and reference concentration
   0.0     1.0   2.0   / -- throughput values
   0.0     1.0   2.0  3.0  / -- velocity values
   -- the rest will be the skin pressure values, 
  -- each row corresponds to one sample points in the throughput direction
   0.     0.1   0.2   0.3 /
   0.     0.2   0.4   0.6 /
   0.     0.3   0.6   0.9 / 
```
The old way please refer to #1216. 

Any comments or suggestions will be much appreciated.


GitPaean:
jenkins build this please

PR #1221, opened 2018-02-28
Remove json library

akva2:
This removes the opmjson library, instead opting for compiling the sources twice (one for generator, one for main library). less hassle.

the second commit fixes debian packaging and thus restores nightly builds.


atgeirr:
> compiling the sources twice

The JSON library only, I assume?

akva2:
yes, two sources (a .c file and a .cpp file).

joakim-hove:
jenkins build this please

PR #1222, opened 2018-03-06
mark module as deprecated

akva2:


akva2:
shouldn't be controversial so i will use my admin privs.

PR #1226, opened 2018-05-30
[WIP] VFP number is always read for WCONINJE

GitPaean:
and a zero-value THP limit for WCONPROD will not be treated as a THP limit.

The history matching part is slightly more tricky, was not handled by this PR yet.


GitPaean:
It was a drama, I should not have kept the opm-parser folder.

